PR #46, opened 2014-12-16
Fix problem with windows builds mistakenly trying to include unistd.h et al.

rplzzz:
I believe I have found a set of preprocessor macros that arranges for `unistd.h`, `sys/stat.h`, and `sys/types.h` to be included on unix and mac, but not on windows.  I have tested the fix on Linux, and everything works as expected, but having no access to a mac or windows system, I can't test there.  Please test on both windows and mac before merging this request.

Here are the behaviors expected on mac:
* The code should build successfully while including the headers mentioned above.
* If the logs directory does not exist at run time, it should be created.
* If a file (not a directory) called logs exists at run time hector should fail with the message "* Program exception: Log directory logs exists but is not a directory"
* If the logs directory exists without write permission at run time hector should fail with the message "* Program exception: Log directory logs lacks write permission."

The expected behavior on windows is:
* The code should not attempt to include any of the headers mentioned above.
* On any of the problems with the logs directory described above, hector should fail with the message "Unable to open log file".

### DIFF
--- a/source/core/logger.cpp
+++ b/source/core/logger.cpp
@@ -9,7 +9,7 @@
 #include <time.h>
 #include "core/logger.hpp"
 
-#if !defined(_WIN32)
+#if defined (__unix__) || defined (__MACH__)
 #include <unistd.h>
 #include <sys/stat.h>
 #include <sys/types.h> 
@@ -245,7 +245,7 @@ const char* Logger::getDateTimeStamp() {
     return ret;
 }
 
-#if !defined(_WIN32)
+#if defined (__unix__) || defined (__MACH__)
 int Logger::chk_logdir(std::string dir)
 {
     // clip trailing /, if any.  


cahartin:
@rplzzz  I merged the new fix into the master branch but it is still not recognizing the unistd.h.  Should I revert back the merge? Or leave it for now and we can take a look at it next week?

rplzzz:
The windows build shouldn't be trying to include `unistd.h` at all.  It looks like I missed `unistd.h` being included from `logger.hpp`.  It's completely unguarded (not even the `#ifdef _WIN32` that I had previously used), so in retrospect it is probably the cause of the original bug.  I'll fix it and push the corrected version.

We are going to have to come up with a better way to test these cross-platform fixes.

bpbond:
Yeah, I had the same "huh?" w.r.t. unistd.h…

We use https://travis-ci.org for CI/testing on the RCMIP5 project…alas, it doesn't support Windows. But it would be really useful to know immediately if any commit breaks the build on any platform.

PR #100, opened 2015-08-20
Needed #include <algorithm>

rplzzz:
Fix windows build error.


rplzzz:
Tested on Linux.  Build and run are both ok.

bpbond:
Will look at this after lunch. Thanks.

PR #107, opened 2015-08-25
Release candidate 1.1.1

rplzzz:
Changes (please add any further changes to this list; we'll add the whole thing to the change log just before we merge):

* Fix OS X build
* Update sample output files
* Add GPL license


rplzzz:
Currently our source files have no copyright notice.  The GPL (in its "How to Use" section) recommends the following:

    Hector -- A Simple Climate Model
    Copyright (C) 2014-2015  Battelle Memorial Institute

    This program is free software; you can redistribute it and/or modify
    it under the terms of the GNU General Public License, version 2 as 
    published by the Free Software Foundation.

    This program is distributed in the hope that it will be useful,
    but WITHOUT ANY WARRANTY; without even the implied warranty of
    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
    GNU General Public License for more details.

    You should have received a copy of the GNU General Public License along
    with this program; if not, write to the Free Software Foundation, Inc.,
    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.

I have omitted the language allowing redistribution under "any subsequent version" because I don't see that it gains us anything, and I'm a little uneasy potentially allowing redistribution under the terms of licenses that haven't even been written yet.

Absent any objection I will add the above text as a block comment at the top of each source file.

pralitp:
Sounds good to me.

bpbond:
Yes, agreed. Thanks for taking this on.

cahartin:
looks good. thanks @rplzzz

rplzzz:
In case you were wondering how to prepend a block of text to a bunch of files without doing each one by hand:
1. create `~/bin/addlegal.sh`:
```sh
sed -i -e '1s/^/\n/' $1
sed -i -e '1r/people/link593/tmp/legal.hpp' $1
sed -i -e '1d' $1
```
2. Then run:
```
find . -name \*.cpp -exec ~/bin/addlegal.sh {} \;
find . -name \*.hpp -exec ~/bin/addlegal.sh {} \;
```

Issue #92, opened 2015-07-09
Future pull requests from non-PNNL folks


cahartin:
...we'll need to be prepared for this w.r.t. IP.

rplzzz:
Turns out there is a relatively easy way to communicate our contribution policies (once we decide on them) to would-be contributors:  https://github.com/blog/1184-contributing-guidelines .  (By the way, note that contribution policies typically cover a lot more than just licenses and copyrights -- we should give some real thought to what we want people to do to validate their contributions before we review them.)

bpbond:
That's an interesting link--thanks. Awesome that GitHub will auto-generate a link when people open a pull request or issue! Definitely a v1.2 issue.

bpbond:
I have created (55744d2935) a `CONTRIBUTING.md` document wrt this issue. @cahartin @rplzzz please take a look.

cahartin:
@bpbond  that's great!

rgieseke:
This will be helpful! Some comments below:

**Getting started**
The link to Hector's version numbering system should probably already here be linked to.

> Trivial changes to comments or documentation do not require creating a new issue. Just start the first line of a commit with '(doc)' instead of an issue number.

Do you want commit messages to start with the issue number? I reference them usually in the description paragraph of the commit message or at the end of the first line. Since this is not used in the commit messages so far it might be helpful to clarify this.
The "proper format" is also referenced in "Making changes", maybe having an example might be useful.

**Making changes**
The branching part is not clear to me yet, maybe just because the branches don't exist yet. Maybe you could elaborate on that. Do you plan to maintain older versions with bug fixes once you switched the major release number?

**Submitting Changes**
I'd think the "statement" requirement and the pull request closing time sound a bit discouraging. The latter I would probably just leave out. Depending on circumstances (and previous interactions) you might want to vary this time span. GitHub's interface makes it also quite easy to re-open something, so you could always do that when the absent-minded professor returns from a sabbatical.

As for the copyright issues, maybe you could word this as a default. The NodeJS project for example has this [paragraph](https://github.com/nodejs/node/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-10):

>By making a contribution to this project, I certify that:
>    (a) The contribution was created in whole or in part by me and I have the right to submit it under the open source license indicated in the file; or
[...]

Issue #61, opened 2015-01-12
Continuous integration of Hector repository


bpbond:
[Travis](https://travis-ci.org/) provides continuous integration, i.e. automatically building any commit or pull request and flagging errors, and integrates seamlessly with Github. (E.g. we do this for [RCMIP5](https://github.com/JGCRI/RCMIP5); see the green "build passing" badge on main page.) This would be good to add for Hector at some point, building on both Linux and Mac OS.

bpbond:
This has been done - see #158 . Closing.

PR #110, opened 2015-08-26
Make CSV paths relative to the INI file path

rgieseke:
The CSV paths in `csv:` notation need to be given relative to the
path of the INI file. This enables running the Hector binary from
any directory without having to change the CSV path in the INI file.

### DIFF
--- a/headers/input/ini_to_core_reader.hpp
+++ b/headers/input/ini_to_core_reader.hpp
@@ -51,7 +51,10 @@ class INIToCoreReader {
     private:
     //! Weak reference to a Core object that will handle parsed values
     Core* core;
-    
+
+    //! Path of the INI file
+    std::string iniFilePath;
+
     //! The exception set by value handler should an exception occur.
     //! Note that this would only be valid if valueHandler returned
     //! an error code.
--- a/input/api-example.ini
+++ b/input/api-example.ini
@@ -24,7 +24,7 @@ tu  = 49000000         ; 4.9e7 high latitude overturning, m3/s
 twi = 12500000         ; 1.25e7 warm-intermediate exchange, m3/s
 tid = 200000000        ; 2.0e8 intermediate-deep exchange, m3/s
 
-;atm_ocean_constrain=csv:input/constraints/cmip5_oceanflux.csv		; some flux record
+;atm_ocean_constrain=csv:constraints/cmip5_oceanflux.csv		; some flux record
 k = 0.2            ; ocean heat uptake efficiency  (W/m2/K)
 
 [simpleNbox]
@@ -54,18 +54,18 @@ f_lucv=0.1			; Fraction of land use change flux from vegetation
 f_lucd=0.01			; Fraction of land use change flux from detritus
 
 ; Anthropogenic contributions: direct emissions, Pg (or Gt) C/yr
-anthroEmissions=csv:input/emissions/gcam_emissions.csv	; historical emissions data only. Future emissions will be provided by GCAM
+anthroEmissions=csv:emissions/gcam_emissions.csv	; historical emissions data only. Future emissions will be provided by GCAM
 
 ; Anthropogenic contributions: land use change
-lucEmissions=csv:input/emissions/gcam_emissions.csv 
+lucEmissions=csv:emissions/gcam_emissions.csv 
 
 ; Atmospheric CO2 levels
 ; If we supply [CO2], the model will match these values
 ; Any residual between model [CO2] and model [CO2] will be put into lucEmissions
 ; If we've also supplied lucEmissions, then the residual will be eliminated by
 ; scaling the natural atmosphere fluxes (NPP, RH, ocean, etc).
-;Ca_constrain=csv:input/constraints/lawdome_co2.csv		; Law Dome CO2 record
-;Ca_constrain=csv:input/constraints/rcp45_co2ppm.csv
+;Ca_constrain=csv:constraints/lawdome_co2.csv		; Law Dome CO2 record
+;Ca_constrain=csv:constraints/rcp45_co2ppm.csv
 
 ; Initial (preindustrial) carbon fluxes
 ;boreal.npp_flux0=5.0
@@ -94,31 +94,31 @@ eps_spinup=0.001	; spinup tolerance (drift), Pg C yr-1
 [so2] 
 S0= 53841.2         ; historical sulphate from year 2000 (Ggrams)
 SN=42000			; natural sulfur emissions (Ggrams)
-;SO2_emissions = csv:input/emissions/RCP45_emissions.csv
-SO2_emissions = csv:input/emissions/gcam_emissions.csv
-;SV = csv:input/emissions/volcanic_RF.csv
+;SO2_emissions = csv:emissions/RCP45_emissions.csv
+SO2_emissions = csv:emissions/gcam_emissions.csv
+;SV = csv:emissions/volcanic_RF.csv
 
 [methane]
 M0=700.0					; preindustrial methane (ppbv) 700 ppbv IPCC, 2001 Table 6.1
 ;UC_CH4=2.746				; Tg(CH4)/ppb unit conversion between emissions and concentrations
 
-Ma = csv:input/concentrations/RCP45_ch4n2o_conc.csv
-;CH4_emissions=csv:input/emissions/RCP45_emissions.csv   ; concentration time series
+Ma = csv:concentrations/RCP45_ch4n2o_conc.csv
+;CH4_emissions=csv:emissions/RCP45_emissions.csv   ; concentration time series
 ;molarMass = 16.04       ; grams
 ;tau = 10				; lifetime in years  oucher et al 2009
 
 [ozone]
 enabled=1
 ;NOX0=0                              ;pre-industrial value
-NOX_emissions=csv:input/emissions/RCP45_emissions.csv   ; emissions time series
+NOX_emissions=csv:emissions/RCP45_emissions.csv   ; emissions time series
 
 ;CO0=0                              ;pre-industrial value
-CO_emissions=csv:input/emissions/RCP45_emissions.csv   ; emissions time series
+CO_emissions=csv:emissions/RCP45_emissions.csv   ; emissions time series
 
 ;NMVOC0=0                              ;pre-industrial value
-NMVOC_emissions=csv:input/emissions/gcam_emissions.csv   ; emissions time series
+NMVOC_emissions=csv:emissions/gcam_emissions.csv   ; emissions time series
 
-Ma = csv:input/concentrations/RCP45_ch4n2o_conc.csv
+Ma = csv:concentrations/RCP45_ch4n2o_conc.csv
 ;molarMass = 16.04       ; grams
 ;tau = 10				; lifetime in years  oucher et al 2009
 
@@ -126,26 +126,26 @@ Ma = csv:input/concentrations/RCP45_ch4n2o_conc.csv
 N0= 273.0				; preindustrial nitrous oxide, ppbv
 ;UC_N2O=4.8				; Tg(N2O)/ppb unit conversion between emissions and concentrations
 
-Na = csv:input/concentrations/RCP45_ch4n2o_conc.csv
-;N2O_emissions=csv:input/emissions/RCP45_emissions.csv   ; concentration time series
+Na = csv:concentrations/RCP45_ch4n2o_conc.csv
+;N2O_emissions=csv:emissions/RCP45_emissions.csv   ; concentration time series
 ;molarMass = 44.01     ; grams
 ;tau = 120;
 
 [forcing]
 baseyear=1750			; when to start reporting; by definition, all F=0 in this year
-;Ftot_constrain=csv:input/constraints/MAGICC_RF_8.5.csv
+;Ftot_constrain=csv:constraints/MAGICC_RF_8.5.csv
 
 [temperature]
 S=3.0 				; equilibrium climate sensitivity for 2xCO2, degC
-;tgav_constrain=csv:input/cmip5_tgav_RCP45_test.csv		; CMIP5 global temperature, RCP 4.5 
+;tgav_constrain=csv:cmip5_tgav_RCP45_test.csv		; CMIP5 global temperature, RCP 4.5 
 
 [bc]
 ; need to be in kg/yr
-BC_emissions=csv:input/emissions/gcam_emissions.csv
+BC_emissions=csv:emissions/gcam_emissions.csv
 
 [oc]
 ; need to be in kg/yr
-OC_emissions=csv:input/emissions/gcam_emissions.csv
+OC_emissions=csv:emissions/gcam_emissions.csv
 
 ; --------------------------------------------------------------------------
 ; Halocarbons in kt/yr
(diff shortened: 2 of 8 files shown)


rgieseke:
Hi,

having played a bit with Hector I would like to propose making the CSV paths in the INI files relative to the INI filepath. This would help with the following use-cases:

- Less problems when trying out Hector
The first times after compiling I cd'ed into the `source` directory where a command like
```
./hector ../input/hector_rcp26.ini 
```
would fail. Only
```
./source/hector input/hector_rcp26.ini 
```
works.

- Being able to run Hector from any location
I toyed with a little Python wrapper (probably similar to the R scripts that come along with Hector) to change some parameters, etc. This is much easier when the Hector binary does not need to be copied around or paths in the INI don't need to be adjusted.

- Packaging
It might be a bit early before Hector becomes packaged but it wouldn't actually be too difficult to do it e.g. for Homebrew. Here it would be nice as well if the CSV location would be relative to the INI file.

I hope I'm not missing something obvious and that the proposed changes make sense. If you'd accept this proposal I'm happy to work on this further if you have any suggestions (I don't have much C++ experience and had some help from a colleague ...) on how to improve or change it.

I tested all RCP scenarios for various path combinations on OS X and Linux.

rplzzz:
@rgieseke, thanks for the patch.  Overall, the concept seems like a pretty good idea.  I think there are a few details we would have to work through (like the one I flagged in 412dd67).  I'll get together with the team and see what they think.  I'm guessing that at the very least we will want to hold this at least for the v 1.2 release, since it introduces visible interface changes.

rgieseke:
Thanks for the feedback and considering it!
Making "filename" an argument to the constructor kind of developed as I tried to figure out how this could be done at all, so there might be other ways or just keeping it with `parse`.

Agreed, that this would require a minor version change since it changes the input format as well.

Let me know if you want this improved or changed in any way.

rplzzz:
Here's what I'm thinking at the moment (the others will probably have more to add when they get in -- the workday is just starting here):

* Revert to the old interface.  Since `filename` is only referenced in `INIToCoreReader::parse()` it's probably better as an argument than as a member variable.  
* Include provision for three possibilities for file names:  
  * Absolute paths:  any name that begins with '/'
  * Paths relative to working directory:  any name that begins with './' or '../'
  * Paths relative to INI file:  all other names.  
  
This gives us maximum flexibility in specifying input files and allows us to determine the convention used for each file on a case-by-case basis.  (I'm thinking of the case where you are generating many input files for an ensemble run, and some of the inputs are the same for all ensemble members, while others are unique to each member.)

I'll give it some more thought and see what else I can come up with.

bpbond:
Hmm. Right now, Hector convention follows that of every other command-line program in the universe, right? There's a working directory, an executable, and a single parameter (the INI) file that can be specified relative to the wd, or absolute. In turn, the INI file paths can also be either relative to the wd, or absolute. How does this impede running Hector from any location? I'm a little unclear on what benefits relative-to-INI would bring.

Maybe I need a strong caffeinated beverage first, and then to re-read the chain above.

Either way thanks @rgieseke !

rplzzz:
The idea is this:  Say I'm writing an INI file for a run I want to do.  What should I specify for the locations of the emissions files?  Right now I have to specify either absolute paths (and I'm not actually sure those work; I've never tried it), or I have to specify paths relative to the directory that Hector will be executed from.  However, at the time I'm writing the INI file I have _no idea_ where Hector will be run from, so I can't guarantee that Hector will be able to find the filenames I write.  What I _do_ know is where I will be saving the INI file, so if I can specify filenames relative to that, then I sidestep the problem of having to know the eventual working directory of the Hector run.

You can see an example of this if you try to run Hector from the source, rather than top-level directory:
```
[link593@constance03]~% cd ~/wrk/hector/source
[link593@constance03]~/wrk/hector/source% ./hector ../input/hector_rcp45.ini
Thu Aug 27 05:54:56 2015:NOTICE:printLogHeader: hector version 1.1
Thu Aug 27 05:54:56 2015:NOTICE:main: hector wrapper start
Thu Aug 27 05:54:56 2015:DEBUG:h_reader: h_reader created for ../input/hector_rcp45.ini
Thu Aug 27 05:54:56 2015:NOTICE:parse: Parsing ../input/hector_rcp45.ini
Thu Aug 27 05:54:56 2015:NOTICE:main: Creating and initializing the core.
Thu Aug 27 05:54:56 2015:NOTICE:main: Setting data in the core.
Thu Aug 27 05:54:56 2015:WARNING:setData: Disabling onelineocean
terminate called after throwing an instance of 'h_exception'
zsh: abort      ./hector ../input/hector_rcp45.ini
```
The proposed change would cause this run to succeed.

bpbond:
Ah, OK, that's compelling. Thanks.

I'm a bit concerned about confusing people. @rplzzz your suggestion ('./' or '../') seems reasonable wrt distinguishing relative-to-wd from relative-to-INI but we'll need to be *very* clear about how this works in the sample INI files.

bpbond:
After talking with @cahartin - this would also let users, if they want, make an INI file and its associated emissions a self-contained unit that could be easily changed (e.g. references to `emissions.csv` instead of `emissions/RCP26_emissions.csv`), which would have some nice benefits.

rgieseke:
Yes, it would also make it easier to archive and pass arond a self-contained directory with scenario and settings.

> Revert to the old interface. Since filename is only referenced in INIToCoreReader::parse() it's probably better as an argument than as a member variable.

Looking at the code again, I think the reason for doing it the way I did was, that I couldn't find a way to let the `valueHandler` function access anything else than a static member variable. It appeared to depend on the included "inih" library so I didn't want to add new function arguments, though this might be another way.

>     Include provision for three possibilities for file names:
        Absolute paths: any name that begins with '/'
        Paths relative to working directory: any name that begins with './' or '../'
        Paths relative to INI file: all other names.

Maybe only have the two cases
  - absolute path (do you care about Windows users here?)
  - relative to the INI file

That would it make it easier to document, I guess, and allow for a directory structure like

    scenario/
        emissions.csv
         settings/
              run1.ini
              run2.ini

with "csv:../emissions.csv" in the INI files.

bpbond:
I thought about that too; is there any compelling reason to allow references relative to the working directory?

rplzzz:
> Is there any compelling reason to allow references relative to the working directory?

I think that as long as we're parsing the file names anyway, it's pretty easy to add, and then if someone comes up with a use for it (perhaps in the context of job arrays on a cluster system?), then it's there.  If not, then it's easy to ignore.

On the other hand, if the concern is that it's confusing to have `inputs` different from `./inputs`, then I get that.  I'll defer to others'  views on that matter.

> I couldn't find a way to let the `valueHandler` function access anything else than a static member variable.  It appeared to depend on the included "inih" library so I didn't want to add new function arguments, though this might be another way.

I'll check, but I don't think adding an argument to valueHandler will cause any problems with inih.  Generally it's best to use arguments over member variables where possible because it eliminates precisely those kinds of dependency questions.

Meeting break is over; more later...

rgieseke:
Hi,

I think the following much simpler patch would allow relative CSV paths and "/"-starting absolute paths requiring no function signature changes. 
If there were agreement on only covering these two cases (and ignoring absolute paths on Windows for now) I'm happy to submit a new pull request or update this one. Do you have any preference on either work-flow? Or any suggestions on how to improve this approach?

```diff
diff --git a/headers/input/ini_to_core_reader.hpp b/headers/input/ini_to_core_reader.hpp
index 769ba41..bb36d62 100644
--- a/headers/input/ini_to_core_reader.hpp
+++ b/headers/input/ini_to_core_reader.hpp
@@ -51,17 +51,20 @@ class INIToCoreReader {
     private:
     //! Weak reference to a Core object that will handle parsed values
     Core* core;
-
+
+    //! Path of the INI file
+    std::string iniFilePath;
+
     //! The exception set by value handler should an exception occur.
     //! Note that this would only be valid if valueHandler returned
     //! an error code.
     h_exception valueHandlerException;
diff --git a/source/Makefile b/source/Makefile
index d1a1f67..cd8049d 100644
--- a/source/Makefile
+++ b/source/Makefile
@@ -55,10 +55,10 @@ DEPS	= $(SRCS:.cpp=.d)

 ## default target
 hector: libhector.a main.o
-	$(CXX) $(LDFLAGS) -o hector main.o -lhector -lgsl -lgslcblas -lm
+	$(CXX) $(LDFLAGS) -o hector main.o -lhector -lgsl -lgslcblas -lm -lboost_system -lboost_filesystem

diff --git a/source/input/ini_to_core_reader.cpp b/source/input/ini_to_core_reader.cpp
index f20a765..4952e81 100644
--- a/source/input/ini_to_core_reader.cpp
+++ b/source/input/ini_to_core_reader.cpp
@@ -25,6 +25,7 @@
 #include <boost/lexical_cast.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/trim.hpp>
+#include <boost/filesystem.hpp>

 #include "core/core.hpp"
 #include "data/message_data.hpp"
@@ -32,8 +33,10 @@
 #include "input/inih/ini.h"
 #include "input/csv_table_reader.hpp"

+namespace fs = boost::filesystem;
+
 namespace Hector {

 using namespace std;

 //------------------------------------------------------------------------------
@@ -62,8 +65,9 @@ INIToCoreReader::~INIToCoreReader() {
  *                         data.
  */
 void INIToCoreReader::parse( const string& filename ) throw ( h_exception ) {
+    iniFilePath = filename;
     int errorCode = ini_parse( filename.c_str(), valueHandler, this );

     // handle c errors by turning them into exceptions which can be handled later
     if( errorCode == -1 ) {
         H_THROW( "Could not open " + filename );
@@ -112,10 +115,18 @@ int INIToCoreReader::valueHandler( void* user, const char* section, const char*
             reader->core->setData( section, nameStr, data );
         } else if( boost::starts_with( valueStr, csvFilePrefix ) ) {
             // the variableName = csv:input/table.csv case

             // remove the special case identifier to figure out the actual file name
             // to process
             string csvFileName( valueStr.begin() + csvFilePrefix.size(), valueStr.end() );
+            // when not starting with a "/" consider the CSV filepath to be
+            // relative to the input file
+            if ( !boost::starts_with( csvFileName, "/") ) {
+              fs::path iniFilePath( reader->iniFilePath );
+              fs::path csvFilePath( csvFileName );
+              fs::path fullPath( iniFilePath.parent_path() / csvFilePath );
+              csvFileName = fullPath.c_str();
+            }
             CSVTableReader tableReader( csvFileName );
             tableReader.process( reader->core, section, nameStr );
         } else {
```

bpbond:
Hi @rgieseke thanks again. It's the end of a rough workweek for all of us here, but I think we can discuss and get back to you early next week. My only qualm about the above is we do (like it or not) need to support Windows, so may drag @pralitp into this discussion.

Generally though, this is an excellent suggestion--looking forward to getting it into `master`.

rgieseke:
Here too :-)
Supporting Windows shouldn't be too hard I guess, a Regex might do, or something in Boost ...

Thanks again for feedback and consideration,
have a good weekend!

rplzzz:
Echoing what @bpbond said.  Also, with regard to additional revisions, we prefer to continue pushing updates to the existing pull request, rather than opening a new one.

rgieseke:
I've updated the branch with a new version which uses the Boost function `is_relative`. This should work on Windows as well.

bpbond:
Picking this up again. We'd like to test this branch on all three platforms to check support for these three cases:
* Absolute paths: any name that begins with '/'
* Paths relative to working directory: any name that begins with './' or '../'
* Paths relative to INI file: all other names.

@rplzzz @cahartin I know you're pretty busy through midday tomorrow w/ SFA, but could we do this after that, by the end of the week?

bpbond:
Update: @rgieseke - this is a tough few days for a bunch of us, but we'll get around to fully testing this pull request. Hopefully in the next week. One thing it's also done is bumped up issue #61 ...

rgieseke:
Thanks, please note that the current patch only does

> - Absolute paths: any name that begins with '/' *(and hopefully full Windows paths as well)*
> - Paths relative to INI file: all other names.

If you all agree that

> - Paths relative to working directory: any name that begins with './' or '../'

should also be possible I'm happy to give this a try. However I think that it could be confusing and it should always be possible to do set the current working directory as the INI files are created/re-written during runs.

rplzzz:
I think we decided making './' and '../' paths relative to the input directory would add confusion for no real gain, so we dropped it.  

I will try to carve out some time to test and review tomorrow.

bpbond:
Agreed; sorry for my confusing comment above. Edited.

rplzzz:
@rgieseke, what version of Boost are you using?  Some of the Boost functions you are using don't seem to exist in my version (Boost 1.45.0)

rgieseke:
Am 02/09/15 um 15:30 schrieb Robert Link:
> @rgieseke, what version of Boost are you using?  Some of the Boost
> functions you are using don't seem to exist in my version (Boost
> 1.45.0)

On OS X I use 1.58.0, on Linux I used a downloaded version in the 1.5x
range, don't remember which, likely a recent one.

However, the wiki states:

> Hector requires Boost version ≥ 1_52_0.

(https://github.com/JGCRI/hector/wiki/BuildHector)

Version 1.5.0 of Boost changed support for boost::filesystem apparently:

> Remove Filesystem Version 2 from the distribution. Version 3 is now
the only distributed version. Those still using V2 are urged to migrate
to V3 as soon as possible.

(http://www.boost.org/users/history/version_1_50_0.html)

Maybe the incompatibility was introduced there.

bpbond:
In the best Linux tradition, I think we can now flame @rplzzz, who obviously didn't read the installation instructions before posting in this forum. :)

rplzzz:
Seeing as how I wrote them, I didn't figure I needed to read them too!

rplzzz:
Ok, I'm satisfied with this one.  @bpbond, @cahartin, do you have any concerns?

@rgieseke, we need one further thing from you to satisfy the legal folks.  Could you post a statement affirming one of the following:

* You own the copyright on the code being contributed, and you hereby grant PNNL unlimited license to use this code in this version or any future version of Hector.  You reserve all other rights to the code.

OR

* Somebody else owns the copyright on the code being contributed (e.g., your employer because you did it as part of your work for them); you are authorized by that owner to grant PNNL an unlimited license to use this code in this version or any future version of Hector, and you hereby do so.  All other rights to the code are reserved by the copyright owner.

Sorry for the hoops, but since this PR has original code in it, we need to make sure we have it on record that you intend to let us use and distribute your contribution.

bpbond:
I'm good.

cahartin:
Good with me.

rgieseke:
Hm, I took your GPL-licensed code, amended it and offered it to you for inclusion via a pull request. So I think that of course you, PNNL or anyone else is free to use it under the terms of the GPL in this or any future version of Hector.

Does that help?

rplzzz:
Fair enough.  I hadn't considered your code as a derived product of Hector (and therefore subject to the GPL by default), but that does seem like a reasonable interpretation.

rgieseke:
Excellent! Thanks for your open-ness in all regards :-)

rplzzz:
Sure.  To further elaborate, github is a little vague about what rights a person submitting a pull request is giving up.  I'm trying to arrange it so that what is given up is the _minimum_ possible to make the contribution viable.  On top of that, if people are working on Hector on their employer's time (and for professional climate scientists, it seems likely that they will be), we need to make sure that they understand that they have to get their employer's approval for what they are doing.

Anyhow, apologies for the awkwardness.  Obviously, none of us are used to dealing with these kinds of issues, but once we've all been through it a few times it should become routine, and we won't even have to stop to think about it.

rgieseke:
No problem at all, I understand the problem and the potential problems of open source code in research institutions ... however, this pull request's code was written on my own time.
Maybe having something in the "Contributing" notes as discussed in issue #92, once this has streamlined a bit, will help.

Issue #111, opened 2015-08-26
Terminology question on CO_2 emissions


rgieseke:
In Hector CO_2 emissions are declared like 
```INI
; Anthropogenic contributions: direct emissions and land use change, Pg C/yr
anthroEmissions=csv:emissions/RCP26_emissions.csv
lucEmissions=csv:emissions/RCP26_emissions.csv 
```
Having worked mostly with MAGICC before (I work at PIK in a group originally started by Malte Meinshausen) where they are called "FossilCO2" and "OtherCO2" I was a bit confused by the "anthro" part. 
Isn't land use change also (at least partly) "anthropogenic"?
In the RCP database they are called "Fossil fuels and industry" and "Land-use change".

bpbond:
That's a good point--I agree, `anthroEmissions` is ambiguous. I'd support renaming this to `fossilEmissions`, or something like that, especially if that makes the corresponding RCP category clearer.

I'd suggest that for now (starting with upcoming 1.2) we allow either `anthroEmissions` or `fossilEmissions`, but print a warning message that the former is deprecated. At version 2.0, we drop `anthroEmissions` entirely.

@cahartin ?

cahartin:
@bpbond  I agree with those changes.   AnthroEmissions doesn't really help the user at figuring out what they are exactly. I think for a v1.2 a total name change is a good idea.

rgieseke:
How about

CO2_ffi_emissions or CO2_fossil_emissions
and
CO2_luc_emissions

to be more similar to the other gases? Ending all gases with "_emissions" would also simplify dropping "_emissions" from the column names when reading the CSV into a R or Python/Pandas data frame.

bpbond:
Addressed; closing.

PR #116, opened 2015-09-17
Rename anthroEmissions 

rgieseke:
To get started on #111 I looked into what would need to be changed. I hope I found everything and am happy to update this if you prefer any other naming scheme. I chose renaming them from `anthroEmisisons` to `ffiEmissions`.

This is in line with RCP data where the category is called 'Fossil fuels and Industry'. In the CSV files the fossil and land use emissions are renamed from ffiEmissions and lucEmissions to ffi_emissions and
luc_emissions for consistency with other categories.


bpbond:
Thanks @rgieseke. This looks good and makes a lot of sense; @kvcalvin notes it's entirely consistent with Integrated Assessment terminology too.

The one issue is that this is a *major* change, in that it changes both the model's inputs and outputs. Per our use of [semantic versioning](http://semver.org), it should be merged into a (not yet existing) v2 branch, not the current master. @rplzzz ?

rplzzz:
@bpbond Yes, I agree.  We can create an rc2.0 branch.  As far as I know, there is no way to retarget a pull request, but it's easy enough to do the merge by hand.

rplzzz:
@bpbond Did we decide that this was ready to go onto the (yet to be created) rc2.0 branch?  Or is there still stuff to do on it?

bpbond:
Ready to go afaik. Note though we (w/ @chartin ) at least) still need to discuss branch management at some point though.

rplzzz:
Yeah. We also need to figure out how we're going to manage the documentation, since it lives in the wiki, not the repository. And we'll need to come up with a plan for the next release. 

I'll make the new branch and resolve the conflicts with this one this afternoon, so at least that part will be ready to go.

rgieseke:
Some comments/questions:

- What's actually covered by LUC?
There is "land use, land use change, and forestry" (LULUCF) but also "Agriculture, Forestry and Other Land Use" (AFOLU)[1]
Not sure which one is the most appropriate ...

- Column names
SOx and NH3 have no trailing "_emissions" in their name. I'd either remove "_emissions" for all (the csv file is likely already named "emissions_scenname") or add it to those two as well for consistency.

Happy to make any amendments to this pull request as you see fit ...

[1] http://theredddesk.org/encyclopaedia/agriculture-forestry-and-other-land-uses

rplzzz:
Since Hector takes its LUC emissions as inputs, "LUC" includes whatever the user chose to include in their input data.  When you use hector with GCAM the emissions you get are AFOLU.

rgieseke:
Should it be called AFOLU then, as this seems to be the broadest category?

cahartin:
anything that does not end in _emissions does not get read into Hector and should be deleted from the emissions files.

rplzzz:
@rgieseke I'll leave that to @bpbond and @cahartin to decide, but personally I think the potential for confusion is small enough to deal with by just adding clarifying notes to the documentation, rather than by generating another change to the API.

rgieseke:
Sure, this makes sense. Just thought to mention it now, as the input files need changes anyway ...

Chartin:
I'm not the chartin you're looking for. Remove me from the thread.

bpbond:
Sorry about that @Chartin . Just click "unsubscribe".

Chartin:
Click where?

On Tue, Nov 10, 2015 at 2:55 PM, Ben Bond-Lamberty <notifications@github.com

bpbond:
@Chartin : here's a link to this discussion https://github.com/JGCRI/hector/pull/116#issue . On the webpage, look on the right side for the "Unsubscribe" button. Click it. Hope this resolves it!

rplzzz:
This branch has been merged into the rc2.0.0 branch.  It will be merged into master when we are ready for the 2.0 release.

PR #121, opened 2015-11-19
Project lib target

pralitp:
Updates to Xcode and visual studio projects to include "lib" targets to hector.  This is necessary to be able to link to GCAM or otherwise.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,10 @@
+Hector 1.1.2
+============
+
+* Add libhector target OS X and Windows project files (it was already
+  available in the Linux build).  This library library is used when
+  you want to link Hector to an outside program.
+
 Hector 1.1.1
 ============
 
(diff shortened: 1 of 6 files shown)


rplzzz:
Is it normal for these CI checks to be failing, or did these changes break them?

bpbond:
Not normal. Where did these CI checks come from? I messed around with Travis for this repo ~1 month ago, but on a branch that was subsequently deleted. @pralitp did you add CI? Or maybe I pushed something to _master_ by mistake. (Currently ooo and can't easily check.)

rplzzz:
I'll dig around and see what I can come up with.

pralitp:
I don't know anything about it, it just showed up. This pull request is
just project files changes and no code changes.

bpbond:
Ok I must screwed up. I'll go delete the Travis entry entirely.

rplzzz:
Yeah, the deal is that once you enable the CI hook, github is going to try to run the service every time an event occurs.  Deleting the CI branch doesn't stop that from happening; it only ensures that the CI build can't possibly succeed because the CI configuration files (the travis.yml mentioned in the logs for the failed checks) have been removed.

I've deleted the Travis hook, so presumably we won't get any more of these messages.  I'm not sure what it takes to get the one here to go away.  The important thing is that it doesn't seem to be related to whether or not the code actually builds; it's being driven by whether or not the CI system is configured, which it isn't.

PR #134, opened 2016-01-19
Xcode project file updates

bpbond:
Project file to Xcode 6.3; C++ runtime used is now compiler default (clang) instead of GNU C++; cleaned up header search paths; removed specific Mac OS version dependency; added linker flags for boost `filesystem` and `system`. This allows building on Mac OS X 10.11, Boost 1.60, and GSL 2.1.


bpbond:
Merging and closing.

Issue #140, opened 2016-01-27
Hector ocean temperature too high


cahartin:
the ocean temperature in Hector is too high compared to CMIP5 results.

cahartin:
branch fix-ocean-temp addresses this issue.  we scale Tgav by 30% to get a better representation of ocean temperature.

PR #141, opened 2016-01-27
Fix ocean temp

cahartin:
this addresses issue #140


rplzzz:
I have tested these changes on linux, and they seem to work.  Since this will change the model outputs, we will probably want to update the sample files in the output directory.

rplzzz:
@cahartin I removed the commit that added the unwanted data file.  Could you double check to make sure that nothing we actually wanted was lost?  Also, you will need to resync the branch in your repository if you intend to do any additional development (such as updating the sample outputs).

cahartin:
@rplzzz good point on the output files.

cahartin:
closing this pull request since we have the addition of DOEClim

PR #143, opened 2016-02-18
Remove input variables which are not being used

rgieseke:
- 'SOx', 'NH3', 'C6F14', 'HALON1202'


bpbond:
Thanks @rgieseke - it makes sense to strip out unused inputs from the Hector input files. I have checked that the new input files in this commit run fine with the Hector `master`, and confirmed with @rplzzz that this doesn't cause any issues vis-à-vis Hector's GCAM integration. Merging and closing.

PR #148, opened 2016-07-25
Make the GSL library optional

pralitp:
Add alternatives to GSL, mostly from boost (header only), for all of the GSL algorithms used by hector so that we may make it optional. Users can configure using/not using GSL by setting the preprocessor definition HAVE_GSL to 1. This opens the door for releasing hector under a license other than GPLv2.

### DIFF
--- a/headers/core/carbon-cycle-solver.hpp
+++ b/headers/core/carbon-cycle-solver.hpp
@@ -25,7 +25,9 @@
  */
 
 #include <string>
+#if HAVE_GSL
 #include <gsl/gsl_odeiv2.h>
+#endif // HAVE_GSL
 
 #include "core/logger.hpp"
 #include "models/carbon-cycle-model.hpp"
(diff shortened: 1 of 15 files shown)


pralitp:
I have verified these changes do not affect model results and can provide diagnostic figures to show the master results match (to at least a few decimal places) results generated by this branch with GSL turned on and turned off.

We may want to do a little more clean up particularly in the carbon cycle solver (do we need the commented out code still) and sea level rise (do we really need to calculate derivatives here).

Also how should I set up the project files?  Create new targets with and with out GSL?

rplzzz:
If the GSL is no longer needed, is there any advantage to keeping it as a build option?  Is it, for example, faster than the Boost alternative or something of that sort?  If not, then maybe we should consider removing the GSL entirely.

pralitp:
Hmm, I guess I'm not sure why we would want to keep both options.  In terms of speed just doing a time source/hector input/hector_rcp85.ini yields a pretty similar time.

What I'm not sure about is if there is a scenario that gives boost a hard time to solve but GSL can handle.  But again in the 8.5 scenario they seem to perform pretty similarly.

rplzzz:
The ODE solvers that we are using in Hector are explicit (Runge-Kutta) schemes, so solver failure isn't an issue the way it is in, say, the GCAM solver, where you have to iterate to convergence.  So, if we're not seeing a performance difference, I say we dump the GSL entirely.

pralitp:
Right, that makes sense.  Maybe the thing to do then is just close this pull request and I'll replace it with one that is more appropriately named: Replace GSL with boost or something like that.

PR #149, opened 2016-07-26
Replace GSL algorithms including ODE integration with various libs from

pralitp:
boost (header only).  This opens the door to release hector under
different licenses that GPLv2.

### DIFF
--- a/headers/core/carbon-cycle-solver.hpp
+++ b/headers/core/carbon-cycle-solver.hpp
@@ -62,8 +62,6 @@ public:
     
     //! Return a carbon pool value. Components will know which one they want.
     double cpool( int i ) const { return c[ i ]; }
-    //! Vector of all carbon pool values
-    //const double *allcpool( void ) const { return c; }
     
     
     // IModelComponent methods
@@ -98,8 +96,6 @@ private:
     int nc;
     //! Array of carbon pools (and such other vars as need to be integrated)
     std::vector<double> c;
-    //! Temporary array for carbon pools (for 2nd order integration)
-    //double *cc;
     //! time counter
     double t;
     
--- a/headers/core/carbon-cycle-solver.hpp
+++ b/headers/core/carbon-cycle-solver.hpp
@@ -25,7 +25,6 @@
  */
 
 #include <string>
-#include <gsl/gsl_odeiv2.h>
 
 #include "core/logger.hpp"
 #include "models/carbon-cycle-model.hpp"
@@ -50,7 +49,7 @@ namespace Hector {
  *
  * Therefore, we adopt a two-stage integration.  The easy/rapidly
  * varying variables we integrate using a standard integrator from
- * the GSL.  To integrate from t_i -> t_i+1 we first integrate
+ * the odeint.  To integrate from t_i -> t_i+1 we first integrate
  * to t_i+1/2.  We update A(t=t_i+1/2), then we integrate
  * t_i -> t_i+1 using the updated A values.
  *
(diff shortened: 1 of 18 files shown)


bpbond:
This is a significant PR (nice work @pralitp !), replacing the GSL ODE solver with equivalent calls to `boost::math:tools`, thus affecting `oceanbox`, `ocean_csys`, `carbon-cycle-solver`,  and the time series interpolation code.

Let's make sure we've thought through things before bringing it into the main branch. Pralit, you've tested results using the Boost ODE solver against the original GSL ODE solver, right? Everything looks good? Any other issues for us to consider (e.g. limitations of Boost versus GSL) that we should be aware of before making this switch? Thanks.

cc @rplzzz @cahartin

pralitp:
Yup, I've tested it out.  I've got diagnostic figures I can supply to show it reproduces (out to the last couple of decimal places at least).  I've also verified that it builds on linux/mac/VS.  I have also verified that the performance difference between the two is also negligible.

headers/core/carbon-cycle-solver.hpp
```diff
@@ -98,9 +97,9 @@ class CarbonCycleSolver : public IModelComponent {
     //! Number of variables to integrate
     int nc;
     //! Array of carbon pools (and such other vars as need to be integrated)
-    double *c;
+    std::vector<double> c;
     //! Temporary array for carbon pools (for 2nd order integration)
-    double *cc;
+    //double *cc;
```
rplzzz: Commented code:  remove if no longer needed.

pralitp:
Oops, yes that was on the task list from the previous pull request.  There is a lot of commented out code in carbon-cycle-solver.  Is that still needed?  If so we might need the cc variable and if not.. well I could just clean it all out.

headers/data/h_interpolator.hpp
```diff
@@ -62,64 +64,42 @@ class h_interpolator {
...
-                return;         // don't set ilast to iprev; since iprev<0
-            else if(xdata[iprev] <= x) {
-                ilast = iprev;  // iprev >=0 in this case
-                return;
-            }
-            else
-                iprev = 0;      // fall through to search
-        }
-    }
-
```
rplzzz: Was there something wrong with this code?  It was there to handle the case (common in ODE integrations) that the interval in a table interpolation moves slowly forward through the table, with occasional backward steps.  In practice it doesn't seem to be a particularly important optimization in Hector, so I'm not opposed to removing it; I'm just wondering what the motivation was.

rplzzz:
RE: the commented-out code, I'd just scrub it all.  If we're ready to approve the PR with it inactive, then clearly it isn't _that_ important.  If for some reason we need to recreate it, we can find it in the history.

Other than that, everything looks good to me. :shipit:

pralitp:
I changed it because if your x values was located at say xdata[3] depending on if you moved forward or backwards in the table your iprev, inext may be (3, 4) or (2, 3).

It didn't really matter in practice since 1) tseries wouldn't even call the interpolator if x fell directory on a xdata value.  2) The linear interpolation would still give the same answer.

However for trying to calculate derivatives I was trying to replicate the logic used in slr_component that if you fall on a xdata value you use the average of the linear segments to the left and right.  The logic was getting messy not knowing if I was going to get (3,4) or (2,3) so I decided I would just use the same implementation from the spline interpolation.  I believe it should be optimized in the same way as the previous implementation but is consistent about which segment it will return.

Note that ideally we would change the spline calculations to use the locate function in h_interpoloator but then the implementation would look different than the fortran original (and it doesn't even look like we are using this thing) so I didn't want to change it.

pralitp:
Great, I'll clean up the commented out code and push up the change.

bpbond:
@pralitp @rplzzz - this all sounds good to me, including removing commented-out code. Thanks.

bpbond:
Note that the documentation will need to be updated to remove GSL instructions.

pralitp:
Ok, documentation (and the wiki submodule tags) have now been updated as well.

cahartin:
i double checked it works on a PC with Visual Studio.

PR #156, opened 2016-09-21
Remove extra visual studio and Xcode project files.  

### LINKED ISSUE #150: Clean up duplicative project files
There are duplicate project files for both Visual Studio and Xcode.  We should remove _out of date_ versions.
For Visual Studio it is pretty clear we need to remove the one under VS15.

For Xcode it appears the one we want is under project_files/Xcode/hector.xcodeproj.  However we should be careful because it appears there may be some casing issues (Xcode vs xcode) which is confusing Git.

pralitp:
This closes #150.


bpbond:
Looks good. Thank you @pralitp !

PR #158, opened 2016-10-14
Add Travis CI (continuous integration) support

bpbond:
Adds continuous integration, so that when someone submits a PR or commit, we get immediate feedback if it 'breaks' the model. Right now, this just checks whether the model builds on Ubuntu, but in the future we can add a OS X/clang build, check that it runs RCP4.5 test case, run dedicated testing code, etc.

Addresses issue #61 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,18 @@
 language: cpp
 
 before_script:
-  - sudo apt-get update -qq
-  - sudo apt-get install libboost-dev libboost-filesystem-dev libboost-system-dev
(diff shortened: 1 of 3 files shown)


bpbond:
We now build against both `gcc` and `clang`, and run the RCP 4.5 case after building.

bpbond:
This PR adds a `.travis.yml` file and a single line (for the status badge) in the README. No code, data, or other scripts are changed. So I'm just going to merge this...and then enable required reviews on `master` branch so no one can ever do this again. :)

PR #162, opened 2017-01-02
bug fixes for dev_doeclim

bvegawe:
Small bug fixes. Built and ran in Xcode.

Figure shows effect of switching from _temperature_ to _temp_doeclim_ as temperature component.
[hist_comparison_tgav_doeclim.pdf](https://github.com/JGCRI/hector/files/681119/hist_comparison_tgav_doeclim.pdf)


source/components/temp_doeclim_component.cpp
```diff
@@ -235,7 +234,10 @@ void TempDOECLIMComponent::prepareToRun() throw ( h_exception ) {
     
     
     // Initializing all model components that depend on the number of timesteps (ns)
-    ns = core->sendMessage( M_GETDATA, D_END_DATE ) - core->sendMessage( M_GETDATA, D_START_DATE ) + 1;
+    ns = core->getEndDate() - core->getStartDate() + 1;
+    if( ns > 1000 ){
+        H_THROW("TempDOECLIM: Number of years larger than hard-coded max (1000). Change in temp_doeclim_component.hpp")
```
rplzzz: This error is weak.  If the reason for the limit is something in _our_ code, then we need to just fix it using vectors that can be resized at run time.  

On the other hand, if the reason for the limit is fixed-size arrays in DOECLIM (or other code we can't change), then at a minimum we need to change the limit value to be a constant defined in a single place.  As it stands, if a user changes the value, they have to trawl through the code looking for all occurrences of the string "1000", change them too, and keep their fingers crossed that they got them all.  More importantly, if the reason for the limit is a limitation in DOECLIM, then trying to change it by upping the limit in Hector probably won't work, and if that's the case, then the error message should indicate as much.

source/components/temp_doeclim_component.cpp
```diff
@@ -248,7 +250,7 @@ void TempDOECLIMComponent::prepareToRun() throw ( h_exception ) {
     //double KTB2[ns];
     //double KTA3[ns];
     //double KTB3[ns];
-    
+
```
rplzzz: All these lines of commented-out code need to be dealt with.  If we don't need them, then delete them.  If we do, then enable them (and don't make them fixed-size arrays).

Also, on line 243 we define our own value of pi.  Don't do that.  Use the ones in the standard math libraries.

source/components/temp_doeclim_component.cpp
```diff
@@ -422,7 +423,8 @@ void TempDOECLIMComponent::run( const double runToDate ) throw ( h_exception ) {
     
     // Some needed inputs
     //double forcing = core->sendMessage( M_GETDATA, D_RF_TOTAL ).value( U_W_M2 );
-    int tstep = runToDate - core->sendMessage( M_GETDATA, D_START_DATE );
+    //int tstep = runToDate - core->sendMessage( M_GETDATA, D_START_DATE );
```
rplzzz: Don't leave commented-out lines of code in the final product.  If we don't need them anymore, delete them.  You don't want to wind up featured on the next episode of Code Hoarders.

bvegawe:
Thanks for reviewing. Suggested changes have been made.

Issue #164, opened 2017-01-17
Better tests!


bpbond:
After fixing #147 - it would be great to include more and better tests, now that we have Travis CI working. To-do list includes:

- [x] All four RCPs
- [x] Atmospheric CO2 constraint run
- [ ] Ocean-atmosphere C exchange constraint run
- [x] Temperature constraint run
- [x] Radiative forcing constraint run
- [x] Change run start and stop dates
- [x] Spinup and not spinup

bpbond:
Probably create a `travis_tests/` directory, and the script runs any INI files inside of it.

bpbond:
Tagging @cahartin re the ocean-atmosphere constraint run missing file - see PR #165

rplzzz:
FYI, we now run an extensive suite of tests including all of the standard scenarios, changes to the major model parameters, and changes to emissions.  Does that answer the mail for this issue?

bpbond:
I think it's pretty solid at this point, agreed.

PR #165, opened 2017-01-17
Update .travis.yml to run all four RCPs

bpbond:
See issue #164


bpbond:
The `input/hector_rcp45.ini` file refers (on line 38) to a "constraints/cmip5_oceanflux.csv" file for constraining Hector to a particular ocean-atmosphere flux record, but this seems to be missing. @cahartin do you have a copy of this?

Issue #172, opened 2017-01-24
Accept double data values in components' setData


swillner:
Currently, most components use `valueStr` of the `message` in `setData`. This is, however, very inefficient, when they are already given as `unitval` with double value (`isVal`). I have changed that, but do not know which brunch to put the pull request on.

bpbond:
Hi @swillner please target `rc1.2`.

swillner:
closing, discussion in #173

Issue #79, opened 2015-02-18
Units in input files


bpbond:
@pralitp implemented this, and provided a demonstration of it (as I remember). We want to extend this to all inputs–units can be (optionally) provided as a check.

bpbond:
See also #88

rgieseke:
Is this syntax an example of this check?
https://github.com/JGCRI/hector/blob/master/input/hector_rcp26.ini#L183

    H0=35.0,pptv ; preindustrial concentration, pptv

bpbond:
Yes.

rgieseke:
Thanks! Then we will try to have a similar option in our wrapper (cc @swillner) !

bpbond:
This seems like it'd be laborious and not worth the gain. OK to close @kdorheim ?

PR #173, opened 2017-01-24
Accept double data values in components' setData

swillner:
As adressed in #172.
Also made public `unitsName` and `parseUnitsName` of class `unitval` as these are handy also for the outside.
The added method `expecting_unit` allows for setting emissions using double, but - as done with strings - with undefined units (created with `U_UNDEFINED`).


bpbond:
@swillner I need more clarification on what problem this is fixing. It _looks_ to me like this weakens one of the main functions of the core/coupler and messaging system: components have to assert that they know what units are associated with a datum they're requesting. This is a [really important check](http://www.cnn.com/TECH/space/9909/30/mars.metric/index.html?_s=PM:TECH).

Can you provide a specific example of how the current system is problematic? In #172 you mention efficiency, but I need to have better documentation, and understanding, of how this is a problem. Thanks.

source/components/ch4_component.cpp
```diff
@@ -104,25 +104,22 @@ void CH4Component::setData( const string& varName,
     try {
          if( varName ==  D_PREINDUSTRIAL_CH4  ) {
             H_ASSERT( data.date == Core::undefinedIndex() , "date not allowed" );
-            M0 = unitval::parse_unitval( data.value_str, data.units_str, U_PPBV_CH4 );
```
swillner: The following leads to a 'bad lexical cast'-exception (since given as unitval, not string) in the original code:
```
double value = 1.0;
Hector::unitval uval(value, U_PPBV_CH4);
Hector::message_data data(uval);
hcore.setData(CH4_COMPONENT_NAME, D_PREINDUSTRIAL_CH4, data);
```

swillner: Hey @bpbond, all my PRs come from @rgieseke 's and my work on a Python and a JS wrapper for hector (https://github.com/swillner/pyhector). I do not what to change your system, only make changes that help a wrapper or other API users ;). This PR now allows sending a `unitval` initialized with a double to a component's `setData` (most of these currently assume the values to be in string). Concerning checks for units situation is the same:
In rc1.2, a value could be given without unit in a string and is equipped with the proper unit in `unitval::parse_unitval` (or, with a uit given, the unit is checked).
In the PR, a value can be given as either string or double (with unit U_UNDEFINED or given) and is equipped with the proper unit in `unitval::set`/`tseries::set` (or unit checked there).
Advantage of the string/double-agnostic `setData` is improved performance, e.g. when providing emission data as number array by preventing unnecessary string<->double conversions. Also, then, there is a single point in allowing/disallowing undefined units (`unitval::expecting_unit`).
All in all, thanks for your wonderful and fast model!

rplzzz: Ok, here are my thoughts on this:

1. We _definitely_ want to allow a `unitval` initialized with a double and a unit to be passed through `sendMessage`.  If we can't do that already, then that's a bug (and thanks for finding and fixing it!)
2. We _don't_ want to allow a caller to pass a `unitval` initialized without the correct unit to through `sendMessage`.  Our philosophy on units is that you don't have to use Hector's unit checking system to interface with Hector, but you do have to show that whoever programmed the interface knew what the expected unit was (and, equally importantly, that the unit hasn't changed in some recent update -- I won't mention the name of the teammate that changed the units from under me a few years back).  <p> Now, it's true that we make an exception to this philosophy for values passed in as strings.  Arguably we shouldn't do that, but values passed as strings _should_ come exclusively from input parsing, and parsing a value and a unit complicates the parsing a lot, especially when the value is coming from a csv table (because in this case the unit appears only in the header line of the table, so you have to match it up with each row).  Arguably we should bite the bullet and parse units on input anyhow because right now if we were to change the unit on an input quantity in the code, the system would silently parse input with the wrong units.  However, fixes like that have taken a back seat to actual science improvements.  You know how it is.
3. You mentioned passing `unitval`s to `setData`.  The `setData` method really shouldn't be called from outside the component.  The fact that it is public instead of private is a holdover from Hector's early development, before the core and its message routing were established.  There has been a TODO comment in that method suggesting that we track down and eliminate the remaining external usages, but it also hasn't been a priority for the same reasons as discussed above.  In practical terms, however, any new code should communicate with Hector components via `Core::sendMessage`.  If there is some use case that we can't support with `sendMessage`, then we should flag that so we can fix the missing functionality, rather than circumventing the intended communication pathway.

Does that clear things up at all?

swillner: @rplzzz Thanks for the clarification! Now I understand differences in handling the string values in the components' `setData` method (depending on if it's an input set via `sendMessage` or not). From the API standpoint I need, however, a way to set input data without the ini-parser, that's why I call the core's `setData` method (also, it's faster if I can cache the component for setting arrays of data without using the string matches in the message system).

rplzzz: So, as I understand it, you are trying to create an alternative front end for Hector, and you want to be able to initialize all of the Hector components _without_ having them read the ini files.  I ran into a similar problem when coupling Hector to GCAM, and in the interest of getting something working quickly, I just punted and had Hector parse an ini file I had created specifically for use with GCAM, but that's not a great solution.

I think the right thing to do here (i.e., what I should have done when I first did the GCAM coupling) is to add some initialization functions to the API.  We could allow these to accept arrays of doubles, with the unit specified just once for the whole array.  We would probably also want to make some provision for designating one (or more?) of the arrays as a key that indexes the others (e.g., the year column in our emissions tables).  We might also arrange for these parts of the API to be disabled once the simulation is running, which would prevent someone from trying to use it inappropriately.

Thoughts?

bpbond: >Arguably we should bite the bullet and parse units on input anyhow 

See issue #79 

@rplzzz summed up things well, and thanks for the nice compliment @swillner !

bpbond:
Thanks to both @swillner and @rplzzz for explanations and thoughts. I'd propose to merge this, with the understanding that we will revisit this issue for v2.0 (in a way that may or may not break your wrapper, @swillner @rgieseke ). To this end, I'll open an issue with the array API proposed by @rplzzz . Sound OK?

swillner:
@bpbond Great! We expect to have to adjust our wrapper to v2.0 anyway, so no worries there.
@rplzzz To circumvent using INI files we immitate the ini reader and csv writer (https://github.com/swillner/hector-wrapper), which lead us to use `setData` directly...

cahartin:
@rplzzz can i go ahead and merge this into rc2.0?

rplzzz:
Yes.

rplzzz:
This is showing up as "closed" instead of "merged".  What happened here, and can we verify that the branch got merged into wherever it was supposed to go?

cahartin:
i don't know what i just did.  It didn't get merged into rc2.0

cahartin:
ok that worked. odd.

rplzzz:
It looks like branch rc2.0.0 got merged into this branch.  Perhaps the branch was showing up as out of date, and the first time around you updated it.

Issue #171, opened 2017-01-24
Make logging into files optional


swillner:
Currently, logging is always writing into log files. Since that takes some amount of running time, it would be great to have that optional.
I have a code solution proposal on that, but do not know which branch to use for the pull request. Also, I would include a compile-time switch for writing logging to files for the debug loggers used in the components.

bpbond:
This is somewhat related to #69 (which proposes, among other things, a 'quiet' command line switch).

I very much agree about the need for this functionality. @swillner can you briefly describe your proposal here? I think we'd ask you to target the `rc1.2` branch.

bpbond:
Note that we'd really like a solution that's runtime (INI or command line setting), not only a compile-time one, I think. @rplzzz ?

rplzzz:
Do we not have logging levels that we can use to control how much logging data is written?  If so, then setting the log level to something higher than the highest log severity normally used would have the effect of suppressing all logging.

swillner:
Most components setup their own logger on the DEBUG level on creation. Also, independent of the level the logging directory and files are created.

rplzzz:
Obviously we need to get the logging level from the user configuration.  The hardwired debug-level was probably put in as a placeholder, and revising it never made it onto anyone's radar.  We could also check for debug levels high enough to preclude all output and suppress creating the empty file in those cases.

swillner:
I suppose, it would be great to also allow for only stdout/stderr logging without writing into files.

PR #175, opened 2017-01-24
Make logging into files optional

swillner:
as a suggestion for #171.

### DIFF
--- a/headers/core/logger.hpp
+++ b/headers/core/logger.hpp
@@ -63,6 +63,9 @@ private:
     
     //! Flag to indicate that this logger has been successfully opened.
     bool isInitialized;
+
+    bool echoToFile;
+    bool enabled;
     
     //! The actual output stream which will handle the logging.
     std::ostream loggerStream;
@@ -103,7 +106,7 @@ public:
     ~Logger();
     
     void open( const std::string& logName, const bool echoToScreen,
-              LogLevel minLogLevel) throw ( h_exception );
+              LogLevel minLogLevel, const bool echoToFile = false) throw ( h_exception );
     
     bool shouldWrite( const LogLevel writeLevel ) const;
     
--- a/headers/core/logger.hpp
+++ b/headers/core/logger.hpp
@@ -64,7 +64,11 @@ private:
     //! Flag to indicate that this logger has been successfully opened.
     bool isInitialized;
 
+    //! Flag to indicate that this logger writes into a log file.
     bool echoToFile;
+
+    //! Flag to indicate that this logger is enabled.
+    //! If false this logger does not log regardless of log level provided.
     bool enabled;
     
     //! The actual output stream which will handle the logging.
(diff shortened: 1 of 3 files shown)


swillner:
I could adjust that to use a runtime option, just let me know

pralitp:
Aside from the comments these changes look fine to me.  Although these changes alone are not sufficient to take care of #171, right?  Unless I am missing something we would still need to include some means to set that `echoToFile` to false in each model component.

pralitp:
Oops, I just noticed that by default you have `echoToFile = false`.  Meaning by default we do not write log messages to file and currently there is no way (other than explicitly changing it at compile time) to renable them.  I'm not sure this is the behavior we want.  @bpbond, thoughts?

rplzzz:
I think the problem here is that we don't have a clear idea of how user control of the logging is _supposed_ to work.  We should write the instructions for how to control the logging, and then we can update the code to make it work that way.

swillner:
As I understand your approach right now, you control the logger at runtime in the "client" code and not via INI, which makes since in my opinion. I would propose to:
1) Have the components' logger inherit the global logger's logLevel and output settings (file yes/no, stdout yes/no)
2) Equip `class IModelComponent` with `Logger logger` (all components use one, anyway) and expose its settings to the outside to override those from 1) if necessary
3) Publicly exposing `getComponentByName` would be handy to access the components in the "client" code, e.g. for 2)

bpbond:
I agree - the default behavior of any change should be identical to that of the current model (write log files), especially as this PR is targeting `rc1.2`, which [should not have dramatic changes](https://github.com/JGCRI/hector/wiki/VersionNumbers).

That's correct @swillner - currently the only control is at compile time.

@rplzzz 's point is a good one that we have no defined log behavior at the moment. Let's start there. Focusing on _funtionality_, not _code_, I'd propose:

* The global logger can be set to to output to a file and/or screen via INI file. Default is both, but this can be changed.
>[core]
>log_screen=1
>log_file=0
* Component loggers inherit global defaults, but can be individually set.
* A command-line `--quiet` switch overrides all logger `log_screen` settings to FALSE, i.e. regardless of the INI file settings.

Thoughts?

pralitp:
Mostly I think this is a good approach.  There is one detail which needs to be thought through however.  Currently loggers configure the log_screen and log_file during `Logger::open` and Loggers are open prior to INI file parsing (presumably to be able to be able to log parsing errors).  Maybe we just re-open the Logger when the settings change.  Do we then cascade those settings changes to the component loggers?  In addition we might we to be careful not to lose messages if for instance the log_file goes from 1 to 0 then back to 1.

swillner:
Then why not just have the INI exclusively for model parameters and control logging via command line switches (e.g., `--logdir DIRECTORY` and `--logscreen`) in the main program / by inheriting from the global logger for API users.

headers/core/logger.hpp
```diff
@@ -103,7 +106,7 @@ class Logger {
     ~Logger();
     
     void open( const std::string& logName, const bool echoToScreen,
-              LogLevel minLogLevel) throw ( h_exception );
+              LogLevel minLogLevel, const bool echoToFile = false) throw ( h_exception );
```
pralitp: Set the default = true so that this pull request does not change behavior.

pralitp:
I just discussed with @bpbond, and I think we are in agreement that a command line switch such as @swillner suggests sounds reasonable and certainly makes life easier for implementation.  That being said let's do that in it's own pull request so that we can keep this on focused on it's goal (adding the ability to turn off logging into files, even if it is a compile time switch).  I think then just taking care of the comments in the code review and this pull request is good to go.

Issue #200, opened 2017-10-13
Halocarbon forcing are being double-counted


bpbond:
Hector is double-counting the forcing effect from halocarbons. This was discovered by Felipe (Feijoo? @cahartin ?). The relevant code is lines 283-290 in `forcing-component.cpp`:

```C
        // Halocarbons can be disabled individually via the input file, so we run through all possible ones
        forcings[ D_RF_halocarbons ].set( 0.0, U_W_M2 );
        for (unsigned hc=0; hc<halos.size(); ++hc) {
            if( core->checkCapability( halos[hc] ) ) {
                // Forcing values are actually computed by the halocarbon itself
                forcings[ halos[hc] ] = core->sendMessage( M_GETDATA, halos[hc], message_data( runToDate ) );
                forcings[ D_RF_halocarbons ] = forcings[ D_RF_halocarbons ] + forcings[ halos[hc] ];
            }
        }
```

Here we walk through the halocarbon list, calling each to get its forcing, and then sticks the value into the `forcings` map TWICE. This then gets summed into `Ftot` down in line 336.

PR #201, opened 2017-10-17
fix for double counting halocarbons

cahartin:
deleted the total forcing from halocarbons so it is not added to the 
total forcing.

### DIFF
--- a/source/components/forcing_component.cpp
+++ b/source/components/forcing_component.cpp
@@ -280,13 +280,11 @@ void ForcingComponent::run( const double runToDate ) throw ( h_exception ) {
         };
         
         // Halocarbons can be disabled individually via the input file, so we run through all possible ones
-        forcings[ D_RF_halocarbons ].set( 0.0, U_W_M2 );
-        for (unsigned hc=0; hc<halos.size(); ++hc) {
+         for (unsigned hc=0; hc<halos.size(); ++hc) {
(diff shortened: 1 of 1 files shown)


cahartin:
this addresses issue #200.

cahartin:
I pulled out a random year of 2000 and plotted the difference between the old version of Hector and the new fixed Hector. We see a significant change in the total harlocarbon forcing, which has been deleted, resulting in a change in total radiative forcing, but we also see a small change in the forcing from CO2.  
rf_diff

bpbond:
The CO2 change is presumably a feedback from previous years? Halocarbons changed, changing temperature, changing the terrestrial and ocean uptakes, changing CO2?

rplzzz:
That sounds right to me.  The direction of the feedback change looks right, at least with respect to ocean uptake.  Higher temperatures should lead to less ocean absorption, which leaves more in the atmosphere.  Does terrestrial uptake go in the same direction?

cahartin:
good idea @rplzzz, I added a comment in the code.  I don't see a problem with SO2.  We have 2 entries, one for indirect and one for direct with no total SO2.

cahartin:
temp_diff


@rplzzz and @bpbond do you want to see some other variables, CO2, fluxes, etc?

cahartin:
Just to be complete - atmospheric CO2.
co2_diff

PR #206, opened 2017-11-22
Dev doeclim

cahartin:
This PR includes the a new temperature component (DOECLIM) to better represent ocean heat uptake and global mean temperature change. 

This [document](https://www.authorea.com/users/170481/articles/207125-hector-doeclim) contains more detailed information on the changes.


headers/components/component_names.hpp
```diff
@@ -26,6 +26,7 @@
 #define OH_COMPONENT_NAME   "OH"
 #define N2O_COMPONENT_NAME "N2O"
 #define TEMPERATURE_COMPONENT_NAME "temperature"
+#define TEMP_DOECLIM_COMPONENT_NAME "temp_doeclim"
```
bpbond: Should we use a more meaningful name?

headers/components/ocean_component.hpp
```diff
@@ -91,15 +89,6 @@ class OceanComponent : public CarbonCycleModel {
     bool spinup_chem;       //!< run chemistry during spinup?
     tseries<unitval> oceanflux_constrain;      //!< atmosphere->ocean C flux data to constrain to
     
-    // Heat flux
-	unitval heatflux;
-	unitval k_max;           //!< maximum ocean heat uptake efficiency, W/m2/K
-    unitval k_min;          //!< minimum ocean heat uptake efficiency, W/m2/K
-    unitval slope;           //!< slope of the curve, 1/K
-//	unitval cp;             //!< Specific heat
```
bpbond: Why is this commented out? Remove?

headers/components/component_names.hpp
bpbond: Ah, the "DOE" here doesn't refer to the department of energy. OK then.

headers/components/temp_doeclim_component.hpp
```diff
@@ -0,0 +1,150 @@
...
+    void invert_1d_2x2_matrix( double * x, double * y);
+    void sum_1d_2x2_matrix( double * x, double * y, double * z);
+    
+    //Hard-coded DOECLIM parameters
+    int dt;                  // years per timestep (this is hard-coded into Hector)
+    int ns;                  // number of timesteps
+    double ak;               // slope in climate feedback - land-sea heat exchange linear relationship
+    double bk;               // offset in climate feedback - land-sea heat exchange linear relationship, W/m2/K
+    double csw;              // specific heat capacity of seawater W*yr/m3/K
+    double earth_area;       // m2
```
bpbond: Seems like many of these should be `const`

headers/components/temp_doeclim_component.hpp
```diff
@@ -0,0 +1,150 @@
...
+    std::vector<double> KTB2;
+    std::vector<double> KTA3;
+    std::vector<double> KTB3;
+    
+    std::vector<double> Ker;
+
+    double A[4];
+    double IB[4];
+    
+    //Dependent DOECLIM parameters
```
bpbond: Space after `//`. Dependent on what? What's the difference between `ocean_area` below and the same variable name above?

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+
+
+//------------------------------------------------------------------------------
+/*! \brief              Calculates inverse of x and stores in y
+ *  \param[in] x        Assume x is setup like x = [a,b,c,d] -> x = |a, b|
+ *                                                                  |c, d|
+ *  \param[in] y        Inverted 1-d matrix
+ *  \returns            void, inverse is stored in y
+ */
+void TempDOECLIMComponent::invert_1d_2x2_matrix(double * x, double * y) {
```
bpbond: Since we depend on [Boost](http://boost.org) anyway, why not use its `ublas` library?

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+    keff = kcon * diff;
+    taubot = pow(zbot,2) / keff;
+    powtoheat = ocean_area * secs_per_Year / pow(10.0,22);
+    taucfs = cas / cfs;
+    taucfl = cal / cfl;
+    taudif = pow(cas,2) / pow(csw,2) * M_PI / keff;
+    tauksl  = (1.0 - flnd) * cas / kls;
+    taukls  = flnd * cal / kls;
+    
+    // First order
```
bpbond: Some more documentation needed here--what's going on?

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+    
+    // Commented section below is for case of user-specified temperature record.
+    // TempDOECLIM component can't handle that at the moment!
+    
+    //// We track total radiative forcing using internal variable `internal_Ftot`
+    //// Need to do this because if we're subject to a user constraint (being forced
+    //// to match a temperature record), need to track the Ftot that *would* have
+    //// produced the observed temperature record. This way there's a smooth
+    //// transition when we exit the constraint period, after which internal_Ftot
+    //// will rise in parallel with the value reported by ForcingComponent.
```
bpbond: Nice commenting and logic.

bpbond:
Note also my comment about using Boost–I doubt we want to roll out own matrix-inversion functions, but will defer to @rplzzz here.

headers/components/temp_doeclim_component.hpp
cahartin: instead of `doubles`?

headers/components/ocean_component.hpp
cahartin: thanks, i missed that one.

headers/components/temp_doeclim_component.hpp
bpbond: No, I mean, something like `const double earth_area = x.xxx;`

headers/components/temp_doeclim_component.hpp
cahartin: @bpbond If these are just `consts` then we don't actually need them in the header?

source/components/temp_doeclim_component.cpp
cahartin: @bvegawe can you add a few comments here? thank you!

source/components/temp_doeclim_component.cpp
cahartin: @bpbond i think that was you back in the day!

headers/components/temp_doeclim_component.hpp
bpbond: Mmm. Feeling fuzzy on my C++. I think actually we'd want `static const x = ...`. As for header or not, it depends; is it referred to in multiple component functions? If not, right, just move into the run method or wherever.

headers/components/component_names.hpp
cahartin: @bpbond and @rplzzz yea it's diffusive ocean energy climate model.  are there any licensing issues we need to be aware of of linking this one model with Hector? Can we still call it Hector?
https://github.com/scrim-network/BRICK

source/components/temp_doeclim_component.cpp
bvegawe: Any Boost inversion implementation I saw in my cursory google search looked a good deal more complex than this simple algorithm for 2x2 inversion. But I could be wrong!

rplzzz: That's correct.  If you know the matrix is 2x2, you can invert it very simply using analytic formulae.  It's similar to how you could solve an equation using the Newton-Raphson method, but if you happen to know it's quadratic, it's simpler to use the quadratic formula.

headers/components/temp_doeclim_component.hpp
rplzzz: `const` means that once it's initialized, it can't be changed.  `static` means that there is one copy for the entire class, instead of one for each object created.  A member can be either `const`, `static`, neither, or both, depending on what you are trying to accomplish.  For physical constants like the area of the earth or the specific heat of seawater, `static const` is the way to go.  For something that might be a settable parameter (`ak`, perhaps?), you would generally do that as a normal (i.e., not static) member.

A static member is initialized just like a global variable, irrespective of whether or not it happens to be const.  In fact, static members are basically just namespaced global variables, except that C++ does respect their `private` or `protected` visibility.

A member that is const, but not static _must_ be initialized in the constructor.  This is one of the rare cases in which the distinction between initialization and assignment matters.  In this case you have to use the `:` notation to make it an initialization instead of an assignment.  If you can't do that for some reason, such as because you have to construct the object before the initialization data is available, then you can't make the members const.  Your best bet in this case is to make them ordinary members and to add a comment to the effect that once they are set they should be treated as const.

A constant that is only used in one method could be moved into that method, as Ben suggests, but sometimes it's best to make them const or static const.  The reason why is that if someone wants to know, what are they using for the xyz parameter, it's a lot easier to find it if it's with all the other constants, rather than tucked away at the beginning of some particular method they might not know the name of.  Also, if you end up refactoring so that more than one method uses the constant, then you'll have to make the constant a member anyhow.  Starting it off that way future proofs your design a bit.

headers/components/temperature_component.hpp
```diff
@@ -1,87 +0,0 @@
-/* Hector -- A Simple Climate Model
```
rplzzz: I really don't love that we are deleting the original temperature component entirely.  The whole idea behind Hector's design is that we are supposed to be able to add and drop components as needed.  I get that we don't want components to proliferate because then we have to maintain them all, but so far maintaining components hasn't been a huge amount of effort.  Is the original temperature component really unable to coexist with the DOEClim temperature (in the code base, I mean; obviously only one of them can be active in any single run)?

source/components/ocean_component.cpp
```diff
@@ -68,20 +68,12 @@ void OceanComponent::init( Core* coreptr ) {
...
-        slope.set( -3.0, U_1_K );      // default
-        
-        m_min_k_so_far = std::numeric_limits<double>::max();
 
 	lastflux_annualized.set( 0.0, U_PGC );
         
     // Register the data we can provide
     core->registerCapability( D_OCEAN_CFLUX, getComponentName() );
 	core->registerCapability( D_OCEAN_C, getComponentName() );
-    core->registerCapability( D_HEAT_FLUX, getComponentName() );
```
rplzzz: I take it these heat flux capabilities have been moved to the DOEClim component?

Edit: Actually, it doesn't look like they have.  Are we just not supporting retrieving the ocean heat flux anymore, or is it going to known by a different name going forward?  Apart from being a serious backward-incompatible change, it would be kind of unfortunate if we could no longer get this information from the model.

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+    
+    return name;
+}
+
+
+//------------------------------------------------------------------------------
+/*! \brief              Calculates inverse of x and stores in y
+ *  \param[in] x        Assume x is setup like x = [a,b,c,d] -> x = |a, b|
+ *                                                                  |c, d|
+ *  \param[in] y        Inverted 1-d matrix
```
rplzzz: This should be documented as `param[out]` or `param[inout]`, depending on whether the function expects `y` to be initialized.

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+    y[3] = temp * x[0];
+        
+    return;
+}
+    
+//------------------------------------------------------------------------------
+/*! \brief              Calculates element-wise sum of matrix x and y, stores in z
+ *  \param[in] x,y      Assume x,y setup like x = [a,b,c,d] -> x = |a, b|
+ *                                                                 |c, d|
+ *  \param[in] z        Summed 1-d matrix
```
rplzzz: This also is `param[out]`.

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+                                    const std::string& datum,
+                                    const message_data info ) throw ( h_exception )
+{
+    unitval returnval;
+    
+    if( message==M_GETDATA ) {          //! Caller is requesting data
+        return getData( datum, info.date );
+        
+    } else if( message==M_SETDATA ) {   //! Caller is requesting to set data
+        H_THROW("TempDOECLIM: sendMessage not yet implemented for message=M_SETDATA.");
```
rplzzz: This looks like boilerplate copied from another component.  Surely we can do better than this.  If there are variables that it makes sense to set from the outside, then we should process them here.  For variables that you aren't allowed to set from outside (which might very well be all of them), the error message should be something like "Setting variable XYZ is not allowed."

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+void TempDOECLIMComponent::prepareToRun() throw ( h_exception ) {
+    
+    H_LOG( logger, Logger::DEBUG ) << "prepareToRun " << std::endl;
+    
+    if( tgav_constrain.size() ) {
+        Logger& glog = Logger::getGlobalLogger();
+        H_LOG( glog, Logger::WARNING ) << "Temperature will be overwritten by user-supplied values!" << std::endl;
+    }
+    
+    dt = 1;         // years per timestep (this is hard-coded into Hector
```
rplzzz: If this is hard-coded in Hector, then we should be getting it from wherever it is hard coded at.  What if we someday change the hard coded value?  Then this component will be inconsistent, and it will likely go unnoticed until it causes problems.  Or worse, it will silently generate incorrect results.

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+    zbot = 4000.0;  // bottom depth of diffusive ocean, m
+    bsi = 1.3;                  // warming factor for marine surface air over SST (due to retreating sea ice)
+    cal = 0.52;                 // heat capacity of land-troposphere system, W*yr/m2/K
+    cas = 7.80;                 // heat capacity of mixed layer-troposphere system, W*yr/m2/K
+    flnd = 0.29;                // fractional land area
+    fso = 0.95;                 // ocean fractional area below 60m
+    
+    
+    // Initializing all model components that depend on the number of timesteps (ns)
+    ns = core->getEndDate() - core->getStartDate() + 1;
```
rplzzz: This is assuming that `dt` is 1, right?  Since you have defined `dt` above, you shouldn't be assuming anything about its value in the rest of the code.  Calculate `ns` using `dt` as if you didn't know anything about its value.

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+    heat_mixed.resize(ns);
+    heat_interior.resize(ns);
+    forcing.resize(ns);
+    
+    for(int i=0; i<3; i++) {
+        B[i] = 0.0;
+        C[i] = 0.0;
+    }
+    
+    for(int i=0; i<ns; i++) {
```
rplzzz: Much of this could be expressed more concisely using the fill constructor for `std::vector`.  E.g.:
```
KT0 = std::vector<double>(ns, 0.0)
```
q.v.: http://www.cplusplus.com/reference/vector/vector/vector/

source/components/temp_doeclim_component.cpp
rplzzz: My guess is that this has all been explained in a paper somewhere.  We should include a pointer to the paper and document any mappings between names used in the paper and names used here that aren't obvious.

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+    }
+    
+    // Sum up the kernel components
+    for(int i=0; i<ns; i++) {
+        
+        Ker[i] = KT0[i] + KTA1[i] + KTB1[i] + KTA2[i] + KTB2[i] + KTA3[i] + KTB3[i];
+        
+    }
+    
+    // Switched on (To switch off, comment out lines below)
```
rplzzz: First, this is rather cryptic.  If there really are circumstances under which we would want to turn off this section, then we should explain what those circumstances are.

Second, we don't want users to be reconfiguring the model by hacking at the code, and we _surely_ don't want to be trying to debug their problems when they try.  If there is a legitimate reason for users to disable this section, then we need an option for that in the config, which this section would check and either do the stuff below or not accordingly.

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+    //// to match a temperature record), need to track the Ftot that *would* have
+    //// produced the observed temperature record. This way there's a smooth
+    //// transition when we exit the constraint period, after which internal_Ftot
+    //// will rise in parallel with the value reported by ForcingComponent.
+    
+    // If we never had any temperature constraint, `internal_Ftot` will match `Ftot`.
+    //
+    // If the user has supplied temperature data, use that (except if past its end)
+    if( tgav_constrain.size() && runToDate <= tgav_constrain.lastdate() ) {
+    //    H_LOG( logger, Logger::NOTICE ) << "** Using user-supplied temperature" << std::endl;
```
rplzzz: Delete this line.  Saving an old version of an error message in a commented line is perhaps the most egregious example of code hoarding I can think of.

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+        H_LOG( logger, Logger::NOTICE ) << "** ERROR - DOECLIM can't currently handle user-supplied temperature" << std::endl;
+    }
+    //  tgav = tgav_constrain.get( runToDate );
```
rplzzz: Lots of commented-out code from here to line 385.  Clean it up.

source/components/temp_doeclim_component.cpp
```diff
@@ -0,0 +1,528 @@
...
+	const double fso2i = core->sendMessage( M_GETDATA, D_RF_SO2i ).value( U_W_M2 );
+	Ftot += fso2i * so2i_b; // % extra
+	*/
+    
+    // Some needed inputs
+    int tstep = runToDate - core->getStartDate();
+    double aero_forcing = double(core->sendMessage( M_GETDATA, D_RF_BC ).value( U_W_M2 )) + double(core->sendMessage( M_GETDATA, D_RF_OC ).value( U_W_M2 )) + double(core->sendMessage( M_GETDATA, D_RF_SO2d ).value( U_W_M2 )) + double(core->sendMessage( M_GETDATA, D_RF_SO2i ).value( U_W_M2 ));
+    forcing[tstep] = double(core->sendMessage( M_GETDATA, D_RF_TOTAL ).value( U_W_M2 )) - ( 1.0 - alpha ) * aero_forcing;
+			    
+    // Initialize variables for time-stepping through the model
```
rplzzz: Are the next 80 lines or so copied/translated from the original DOEClim code base?  If so, we should maybe include a pointer to that code, both so that anyone who wants to look at it can, and also to document that it was copied with the knowledge and permission of its authors and maintainers.

headers/components/component_names.hpp
```diff
@@ -26,6 +26,7 @@
 #define OH_COMPONENT_NAME   "OH"
 #define N2O_COMPONENT_NAME "N2O"
 #define TEMPERATURE_COMPONENT_NAME "temperature"
```
rplzzz: If we are removing the old temperature component, then shouldn't we also remove this `#define`?

headers/components/component_names.hpp
rplzzz: We are using DOEClim with the permission of its authors, right?  If so, then it's fine.  We should, however, acknowledge the code they donated.  I have a note about that elsewhere.

source/components/ocean_component.cpp
bvegawe: With DOECLIM, we calculate a flux into the mixed layer and a flux into the interior. For backward compatibility, I'll make a note to also do a total heat flux.

headers/components/component_names.hpp
bvegawe: I have not done any investigations into permissions. DOECLIM has been used several times by Klaus Keller's group and I adapted their C++ implementation here. I'll check in with Ryan and Klaus about it.

rplzzz: It looks like one or more versions of DOECLIM have been released under GPLv3.  Just get verbal confirmation from one of the original authors that they're ok with us releasing it under GPLv2.  (The differences aren't really relevant to what we're doing; I like v2 because it's shorter and written in plain language.)  Then we can just indicate that the code is used with permission.

headers/components/temperature_component.hpp
bvegawe: In my experience, it works fine with both components and just specifying which version you want to use in the .ini file. It does lead to a couple warnings like below:
WARNING:registerCapability: temperature is declaring capability Tgav previously registered

cahartin: They can exist together and do currently.  My concern is that the old way was a hack with a bunch of bugs that were found. I don't think I want people having an easy option to get back incorrect results by turning on that old component.

input/hector_rcp85_doeclim.ini
```diff
@@ -0,0 +1,401 @@
+; Config file for hector model: RCP85
```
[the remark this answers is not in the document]
cahartin: good point. this was our initial test case file.

source/components/ocean_component.cpp
cahartin: Heat flux is calculated in the DOECLIM component now.  @bvegawe that's a good idea to have total heat flux into the ocean.

headers/components/temperature_component.hpp
rplzzz: That sounds reasonable.  Since that leaves the doeclim temperature component as the _only_ temperature component, maybe we should just rename it to "temperature_component".  (We would still explain its doeclim origins in comments in the source file, of course.)

headers/components/temp_doeclim_component.hpp
bvegawe: Based on @rplzzz 's comment below it looks like we should move these initializations to the header (some are called in more than one function so if we want to keep them together they need to be there). Those with well-known values will have `const static`, while those that are potentially tweakable will have `const`. Does that sound like what we want?

headers/components/component_names.hpp
cahartin: renamed component.  Still need to double check licensing.
(CH - tracking her changes)

headers/components/component_names.hpp
cahartin: done

headers/components/temperature_component.hpp
cahartin: done

source/components/ocean_component.cpp
cahartin: done

source/components/temp_doeclim_component.cpp
cahartin: @bpbond @rplzzz we should probably get this constraint working again in the new component.  for now i'll leave the comments in and we can test this later.

headers/components/component_names.hpp
bvegawe: A short update on the permissions: we now have emailed permission from Elmar Kriegler (original creator of DOECLIM) to use DOECLIM in Hector under GPLv2. I'm waiting for a response from Klaus Keller (coauthor of the paper for which the GPLv3 version was written).

bvegawe: We now also have written permission from the author of the DOECLIM version released under GPLv3

PR #209, opened 2017-11-30
Dev doeclim: improved DOECLIM descriptions, total heat flux output

bvegawe:


source/components/temp_doeclim_component.cpp
```diff
@@ -197,29 +184,29 @@ void TempDOECLIMComponent::prepareToRun() throw ( h_exception ) {
     ak = 0.31;      // slope in climate feedback - land-sea heat exchange linear relationship
     bk = 1.59;      // offset in climate feedback - land-sea heat exchange linear relationship, W/m2/K
     csw = 0.13;     // specific heat capacity of seawater W*yr/m3/K
-    earth_area = 5100656 * pow(10.0, 8);
+    earth_area = 5100656 * pow(10.0, 8);  // m2
     kcon = 3155.0;  // conversion from cm2/s to m2/yr
```
bpbond: It would be clearer to spell this out: `1 / 100 / 100 * secs_per_Year`

source/components/temp_doeclim_component.cpp
```diff
@@ -197,29 +184,29 @@ void TempDOECLIMComponent::prepareToRun() throw ( h_exception ) {
...
     q2co = 3.7;     // radiative forcing for atmospheric CO2 doubling, W/m2
     rlam = 1.43;    // factor between land clim. sens. and sea surface clim. sens. T_L2x = rlam*T_S2x
     secs_per_Year = 31556926.0;
```
bpbond: Move above line 188, and change to `60 * 60 * 24 * 365.25`

source/components/temp_doeclim_component.cpp
```diff
@@ -197,29 +184,29 @@ void TempDOECLIMComponent::prepareToRun() throw ( h_exception ) {
...
     zbot = 4000.0;  // bottom depth of diffusive ocean, m
-    bsi = 1.3;                  // warming factor for marine surface air over SST (due to retreating sea ice)
-    cal = 0.52;                 // heat capacity of land-troposphere system, W*yr/m2/K
-    cas = 7.80;                 // heat capacity of mixed layer-troposphere system, W*yr/m2/K
-    flnd = 0.29;                // fractional land area
-    fso = 0.95;                 // ocean fractional area below 60m
+    bsi = 1.3;      // warming factor for marine surface air over SST (due to retreating sea ice)
+    cal = 0.52;     // heat capacity of land-troposphere system, W*yr/m2/K
+    cas = 7.80;     // heat capacity of mixed layer-troposphere system, W*yr/m2/K
```
bpbond: Are the units on these two lines correct? 1 W = 1 J/s, right, so wouldn't things cancel? Can this be more clearly written as J/m2/K?

headers/components/temp_doeclim_component.hpp
```diff
@@ -81,40 +81,30 @@ class TempDOECLIMComponent : public IModelComponent {
     double flnd;             // fractional land area
     double fso;              // ocean fractional area below 60m
     
-    //Intermediate DOECLIM parameters
+    // Intermediate DOECLIM parameters
```
bpbond: What does "Intermediate" mean here?

source/components/temp_doeclim_component.cpp
```diff
@@ -237,31 +224,25 @@ void TempDOECLIMComponent::prepareToRun() throw ( h_exception ) {
...
-    keff = kcon * diff;
-    taubot = pow(zbot,2) / keff;
-    powtoheat = ocean_area * secs_per_Year / pow(10.0,22);
-    taucfs = cas / cfs;
-    taucfl = cal / cfl;
-    taudif = pow(cas,2) / pow(csw,2) * M_PI / keff;
-    tauksl  = (1.0 - flnd) * cas / kls;
-    taukls  = flnd * cal / kls;
+    // Intermediate Model Parameters
+    ocean_area = (1.0 - flnd) * earth_area;    // m2
```
bpbond: Still seems better to do `const double ocean_area = ...` but OK

source/components/temp_doeclim_component.cpp
```diff
@@ -197,29 +184,29 @@ void TempDOECLIMComponent::prepareToRun() throw ( h_exception ) {
     ak = 0.31;      // slope in climate feedback - land-sea heat exchange linear relationship
     bk = 1.59;      // offset in climate feedback - land-sea heat exchange linear relationship, W/m2/K
     csw = 0.13;     // specific heat capacity of seawater W*yr/m3/K
-    earth_area = 5100656 * pow(10.0, 8);
+    earth_area = 5100656 * pow(10.0, 8);  // m2
```
bpbond: This surely should be `const`

headers/components/temp_doeclim_component.hpp
bvegawe: These are the DOECLIM parameters that are calculated from the above constants and are in turn used to calculate the model outputs. Perhaps there's a better way to label them?

bpbond: "DOE parameters calculated from constants above" ?

source/components/temp_doeclim_component.cpp
bvegawe: This is how the variables are defined in Kriegler, 2005 (Table 2.1). They are used to calculate time scales that have units of [years], which I presume is why they are defined here with the odd units.

rplzzz: Usually you get units with internal calculations because the quantity involved is used in a formula with other measured quantities, each with its own natural unit, and leaving the units uncanceled makes it easier to apply the formula (and possibly also makes the numerical values of the constant nicer).  For example, the Hubble constant is usually quoted in km/s/Mpc.  You could cancel that to s<sup>-1</sup>, but that obscures the fact that in normal use, you multiply the constant by a distance in Mpc to get a radial velocity in km/s.  

I expect something like that is going on here.  There is a W/m<sup>2</sup>, which looks like a forcing, so these constants are probably meant to be used with a forcing and a time (in years) to get you a temperature change.  This calculation would be a pain if you converted the unit to J/m<sup>2</sup>/K.  (What the heck does a J/m<sup>2</sup> represent?)

PR #211, opened 2018-01-03
dev_doeclim: a bug fix, const parameters, updated references

bvegawe:
Bug fix in total heat flux calculation. Improved DOECLIM references. Made some parameters const and moved their assignment to the header.

### DIFF
--- a/project_files/xcode/hector.xcodeproj/project.pbxproj
+++ b/project_files/xcode/hector.xcodeproj/project.pbxproj
@@ -7,9 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		7DD2A4C51E19A6F20004BBA6 /* temp_doeclim_component.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 7DD2A4C41E19A6F20004BBA6 /* temp_doeclim_component.cpp */; };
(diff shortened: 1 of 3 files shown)


project_files/xcode/hector.xcodeproj/project.pbxproj
```diff
@@ -7,9 +7,6 @@
 	objects = {
```
cahartin: @bpbond and @rplzzz do we want to commit the xcode project file?

rplzzz: You probably should.  Windows users will need it in order to build the code, so it's best if it comes in the package.

bvegawe:
@cahartin Hmm, actually I was previously testing this on my mac using xcode. When I now try to build on linux (using `make hector`), I get the following for all of these header initializations: `error: ISO C++ forbids initialization of member`...trying to figure out a solution where we keep them as `const`.

cahartin:
@bvegawe let me tag @pralitp and @rplzzz to see if they know what the issue is.

cahartin:
@bvegawe did you try make clean? and then what compiler are you using? did that change?

bvegawe:
I guess the error depends on the version of gcc. My local cluster default is 4.4.7. When I switch to 6.1.0, the error goes away. 

Do you think that version sensitivity is a problem?

cahartin:
Yea, we make sure to always load 6.1.0 when running GCAM/Hector.

bvegawe:
Ah ok; I'll do the same from now on.

PR #213, opened 2018-01-05
Bugfix hcarb

cahartin:
merging into v2.0


cahartin:
See for comments and figures on commit. https://github.com/JGCRI/hector/pull/201

PR #217, opened 2018-01-16
Apply setData changes as in PR #173 to accept double data values

swillner:
See also: #173


swillner:
I think, it was due to concurrent development. A way around that would be to e.g. privatize `messagedata::isVal` et al. to encourage usage of `messagedata::getUnitval`. But I did not want to fiddle around in your code too much... ;)

PR #218, opened 2018-01-16
Make component loggers inherit settings from global logger

swillner:
Until component loggers can be setup using configuration ini-files, this makes it much easier to turn off logging, e.g. in Pyhector. Hector logging output should be unchanged.

### DIFF
--- a/headers/core/logger.hpp
+++ b/headers/core/logger.hpp
@@ -99,8 +99,8 @@ public:
     Logger();
     ~Logger();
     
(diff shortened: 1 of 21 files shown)


bpbond:
I agree with @rplzzz that this needs revisiting later, but thanks @swillner - looks good.

bvegawe:
I'm a bit confused about what's changed here. When I run hector from the rc2.0.0 branch (e.g. with hector_rcp85.ini), no component logs seem to be created. Is this intended? How would one re-enable them?

PR #219, opened 2018-01-16
Make h_exception subclass of std::exception

swillner:
This way `h_exception`s can be caught by `catch (const std::exception&)`. We need this for Pyhector and believe it to be good practice in general.

### DIFF
--- a/headers/h_exception.hpp
+++ b/headers/h_exception.hpp
@@ -16,23 +16,38 @@
 
 #include <string>
 #include <iostream>
+#include <exception>
 
 //-----------------------------------------------------------------------
 /*! \brief Exception class.
  *
  *  Sends string (message), line, function and file up the call chain.
  */
-class h_exception /* : public std::exception */ {
-public:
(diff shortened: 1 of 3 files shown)


headers/h_exception.hpp
```diff
@@ -16,34 +16,48 @@
...
 	int linenum;
+
+  public:
+    h_exception() : linenum(0) {}
+    h_exception(const std::string& msg_p, const std::string& func_p,
+                const std::string& file_p, const int linenum_p)
+            : msg(msg_p), func(func_p), file(file_p), linenum(linenum_p) {
+    }
+    virtual ~h_exception() throw() {}
+    inline const std::string get_filename() const {
```
rplzzz: Again, doesn't make sense to const qualify a return value that is neither a pointer nor a reference.  Some quick googling suggests that there is some C++ textbook out there that recommends this, but that just goes to show you shouldn't believe everything you read.

swillner:
Thanks for the quick review!

PR #220, opened 2018-01-16
Make getComponentByName public

swillner:
We use this in Pyhector. `getComponentByName` (which is `const`) is a function generally useful from outside the core.

### DIFF
--- a/headers/core/core.hpp
+++ b/headers/core/core.hpp
@@ -85,10 +85,10 @@ public:
     
     static double undefinedIndex();
     
-private:
-    
     IModelComponent* getComponentByName( const std::string& componentName
                                         ) const throw ( h_exception );
+    
+private:
     
     //------------------------------------------------------------------------------
     //! Current run name.


headers/core/core.hpp
```diff
@@ -85,11 +85,11 @@ class Core : public IVisitable {
     
     static double undefinedIndex();
     
-private:
-    
     IModelComponent* getComponentByName( const std::string& componentName
```
rplzzz: This const method is returning a non-const pointer to a model component.  That's kind of sketchy, inasmuch as it exploits a loophole in the C++ const system.  Conceptually, we're promising not to allow anything inside of the `Core` object to be changed using this method.  However, the const method of `Core` only promises not to change any of the `Core` object's member data.  It makes no guarantees about objects accessed through handles held by the core object.

Anyway, would it break anything to make this return a const pointer?  (It might, but if there is an easy way to patch this up, I'd feel more comfortable if we did so.)

swillner: Hmm, that's right. It's only the pointer itself, which could be made `const` (unlike the pointed-to object, so it's not really a loophole, but more counterintuitive, I think). Since we use the function to call `setData` for the actual component, a `IModelComponent const *` would not work, unfortunately. The checking here probably needs to be done at runtime, I'm afraid.

rplzzz: I thought that might be the case.  Oh, well, at least we tried.

PR #224, opened 2018-01-22
Fix altering forcings in ForcingComponent::getData

rgieseke:
While adding more forcings as output variables in Pyhector we've found an issue with adding SO2/SO2i/SO2d forcings. The ordering of calling these matters in Pyhector, so FSO2 needs to be initialized first.

The error was a "Assertion failed: units mismatch'
(openclimatedata/pyhector#24)

It is not a problem within Hector I think but might potentially affect other API users. 
It's possible working around this on the Python side (to ensure ordering) but @swillner came up with a fix in `source/components/forcing_component.cpp`. 

Setting the forcing to 0 later in the if-clause would also have the benefit that it does not matter whether D_RF_SO2d and D_RF_SO2i are read before D_RF_SO2.

When using `getData` an API-user probably expects that function not to have any side effects (such as setting the value to 0 if non-existent in `forcings`). In the end it might be better to have that function be `const`.

### DIFF
--- a/source/components/forcing_component.cpp
+++ b/source/components/forcing_component.cpp
@@ -362,22 +362,37 @@ unitval ForcingComponent::getData( const std::string& varName,
     
     H_ASSERT( date == Core::undefinedIndex(), "Date not allowed for forcing data" );
     
-    if ( forcings.find( varName ) == forcings.end() && baseyear > currentYear ) {
-        // No forcing exists. This probably means we haven't yet reached baseyear,
-        // so create an entry of 0.0 to return below.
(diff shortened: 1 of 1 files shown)


rgieseke:
@bpbond @rplzzz @cahartin Many thanks for reviewing and merging this and the other PRs! Pyhector is now building directly with Hector's rc2.0.0 branch. This should it make much easier to update and test in the future …

rplzzz:
You bet.  I believe @cahartin has one more to look at, and I'm going to tackle the MCMC one this afternoon.  Thanks for all your contributions!

Issue #208, opened 2017-11-29
Automated calibration of Hector against time series


bpbond:
Any re-calibration of Hector for v2 (after merging of #206 ) should use a script that we bundle in the repo.

cahartin:
Yes I think calibrating to observational temperature would be ideal.  HadCrut data is used a lot. But it would be cool to see how different hector is when calibrated to different temp data sets.

bpbond:
Well, I think this will essentially be solved with your and Kalyn's work.

rplzzz:
Yes, but it would still be good to add an analysis code that sets up the calculations and runs them.

PR #229, opened 2018-02-09
Calibration directory, including constraints folder

bvegawe:
A first version of the adaptive MCMC calibration. Tested on Linux with climate sensitivity, ocean heat diffusivity, and aerosol scaling as free parameters, calibrating against observed surface temperature and ocean heat. Parallel chains should be possible but aren't yet implemented. Addresses #208


bvegawe:
This calibration code will instead be part of a new repo: [bvegawe/hector_probabilistic](https://github.com/bvegawe/hector_probabilistic). I'd still love to hear any feedback, though!

Issue #216, opened 2018-01-09
Update changelog.txt


bpbond:
The `./changelog.txt` file needs to be updated with a succinct but complete list of changes before `rc2.0` is merged into `master`.

cahartin:
address in PR #231

PR #231, opened 2018-03-07
updates to change log for v2.0.0

cahartin:
address #216

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,3 +1,17 @@
+Hector 2.0.0
+============
+* Incorporated 1-D diffusive ocean heat model (DOECLIM)
+* deleted the old temperature component
+* bugfix: double counting halocarbon radiative forcing
+* bugfix: re-enabled CO2 concentration constraint
+* initalized FSO2 first before FSO2i and FSO2d
+* getComponentName now public
+* h_exeptions are now a subclass of std::exception
+* component loggers are now optional
+* double values accepted in setData
+* renamed anthro emissions to ffi emisisons (fossil fuel industrial)
+
+
 Hector 1.1.2
 ============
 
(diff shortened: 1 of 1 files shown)


changelog.txt
```diff
@@ -1,3 +1,17 @@
+Hector 2.0.0
+============
+* Incorporated 1-D diffusive ocean heat model (DOECLIM)
```
bpbond: Be consistent with capitalization: either don't capitalize this line, or capitalize the following ones.

changelog.txt
```diff
@@ -1,3 +1,17 @@
...
+* deleted the old temperature component
+* bugfix: double counting halocarbon radiative forcing
```
bpbond: Ideally provide issue numbers for reference: "...forcing (#201 )"

changelog.txt
```diff
@@ -1,3 +1,17 @@
...
+* bugfix: re-enabled CO2 concentration constraint
+* initalized FSO2 first before FSO2i and FSO2d
+* getComponentName now public
```
bpbond: Why? Was this to better support pyhector? Maybe clarify

rplzzz: I wouldn't bother with stuff like this in the change log, as it's not user-visible.  The change log is for telling users what has changed from their perspective.  What is different in the interface?  How will the results be different?  It isn't necessary or desirable to note implementation changes with no effects that are visible to users.

cahartin: yes. most of those changes after bug fix were to support pyhector

rplzzz: You could collapse all such changes into a single line something like: 

* Various changes to internals to support pyhector

bpbond: Then maybe a single line saying "Variety of internal changes to better support calling Hector from external code like pyhector"

rplzzz: Ha! Beat you by 18 seconds!

PR #233, opened 2018-03-07
Sample outputs v2 0

cahartin:


cahartin:
good point.  i'll keep them with `samples_` so that they don't get confused with a users output

PR #235, opened 2018-03-13
fixed license - intended to be GPLv3

cahartin:


cahartin:
@rpl @bpbond what should we do about the released Hector v2.0?

rplzzz:
I'm not sure there is much we can do in this respect.  Having given people permission to use it under certain terms, we can't really revoke that.  What we could do is draft release 2.0.1 with the new license and delete release 2.0.0 from the releases page, but release 2.0.0 with the old license will always be accessible for to anyone who cares enough to go looking for it.

The good news is that the differences in the license terms probably won't matter for anything that anyone is likely to be doing with Hector.

rplzzz:
Hmm.  Apparently we _can_ change the tag pointed at by the GitHub release.  However, I think the DOI would still point at the original version, which may not be ideal.

rgieseke:
And it's embedded in Pyhector (and its Zenodo archive) ... but I don't think it's a big problem, if I were a lawyer I'd probably argue that since GPL-only licensed code was included in the ECL wasn't valid anyway ;-)

The other way round seems to be no problem, I had started looking on the compatibility of ECL and GPL to check wether it's compatible with Pyhector's AGPL, which seems to be the case (https://en.wikipedia.org/wiki/Educational_Community_License).

Happy to quickly push out Pyhector 2.0.1 as well :-)

rplzzz:
Personally, I think 2.0.1 is the best way to go, but that's really up to @cahartin.

Whatever we decide on the releases, is there any reason not to go ahead and merge this?

cahartin:
2.0.1 sounds good to me.  I'd like the DOI's to stay with the correct version.

PR #237, opened 2018-03-21
Update version number in headers file

### LINKED ISSUE #236: Internal version number
Is the internal version number in 

https://github.com/JGCRI/hector/blob/master/headers/h_util.hpp#L29

on purpose only having 2 digits or should it be updated with 2.0.1 as well?

I've been thinking about having a better way to see from Pyhector with which Hector version it has been compiled. For the Python part of Pyhector we use [Versioneer](https://github.com/warner/python-versioneer) which takes the version string from the git commit or tag.

It seems something similar exists for C++: https://github.com/smessmer/gitversion
(Not sure if it would work with Hector.)

rgieseke:
Closes #236 

(Merging this would clarify that MODEL_VERSION is supposed to follow the semantic versioning scheme.)

### DIFF
--- a/headers/h_util.hpp
+++ b/headers/h_util.hpp
@@ -26,7 +26,7 @@
  * \brief The model version number to be included in logs and outputs.
  * \note This must be updated manually when the model version changes.
  */
-#define MODEL_VERSION "2.0"
+#define MODEL_VERSION "2.0.1"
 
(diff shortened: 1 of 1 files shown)


rplzzz:
I'm not sure why this never got approved, but at this point I've got a branch that is going to become v. 2.1, so it will be easier to deal with it there.

PR #243, opened 2018-08-14
Reset capability

### LINKED ISSUE #240: Add reset capability
Add a capability to reset the core and all components to a previous date, or to the beginning of the simulation.  Right now the only way to do this is to destroy the core and set up a new one (which is what we do in GCAM), but that is undesirable because it triggers an unnecessary reread of all of the input files.  

This capability is needed for problems where you want to run several alternate versions of a hector scenario with different emissions and/or parameters.  Examples include target-finding in GCAM, or Markov chain Monte Carlo.

rplzzz:
Add method `reset` to the public interface of `Core`.  Calling this method causes the entire model to reset its state to a specified (previous) time.  Additionally, resetting to a time prior to the start of the model causes the spinup to be rerun, allowing us to change model parameters and restart the scenario in a state consistent with the new parameters.

Closes #240

### DIFF
--- a/headers/data/tvector.hpp
+++ b/headers/data/tvector.hpp
@@ -29,6 +29,7 @@
 #include <map>
 #include <limits>
 #include <string>
+#include <cmath>
 
 #include "core/logger.hpp"
 #include "h_exception.hpp"
@@ -48,6 +49,16 @@ public:
     const T_data &get(double) const throw( h_exception );
     T_data &get(double) throw(h_exception); // allow modify-in-place
     bool exists( double ) const;
+
+    // indexing operators: the const version is just syntactic sugar
+    // for get().  The non-const version works differently.  If the
+    // requested object doesn't exist, it default constructs it,
+    // assigns the new object, and returns the reference to the newly
+    // created object.  This allows you to write tvec[t] = xx.
+    const T_data &operator[](double t) const throw(h_exception) {
+        return get(t);
+    }
+    T_data &operator[](double t);
     
     double firstdate() const;
     double lastdate() const;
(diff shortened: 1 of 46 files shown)


headers/data/tvector.hpp
```diff
@@ -0,0 +1,206 @@
...
+    void set(double, const T_data &);
+    const T_data &get(double) const throw( h_exception );
+    T_data &get(double) throw(h_exception); // allow modify-in-place
+    bool exists( double ) const;
+
+    // indexing operators: the const version is just syntactic sugar
+    // for get().  The non-const version works differently.  If the
+    // requested object doesn't exist, it default constructs it,
+    // assigns the new object, and returns the reference to the newly
+    // created object.  This allows you to write tvec[t] = xx.
```
bpbond: 👏

headers/data/tvector.hpp
```diff
@@ -0,0 +1,206 @@
...
+int tvector<T_data>::size() const {
+    return int( mapdata.size() );
+}
+
+/*! \brief truncate a time vector
+ *
+ *  \details The default is to wipe all of the data in the time vector
+ *           after the input date.  By setting the optional after
+ *           argument to false, you can wipe data before the input
+ *           date instead.
```
bpbond: Maybe add a note describing why this is necessary, i.e. when `reset` then we truncate so that...

source/components/bc_component.cpp
```diff
@@ -132,6 +132,16 @@ unitval BlackCarbonComponent::getData( const std::string& varName,
     return returnval;
 }
 
+void BlackCarbonComponent::reset(double time) throw(h_exception)
+{
+    // Set time counter to requested date; there are no outputs to
+    // reset.
+    oldDate = time;
+    H_LOG(logger, Logger::NOTICE)
+        << getComponentName() << " resetting to time= " << time << "\n";
```
bpbond: All the other messages say "reset to time" - let's be consistent

headers/data/tvector.hpp
rplzzz: FWIW, I cribbed the truncate method, the `after` argument, and the description you quoted from the `tseries` class.  I wound up using it in the reset methods, but the original purpose was to make `tvector` and `tseries` work the same way as much as possible.

source/components/bc_component.cpp
rplzzz: Ok, done.

PR #251, opened 2018-09-07
BUILD: Change BOOSTROOT/stage/lib to BOOSTROOT/lib

### LINKED ISSUE #248: Compile against pre-built boost libraries?
Based on the Wiki installation documentation, and the Makefile source, it looks like the Hector build process also has to compile Boost libraries from source. Is it possible to compile Hector against a satisfactory version of Boost that already exists on the machine (e.g. one installed via a package manager, such as `apt` for Ubuntu or `brew` on MacOS)?

If not, perhaps this could be accomplished though a variable like `BOOSTLIB` (`BOOSTLIB=/path/to/boost_X.Y.Z make hector`)?

ashiklom:
This should make it easier to compile Hector against system or package manager installations of Boost. I also included instructions for symlinking in case Boost is installed from source, as it was originally.

EDIT: This fixes #248. (Added so merging this closes that issue automatically).


rgieseke:
Not sure if that's helpful, but here is Pyhector's travis config where we use system Boost (and ask people to install via `apt-get` or `brew`):

https://github.com/openclimatedata/pyhector/blob/master/.travis.yml

rplzzz:
Thanks @rgieseke.  That is exactly what I was envisioning us doing here.

PR #255, opened 2018-09-17
R interface for hector

### LINKED ISSUE #246: Define R interface for Hector
Determine what Hector functions should be exposed via the R interface and set up a skeleton for them.

Candidate interface:

- [x] `setup(inifilename)`: Instantiate the model core, read input file, run through spinup
- [x] `run(time)`: Run the model up through `time`.  
- [x]  `DataFrameVisitor`: Visitor to collect and format results.
- [x] `setData(component, what, time, value)`: Set data in `component`.  Should support vectorized `time` and `value`.
- [x] `reset(time)` Reset the model to a previous time, rerunning spinup if necessary.
- [x] `shutdown()` Shut down the model.  
- [x] `getStartDate()` Report the starting date for the model.

Questions to answer:

* ~Do we want to try to support having multiple instantiated cores?  This might be useful for doing comparisons between multiple scenarios, or for having an MC likelihood account for performance over several scenarios. However, it's not clear to me that Hector can support having its components multiply instantiated.  We've never tested it that way.  If we do decide to support multiple cores, how do we keep track of them, both on the R side and on the C++ side?~
This turns out to be pretty easy to support.  You hold the core objects in a vector on the C++ side, and the R objects hold an index into that vector.

rplzzz:
Provide an R interface for Hector.

Fixes #246


rplzzz:
This concludes the development on the hector R package per se.  I'm now going to be turning my attention to the Monte Carlo calculations we want to do.  Eventually we will want to sync up the R package version numbering with the hector version numbering, but until we get a little experience with the R version I'm going to keep it as a 0.x version.

ashiklom:
FWIW, I tested this out and it seems to work really well!

``` r
library(hector)

ini_file <- system.file("input/hector_rcp45.ini", package = "hector")
test_core <- newcore(ini_file)
#> ed Sep 19 13:31:53 2018:WARNING:printLogHeader: hector version 2.1.0
run_out <- run(test_core)
results <- fetchvars(test_core, 2000:2300)

done_out <- shutdown(test_core)

head(results)
#>   year variable    value    units
#> 1 2000       Ca 362.2546 ppmv CO2
#> 2 2001       Ca 363.8910 ppmv CO2
#> 3 2002       Ca 365.5739 ppmv CO2
#> 4 2003       Ca 367.3020 ppmv CO2
#> 5 2004       Ca 369.1146 ppmv CO2
#> 6 2005       Ca 371.0644 ppmv CO2

library(ggplot2)
import::from("magrittr", "%>%")
import::from("dplyr", "filter")

results %>%
  ggplot() +
  aes(x = year, y = value) +
  geom_line() +
  facet_wrap(~variable, scales = "free")
```

bpbond:
That is COOL. Wow this will be a game-changer @rplzzz

rplzzz:
Thanks.  It occurs to me that we should probably add some installation instructions to the splash page.  I'll try to get to that over the next day or so.  (Unless someone else wants to do me a solid and knock that out...)

ashiklom:
@rplzzz I can do that. My time is considerably cheaper than yours 😉 Basically, just `devtools::install_github` and how to set the environment variables for Boost (`Sys.setenv` or `.Renvironment`), right?

bpbond:
@ashiklom This would be handy as I just tried to install the `hector` package and it failed because of Boost...

rplzzz:
@ashiklom That's correct.  If you could do that, I'd be grateful.  I've got some other stuff due this week and won't be able to get back to this for a few days.

ashiklom:
How's this look? https://github.com/JGCRI/hector/wiki/BuildHector

Once this gets merged, we just have to drop the `@rcpp` tag. Tested this on my machine and it seemed to work.

rplzzz:
Are github org names not case sensitive?  I never knew that.

ashiklom:
I'm pretty sure the entire GitHub reference, including the repository name, is case-insensitive. Which is great for working with code from such annoyingly-capitalized projects as FoRTE and PEcAn 😜

src/core.cpp
```diff
@@ -692,4 +682,47 @@ double Core::undefinedIndex() {
     return -1;
 }
 
+std::vector<Core *> Core::core_registry;
+
+/*! Create a core and add it to the registry 
+ */
+int Core::mkcore(bool logtofile, Logger::LogLevel loglvl, bool logtoscrn)
+{
```
bpbond: Not a big deal, I guess, but this syntax (brace on newline) doesn't match anything else in Hector.

src/rcpp_hector.cpp
```diff
@@ -0,0 +1,298 @@
...
+            msg << "While parsing hector input file: " << e;
+            Rcpp::stop(msg.str());
+        }
+
+
+        // The Following three lines of code are occasionally useful for
+        // generating debugging output; however, they leak memory.
+        // Therefore, they should only be used for short tests where
+        // you need the CSV output to compare to a benchmark run.
+        // TODO:  Remove these before release.
```
bpbond: Remove?

src/rcpp_hector.cpp
```diff
@@ -0,0 +1,298 @@
...
+//' This function should be called as \code{mycore <- shutdown(mycore)} so that the change
+//' from active to inactive will be recorded in the caller.
+//'
+//' @param core Handle to a Hector instance
+//' @return The Hector handle, modified to show that it is no longer active
+//' @export
+// [[Rcpp::export]]
+List shutdown(List core)
+{
+    // TODO: check that the list supplied is an hcore object
```
bpbond: Do, open an issue, or remove.

tests/testthat/test_hector.R
```diff
@@ -0,0 +1,88 @@
+## test basic hector functionality
```
bpbond: Nice job

rplzzz:
FYI, the nine new commits that I just pushed are from merging the n2o branch, so everything there has already been looked at in other PRs.

DESCRIPTION
```diff
@@ -0,0 +1,19 @@
+Package: hector
+Title: Run the Hector Simple Climate Model
+Version: 0.1.0
+Authors@R: person("First", "Last", email = "first.last@example.com", role = c("aut", "cre"))
```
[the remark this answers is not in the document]
rplzzz: Derp.  Fixed.

R/messages.R
```diff
@@ -0,0 +1,72 @@
...
+
+default_fetchvars <- c(ATMOSPHERIC_CO2, RF_TOTAL, RF_CO2, GLOBAL_TEMP)
+
+#' Fetch results from a running Hector core
+#'
+#' This function will fetch results and return them as a data frame.  The core
+#' must be active in order for this to work; attempting to fetch results from a
+#' core that has been shut down will cause an error.
+#'
+#' The variables to fetch should be given as a sequence of character strings
```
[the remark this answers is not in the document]
rplzzz: A "sequence" is anything that could be iterated over, either a vector or a list.  Either one will work in this function.

R/messages.R
```diff
@@ -0,0 +1,72 @@
+#### Code for interacting with the hector messaging system
+
+
+default_fetchvars <- c(ATMOSPHERIC_CO2, RF_TOTAL, RF_CO2, GLOBAL_TEMP)
```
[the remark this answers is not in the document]
rplzzz: `format.hcore` is the specialization of the `format` generic to the `hcore` class.  It _has_ to be named that way, per the S language specification (of which R is an implementation).

`default_fetchvars` is an object that is _not_ a specialization of a generic, so it doesn't use a `.` as a word separator so as to avoid giving the impression that it might be.  Base R doesn't use this convention, and you see names like `read.csv` all over the place, but tidyverse conventions specify that the `.` should be reserved for specializing generics, and I agree with them.

`isactive` is kind of a special case.  The convention of using `isfoo` to name a predicate testing the "foo" property goes back at least to the 1990s (*).  Prior to that you probably would have used `foop`, but these days only programmers of advanced age and thinning hairline like me are familiar with that usage.  You see this some in base R (`isS3method`, `isGeneric`, etc.), but it's inconsistent.  The `is` function is a generic for testing class membership, so `is.foo` should be reserved for specializations of that method.  Tidyverse style would suggest `is_foo`.  Personally, I think that's needless extra typing, and simply running the "is" together is more readable to me, but then I've been reading code written this way for a very long time.

I'll change `isactive` to `is_active` if you really feel strongly about it.  The other two are pretty much dictated by the structure of the language.

(*) If, however, the property being tested is a verb, you normally drop the "is", as in `inherits`.

R/messages.R
```diff
@@ -0,0 +1,72 @@
...
+#'
+#' The variables to fetch should be given as a sequence of character strings
+#' corresponding to the capabilities declared by hector components.  These
+#' strings should be generated using the capability identifier functions (see
+#' below for a list of documentation pages for these).
+#'
+#' The list of variables to fetch if you don't specify \code{vars} is stored in
+#' the \code{hector.default.fetchvars} option.  If this option is also unset, then
+#' the default variable list is CO2 concentration, total radiative forcing, CO2
+#' forcing, and global mean temperature.
```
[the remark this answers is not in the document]
rplzzz: Thanks.  The R runtime's option facility is very useful.  More packages should take advantage of it.

R/messages.R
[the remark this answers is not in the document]
bpbond: Ah, right about `format.hcore`, missed that. I prefer `is_active` but ymmv so either.

src/core.cpp
rplzzz: Weird.  Opening brace on its own line for function definitions (and nowhere else) goes all the way back to K&R.  Putting the opening brace on the definition line seems to be a Java thing.  K&R, Stroustrup, Linux foundation, and GNU all put the opening brace of a function definition on its own line.

Examples:
https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces
https://en.wikipedia.org/wiki/Indentation_style#K&R_style
https://en.wikipedia.org/wiki/Indentation_style#GNU_style

Personally, I prefer the Stroustrup formatting (as defined in _The C++ Programming Language_ -- it turns out Stroustrup himself now does something different), and I _really_ hate cuddling the brace on a function definition.  It messes with some text editors' functions for jumping to the beginning of a function, and it makes it harder for me, personally, to visually scan for the beginning of a function.

Again, I'm not saying we should go back and change all the existing code, but we should consider encouraging the Stroustrup formatting style going forward.

src/rcpp_hector.cpp
rplzzz: Yep, not needed anymore.

tests/testthat/test_hector.R
rplzzz: Thanks.

PR #263, opened 2018-10-03
BUILD: Use `BH` package to automatically provide Boost libraries

ashiklom:
This removes the need to manually install Boost by depending instead on a well known, widely-used, R-friendly distribution (via the [BH package](https://github.com/eddelbuettel/bh)).

To test, run the following:

```r
devtools::install_github("ashiklom/hector@bh-boost")
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -7,8 +7,12 @@ Depends: R (>= 3.3)
 License: GPL-3
 Encoding: UTF-8
 LazyData: true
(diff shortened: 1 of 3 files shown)


ashiklom:
Looks like this doesn't actually work unless Boost is already installed in a location known by `ld`, and may not work at all on Windows. So we should hold off on merging this, or abandon it altogether.

Issue #151, opened 2016-08-03
Avoid the need to link to compiled boost libraries


pralitp:
It would be nice if we could avoid having to link to compiled boost libraries.  We currently need to do this for boost.filesystem.

However boost.filesystem is being adopted as part of the C++ 17 standard.  We could switch to that once compiler support is there.  At the moment [GCC v5.3](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dynamic_or_shared.html) and [Visual Studio 2015](https://msdn.microsoft.com/en-us/library/hh874694.aspx) support it as experimental feature.  [Clang](http://libcxx.llvm.org/cxx1z_status.html) (libc++) does not yet support it at all:.

With this being said it may be worthwhile to think about other boost libraries that require compiled libraries in particular (Boost.program_options)[http://www.boost.org/doc/libs/1_61_0/doc/html/program_options.html] looks like it could be useful as a more robust and flexible alternative to the INI parser we currently use.

bpbond:
Sounds good. Right, for `Boost.program_options` we'd want to weigh the inconvenience of compiling (bad) versus presumably better Boost code that we're not responsible for maintaining (good).

ashiklom:
This is a good idea, but I don't think it's widely supported yet (e.g. [this](https://stackoverflow.com/questions/42633477/macos-clang-c17-filesystem-header-not-found) makes it seem like it's not available on the OSX version of Clang).

PR #267, opened 2018-10-05
Remove boost::filesystem dependency by using R file functions.

ashiklom:
The only compiled boost library we depended on was `boost::filesystem` (which in turn depends on `boost::system`, also compiled). The only reason we needed it was to check if the CSV file path was absolute, and if not, prepend the directory of the INI file.

By eliminating the boost compiled library dependency (and replacing it with the boost header-only R package `BH`), this means that this should install without issue or any external dependencies on any machine, including Windows. I tested this on a Windows VM and, indeed, it installed just fine.

To test, use the following installation command:

```r
devtools::install_github("ashiklom/hector@new-bh-boost")
```

Here, I replaced calls to `boost::filesystem::path` functions with calls to R's own path manipulation functions (`normalizePath`, `dirname`, and `file.path`) via Rcpp. I also subtly changed the logic to be as follows: If the CSV file path normalizes to an existing file, point to that file. Otherwise, point to the CSV file in the same directory as the INI file. I believe this means that, if you have a CSV file in the same directory as you're running hector (i.e. your R working directory), you can pass that in as a relative path ( (i.e. just the file name) to the INI file and it should work. I'll need to test this out, though.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -7,9 +7,8 @@ Depends: R (>= 3.3)
 License: GPL-3
 Encoding: UTF-8
 LazyData: true
-LinkingTo: Rcpp
-Imports: Rcpp (>= 0.12)
-VignetteBuilder: knitr
+LinkingTo: Rcpp, BH
+Imports: Rcpp (>= 0.12), BH (>= 1.65)
 RoxygenNote: 6.0.1
 Suggests:
     testthat,
--- a/src/Makevars
+++ b/src/Makevars
@@ -1,11 +1,2 @@
-## The following environment variables must be set in .Renvironment:
-##   * BOOSTROOT  : Top level of Boost library directories
-##   * BOOSTLIB   : Location of compiled boost libraries
-
-
-PKG_CPPFLAGS = -I../inst/include -I${BOOSTROOT} -DUSE_RCPP 
-
-PKG_LIBS = -L${BOOSTLIB} -Wl,-rpath ${BOOSTLIB} -lboost_system -lboost_filesystem
-
-
-
+PKG_CPPFLAGS = -I../inst/include -DUSE_RCPP 
+PKG_LIBS = -Wl,-rpath 
--- a/src/Makevars
+++ b/src/Makevars
@@ -1,2 +1,1 @@
 PKG_CPPFLAGS = -I../inst/include -DUSE_RCPP 
-PKG_LIBS = -Wl,-rpath 
--- a/src/ini_to_core_reader.cpp
+++ b/src/ini_to_core_reader.cpp
@@ -15,7 +15,7 @@
 #include <boost/lexical_cast.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/trim.hpp>
-#include <regex>
+#include <Rcpp.h>
 
 #include "core.hpp"
 #include "message_data.hpp"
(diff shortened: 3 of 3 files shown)


bpbond:
As @pralitp asked on Slack--aren't we just trading one dependency for another? Are we gaining anything? I mean, I guess R is easier for a novice to install than Boost, but...

bpbond:
For example, does this mean that the GCAM-Hector installation will require R as well?

bpbond:
I admit the benefit here is not clear to me.

ashiklom:
Yes, this makes R and Rcpp a dependency, so yes, in that sense, it trades off one dependency for another.

@bpbond Yes, the main benefit is that, IMHO, R/Rcpp is easier to install than Boost, especially on Windows systems. I was inspired to do this after chatting with @kdorheim about her experiences with Hector and seeing firsthand how challenging Hector installation currently was on Windows machines, even working together with me and @rplzzz . So I think the ease of installation benefit should not be understated.

Isn't the GCAM data system closely tied to R, including a bunch of `tidyverse` stuff (which, IIRC, depends on Rcpp)? How big is the GCAM user base that does not also have access to R? If it's large (or we're expecting it to be fairly large), then I agree that this is a bad fix and should be rejected. I honestly still don't know much about Hector's user base, including how it works with GCAM, so I don't know all the use cases we're trying to support.

Incidentally, I also tried to implement a home-grown solution using standard C++ in an earlier commit ([link](https://github.com/JGCRI/hector/pull/267/commits/040169f6b01a1c0e16c6b323d5a990bfc626b613#diff-e70fd5bedd4481229b76d1255d96ba6dR113)), but, as @rplzzz pointed out, it's probably incomplete, especially for Windows.

Anyway, if you all think this is a bad idea, feel free to close this. Like I said, I'm still wrapping my head around how Hector is being used, and I definitely don't want anything I do to fly in the face of essential use cases.

pralitp:
> How big is the GCAM user base that does not also have access to R?

@ashiklom , it's hard to say but typically in the past the GCAM community has not run R / the gcam data system at all.  Since GCAM 5.1 / the inclusion of gcamdata we will try to get more of the community generating GCAM inputs through the data system.  Our first community modeling meeting with GCAM 5.1 is coming up so we'll see how that goes.

> Incidentally, I also tried to implement a home-grown solution using standard C++

Yea there certainly a lot of pitfalls you would have to look out for.

Have you looked into #151 at all?  Do you know which version of GCC is used on windows?  They have had support for `std::filesystem` (basically exactly the same as `boost::filesystem`) for a while albeit in the experimental phase:
https://stackoverflow.com/questions/33149878/experimentalfilesystem-linker-error

ashiklom:
> it's hard to say but typically in the past the GCAM community has not run R / the gcam data system at all

Got it. I do agree with the larger point that _forcing_ people to have R for this is not worth it. I was tunnel-vision-ing a bit on Hector as an R package and was missing the bigger picture of Hector applications.

> Have you looked into #151 at all?

No, but now that I have, that does seem like a great way to go about this. If this works, I agree it's a better way to go. I think it may require explicitly specifying the C++11 standard when compiling in R, but that shouldn't be too hard to do.

> Do you know which version of GCC is used on windows?

Windows's R uses [MinGW-w64](https://mingw-w64.org/doku.php), which looks like it was forked from GCC about 10 years ago. It seems to have at least some C++11 support, so that could work. Beyond that, there seem to be a bunch of different Unix emulators for Windows floating around (e.g. Cygwin, git-bash). I have no idea what they ship with. Windows 10 also has the Linux subsystem, which is Ubuntu-based.

ashiklom:
I'll close this PR and open a fresh one once I have a satisfactory solution.

PR #271, opened 2018-10-21
Add aerosol scale factor as an adjustable model parameter.

rplzzz:


cahartin:
Aerosol scaling accounts for the uncertainty in the aerosol radiative efficiency factors. It is useful in calibrations of DOECLIM (Urban and Keller, 2010).  We can change 1 parameter - the scaling parameter instead of changing all of the radiative efficiency factors for each aerosol.

rplzzz:
@bpbond Whoops, didn't see that you had commented.  In addition to what @cahartin said, it came up in one of the talks last week that the UIUC team had included it in their historical calibrations, so it seems like a good candidate for us to investigate as well.   (Not to mention supporting their calculations in the future, should they decide to do it through the R interface).

cahartin:
example of the calculation of how alpha plays a role in Hector. 

RF = RF_total - (1-alpha) * (RF_BC + RF_OC + RF_SO2_indirect + RF_SO2_direct)

PR #275, opened 2018-11-13
When an input is registered, also register it as a capability.

rplzzz:
Now `registerInput` calls `registerCapability` directly.  Because multiple components can share an input (this only happens between O3 and OH, as far as I can tell), but only one component can register a particular capability, we also provide a way for `registerInput` to suppress the warning message that would otherwise be generated.

### DIFF
--- a/src/core.cpp
+++ b/src/core.cpp
@@ -543,28 +543,47 @@ IModelComponent* Core::getComponentByCapability( const string& capabilityName ) 
 /*! \brief Register a capability as associated with a component.
  *  \param capabilityName The capability of the component to register.
  *  \param componentName The name of the associated component.
+ *  \param warndupe If true (the default) warn when a duplicate capability
+ *                  is declared.  This should only be set to false when being
+ *                  called from registerInput, which sometimes creates duplicates
+ *                  for benign reasons.
  *  \exception h_exception If the componentName was not recognized.
  */
-void Core::registerCapability( const string& capabilityName, const string& componentName
+void Core::registerCapability(const string& capabilityName, const string& componentName, bool warndupe
                               )  throw ( h_exception ){
     H_ASSERT( !isInited, "registerCapability not available after core is initialized")
-    
-    if( componentCapabilities.count( capabilityName ) ) {
+
+    // check whether the capability already exists
+    int ncap = componentCapabilities.count(capabilityName);
+    if(ncap > 0  && warndupe) {
+        // If this capability is a dupe, issue a warning, unless we
+        // have been instructed not to.
         H_LOG( glog, Logger::WARNING ) << componentName << " is declaring capability " << capabilityName << " previously registered" << endl;
     }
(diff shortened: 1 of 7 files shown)


bpbond:
Now that I see this in black and white...

One problem with this is that it's muddying the original definition of "capability": instead of "output produced by this component", it's become "any input or output of the component". This imprecision already existed in a limited way (i.e. those `// Allow parameter values to be queried` lines in a few components) but now it seems more serious, as the component inputs aren't by any stretch part of its "capabilities". 

Do we want to rethink the term? Or define separate "output" and "readable_data" tags? Or am I making too much of this?

rplzzz:
@bpbond I guess the way I would square that circle is to say that we are now requiring that all inputs also be capabilities.

bpbond:
That's the English translation of the code, agreed. My concern is semantic though: it doesn't match any common-sense meaning of "capability".

rplzzz:
Ah, I didn't scroll down far enough to see the rest of your concern.  A "capability", at least the way I've been defining it, is any piece of information that a component wishes to expose for use to the rest of the framework.  It doesn't _have_ to be output per se.  It's just data owned by the component that it is prepared to allow other components to inspect.

I have to admit, this has just been my personal nomenclature; I don't know that I ever talked about it to anyone else, but I've also been using it that way in other modeling frameworks I've been developing.  I actually talked about this a bit in my talk at the meetings last week, and it got a fairly positive response.

bpbond:
Okay. I think some clear comments will help things then.

src/core.cpp
```diff
@@ -543,28 +543,47 @@ IModelComponent* Core::getComponentByCapability( const string& capabilityName )
 /*! \brief Register a capability as associated with a component.
  *  \param capabilityName The capability of the component to register.
  *  \param componentName The name of the associated component.
+ *  \param warndupe If true (the default) warn when a duplicate capability
+ *                  is declared.  This should only be set to false when being
+ *                  called from registerInput, which sometimes creates duplicates
+ *                  for benign reasons.
  *  \exception h_exception If the componentName was not recognized.
```
bpbond:
```suggestion
 *  \exception h_exception If the componentName was not recognized.
 * \note By "capability" we mean any piece of data that the component wants to expose to the model core or other components; this can be an input, an output, or an internal variable.
```

src/core.cpp
```diff
@@ -543,28 +543,47 @@ IModelComponent* Core::getComponentByCapability( const string& capabilityName )
...
+        // Adding a duplicate capability has no useful effect anyhow.
+        componentCapabilities.insert( pair<string, string>( capabilityName, componentName ) );
+    }
 }
 
 //------------------------------------------------------------------------------
 /*! \brief Register a component as accepting a certain input
+ *
+ *  \details Associate the input capability with a component.  We also
+ *           implicitly register the capability (under the same name)
```
bpbond:
```suggestion
 *           register the input as a capability (under the same name), allowing other objects to query the host component for its value(s).
```

PR #276, opened 2018-11-13
Fix potential memory leaks when hector cores are garbage collected.

rplzzz:
This patch ensures that whenever a hector core is garbage collected in R, the underlying C++ data structures will be destroyed and their memory reclaimed.  I also bundled in a small documentation update that K requested.


rplzzz:
Because the word "collected" in the title goes on the second line, at first glance the title of the PR is "... when Hector cores are garbage"  That may just be the best thing I've seen all day.

rplzzz:
Thanks.

PR #282, opened 2019-01-10
If building R package, use R functions. Otherwise, use boost::filesystem.

ashiklom:
(This is a revisit of #267 -- see that discussion for additional context).

This commit removes the dependency on compiled `boost` libraries when installing Hector as an R package. It does this by using `Rcpp` bindings to base R functions to process file paths. This -- combined with the use of the `BH` package to provide Boost headers -- means that the R package version can be installed on _any_ system (including Windows -- looking at you, @cahartin and @kdorheim) with just `devtools::install_github` _without installing any other system dependencies_.

You can test this out by running the following in an R session:

```r
devtools::install_github("ashiklom/hector@windows")
```

I tested this on a clean Windows VM and on AppVeyor, and it seemed to work.

Per the discussion on #267, when Hector is _not_ installed as an R package (technically, when the `USE_RCPP` macro is not defined), it falls back on `boost::filesystem`. Although I haven't tried, it _may_ be possible to build a standalone Hector executable that also uses R instead of Boost, if you add the `-DUSE_RCPP` flag and include the path to the Rcpp library.

Also, note that I slightly modified the algorithm for finding the CSV file relative to the INI file as follows:

- First, check if the path given in the INI file points to an existing file.
- If yes, use the file that it found.
- If no, assume the path is relative to the INI file directory

I did this mainly because base R has no `isAbsolutePath` function (or similar). There is [`R.utils::isAbsolutePath`](https://www.rdocumentation.org/packages/R.utils/versions/2.7.0/topics/isAbsolutePath), which wouldn't be too hard to use, but it's implemented with just a few regexes (which doesn't seem very robust) and it would force us to add another R package dependency.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/JGCRI/hector.svg?branch=master)](https://travis-ci.org/JGCRI/hector)
+[![Build Status](https://travis-ci.org/JGCRI/hector.svg?branch=master)](https://travis-ci.org/JGCRI/hector) [![AppVeyor Build status](https://ci.appveyor.com/api/projects/status/4ld4ay2gyg2s8w6w?svg=true)](https://ci.appveyor.com/project/ashiklom/hector)
 
 Hector
 ======
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-[![Build Status](https://travis-ci.org/JGCRI/hector.svg?branch=master)](https://travis-ci.org/JGCRI/hector) [![AppVeyor Build status](https://ci.appveyor.com/api/projects/status/4ld4ay2gyg2s8w6w?svg=true)](https://ci.appveyor.com/project/ashiklom/hector)
+[![Build Status](https://travis-ci.org/JGCRI/hector.svg?branch=master)](https://travis-ci.org/JGCRI/hector)
+[![Build status](https://ci.appveyor.com/api/projects/status/2jmhvq7n2ap75u1o/branch/master?svg=true)](https://ci.appveyor.com/project/rplzzz/hector/branch/master)
 
 Hector
 ======
--- a/src/ini_to_core_reader.cpp
+++ b/src/ini_to_core_reader.cpp
@@ -15,7 +15,7 @@
 #include <boost/lexical_cast.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/trim.hpp>
-#include <regex>
+#include <Rcpp.h>
 
 #include "core.hpp"
 #include "message_data.hpp"
@@ -80,6 +80,12 @@ void INIToCoreReader::parse( const string& filename ) throw ( h_exception ) {
 int INIToCoreReader::valueHandler( void* user, const char* section, const char* name,
                                   const char* value )
 {
+    // Load R functions for path management
+    Rcpp::Environment base("package:base");
+    Rcpp::Function normalizePath = base["normalizePath"];
+    Rcpp::Function dirname = base["dirname"];
+    Rcpp::Function filepath = base["file.path"];
+
     static const string csvFilePrefix = "csv:";
     INIToCoreReader* reader = (INIToCoreReader*)user;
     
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/JGCRI/hector.svg?branch=master)](https://travis-ci.org/JGCRI/hector)
+[![Build Status](https://travis-ci.org/JGCRI/hector.svg?branch=master)](https://travis-ci.org/JGCRI/hector) [![AppVeyor Build status](https://ci.appveyor.com/api/projects/status/4ld4ay2gyg2s8w6w?svg=true)](https://ci.appveyor.com/project/ashiklom/hector)
```
ashiklom: NOTE: This badge needs to be updated to point to the JGCRI AppVeyor account. I'm not sure if I have all the right permissions to access it.

rplzzz: I'll look into the appveyor config tomorrow.  Thanks for looking into this.

src/ini_to_core_reader.cpp
```diff
@@ -82,6 +92,14 @@ void INIToCoreReader::parse( const string& filename ) throw ( h_exception ) {
 int INIToCoreReader::valueHandler( void* user, const char* section, const char* name,
                                   const char* value )
 {
+    #ifdef USE_RCPP
+    // Load R functions for path management
+    Rcpp::Environment base("package:base");
+    Rcpp::Function normalizePath = base["normalizePath"];
+    Rcpp::Function dirname = base["dirname"];
+    Rcpp::Function filepath = base["file.path"];
+    #endif
```
ashiklom: If we did want to use `R.utils::isAbsolutePath`, we would have to add the following here:

```cpp
Rcpp::Environment rutils("package::R.utils");
Rcpp::Function isAbsolutePath = rutils["isAbsolutePath"];
```

Then, further down, I _think_ we could just do something like this:

```cpp
if (isAbsolutePath(csvFilePath)) {
   ...
}
```

src/ini_to_core_reader.cpp
```diff
@@ -110,14 +128,37 @@ int INIToCoreReader::valueHandler( void* user, const char* section, const char*
...
+	    // as an R package.
+            //
+	    // If the csvFileName normalizes to a real path, use that.
+            // Otherwise, assume that it is pointing to a file in the
+            // same directory as the INI file.
+            try {
+                // Second argument is "winslash". "\\" (single forward
+                // slash) is the default. Need it here to access the
+                // third argument -- mustWork -- which throws the
+                // error to be caught if the path doesn't exist
```
bpbond: 👍

src/ini_to_core_reader.cpp
```diff
@@ -15,15 +15,25 @@
...
+#endif
 
 #include "core.hpp"
 #include "message_data.hpp"
 #include "ini_to_core_reader.hpp"
 #include "ini.h"
 #include "csv_table_reader.hpp"
 
+#ifndef USE_RCPP
 namespace fs = boost::filesystem;
```
pralitp: Maybe just move this up with the `#include <boost/filesystem.hpp>` to keep them together.  Then maybe in the future if we want add support to use the filesystem that is part of the standard library and fall back to boost as a last resort we could just have something like:
```
#if SOME_SORT_OF_STD_FILESYSTEM_CHECK
#include <filesystem>
namespace fs = std::filesystem;
# else
#include <boost/filesystem.hpp>
namespace fs = boost::filesystem;
#endif
```

src/ini_to_core_reader.cpp
pralitp: Given your comments about this, I'd say let's just not worry about it for now.  Your current solution is likely sufficient and if I'm wrong we can deal with it then.

src/ini_to_core_reader.cpp
ashiklom: Good idea. Addressed in e38498333c1162edcdf1e8c30ab296a260e11114.

src/ini_to_core_reader.cpp
```diff
@@ -110,14 +125,37 @@ int INIToCoreReader::valueHandler( void* user, const char* section, const char*
...
-            // relative to the INI file
+            #ifdef USE_RCPP
+            // ANS: This is the algorithm used if Hector is compiled
+	    // as an R package.
+            //
+	    // If the csvFileName normalizes to a real path, use that.
+            // Otherwise, assume that it is pointing to a file in the
+            // same directory as the INI file.
+            try {
+                // Second argument is "winslash". "\\" (single forward
```
rplzzz:
```suggestion
                // Second argument is "winslash". "\\" (single backward
```
Is it worth noting here that all this argument does is to specify what the path separator is on Windows, and that it doesn't do anything on non-Windows platforms?  Reading this, I was a little concerned that we were specifying backslash as the default path separator, and had to go look up the docs for `normalizePath` to see what was up.  Then again, this code will likely never get touched again, so maybe it's not worth fussing with.

src/ini_to_core_reader.cpp
```diff
@@ -15,16 +15,23 @@
 #include <boost/lexical_cast.hpp>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/algorithm/string/trim.hpp>
+
+// ANS: If using the R package, use Rcpp to call R's file processing
```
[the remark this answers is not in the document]
ashiklom: Yes, "Alexey Nikolaevich Shiklomanov", and happy to remove these.

That said, in my experience, I don't find `git blame` (and `git log <filename>`) to be 100% reliable. For one, when files get moved without `git mv` (a common mistake) and then heavily edited -- a frequent combination for a major refactor -- the git history for individual files can get lost. On the other hand, you can get false positives when comments get reformatted -- for instance, @bpbond is more-or-less the default `git blame` target of a bunch of the PEcAn code base because he ran an automatic code formatter on all of our code, even almost none of those changes were functional. Anyway, just my 2 cents...

src/ini_to_core_reader.cpp
ashiklom: Addressed in 09ade41.

rplzzz:
Looks like the appveyor build ran.  It got the usual note about file sizes, and one about the extraneous appveyor.yml file, which is easily fixed by adding that to .Rbuildignore

It also got a warning (which apparently doesn't cause a failure on appveyor -- I believe it does on travis) regarding initialization order in one of our classes.  I _so_ hate that warning; I've _never_ seen it be anything but a false alarm, and I've been writing C++ since the mid '90s.

Anyhow, @ashiklom could you add `appveyor.yml` to the `.Rbuildignore`?  I'd do it myself, but I doubt I have write access to your fork.  I'll open an issue about that warning.  Someone can fix it when they're stuck on a boring telecon or something.

rplzzz:
@JGCRI/hector is this PR ready to merge?

bpbond:
It would be good to have a windows end user try it if possible? @kdorheim ?

kdorheim:
@bpbond @ashiklom 

I just do? 

`devtools::install_github("ashiklom/hector@windows")`

ashiklom:
Was there anything else you wanted me to do on this, or is it ready to be merged?

rplzzz:
I say merge it.

PR #284, opened 2019-01-18
INI read/write functions

ashiklom:
This defines two new exported functions (and a few internal helpers) for working with the INI file:

- `read_ini` reads an INI file into a nested list (like `list(core = list(run_name = "rcp45", startDate = 1745, ...), onelineocean = list(enabled = 1, spinup_chem = 0), ...)`. By default, it also tries to be clever in two ways:
  - Relative paths are converted to full, normalized paths following the same logic as the C code (if the path exists relative to the working directory, use that. Otherwise, assume it's relative to the INI file location). This is toggled with the `normalize_paths` boolean.
  - Time series variables of the form `variable[year] = value` are converted to `data.frame`s with `year` and `variable` as columns. For instance, this part of the RCP45 INI file:

    ```
    Ftalbedo[1750]=0.0
    Ftalbedo[1950]=-0.2
    ```

    ...is interpreted as follows:

    ``` r
    rcp45 <- hector::read_ini(
      system.file("input", "hector_rcp45.ini", package = "hector")
    )
    rcp45$simpleNbox$Ftalbedo
    #>   date Ftalbedo
    #> 1 1750      0.0
    #> 2 1950     -0.2
    ```
    
    This behavior is controlled by the `parse_timeseries` boolean.
- `write_ini` writes an INI list to a file, automatically converting time series `data.frame`s like the one above back to entries in the `variable[year] = value` form.

Because the INI file is read as a list, it can be conveniently modified via base R's `modifyList`. For example:

```r
rcp45 <- hector::read_ini(
  system.file("input", "hector_rcp45.ini", package = "hector")
)
custom_run <- modifyList(rcp45, list(
  core = list(run_name = "custom_run", endDate = 2100),
  simpleNbox = list(beta = 0.40)
))
```

I also modified `newcore` such that if `inifile` is a list, it will write it to a temporary file with `write_ini` before running. I also added a `...` argument to `runscenario` that forwards arguments to `newcore`. Collectively, this means you can now run Hector like this:

``` r
rcp45 <- hector::read_ini(
  system.file("input", "hector_rcp45.ini", package = "hector")
)
custom_run <- modifyList(rcp45, list(
  core = list(run_name = "custom_run", endDate = 2100),
  simpleNbox = list(beta = 0.40)
))
result <- hector::runscenario(custom_run, name = "custom_run")
head(result)
#>     scenario year variable    value    units
#> 1 custom_run 1745       Ca 287.7619 ppmv CO2
#> 2 custom_run 1746       Ca 287.7624 ppmv CO2
#> 3 custom_run 1747       Ca 287.3882 ppmv CO2
#> 4 custom_run 1748       Ca 287.1020 ppmv CO2
#> 5 custom_run 1749       Ca 286.8423 ppmv CO2
#> 6 custom_run 1750       Ca 286.5965 ppmv CO2
```

Another unrelated change, is that I compiled the documentation using an updated version of Roxygen (6.1.1), which added a note to the DESCRIPTION and added a `\concept{...}` section to each Rd file. I also enabled Roxygen markdown documentation in the DESCRIPTION, which makes it easier to write Roxygen tags.


bpbond:
Hey @kdorheim do you mind taking a look at this too? ...given your current work. Thanks.

R/hector.R
```diff
@@ -86,6 +91,11 @@ runscenario <- function(infile)
 newcore <- function(inifile, loglevel=0, suppresslogging=FALSE,
                     name="unnamed hector core")
 {
+    if (is.list(inifile)) {
+        tmp <- tempfile()
+        write_ini(inifile, tmp)
```
rplzzz: We need to ensure that this temporary file is cleaned up once the core has been created (and that it is cleaned up if the function exits abnormally for some reason.)

R/read_ini.R
```diff
@@ -0,0 +1,191 @@
...
+#' @examples
+#' rcp45_ini <- read_ini(
+#'   system.file("input", "hector_rcp45.ini", package = "hector")
+#' )
+#' rcp45_ini$core$run_name
+#' rcp45_ini$onelineocean$enabled
+#' # This one will be a `data.frame`
+#' rcp45_ini$simpleNbox$Ftalbedo
+#' @export
+read_ini <- function(file,
```
rplzzz: Why did we roll our own INI parser instead of using something like the `ini` or `ConfigParser` packages?  I get that we want to be able to interpolate in the time series data, but it seems like we could use the functionality of one of those packages to do the initial read-in and then make a pass through the structure to find any time series that needed to be read.  Is there some other functionality we're incorporating that we can't easily get from one of those packages?

man/shutdown.Rd
```diff
@@ -27,3 +27,4 @@ Other main user interface functions: \code{\link{fetchvars}},
   \code{\link{newcore}}, \code{\link{reset}},
   \code{\link{run}}, \code{\link{setvar}}
 }
+\concept{main user interface functions}
```
rplzzz: I don't love that the `@family` tag adds these as search keywords.  I picked the family tags to make the text in the `\seealso` make sense (in this case, "See also: Other main user interface functions...".  I wouldn't necessarily want these functions to show up when a user searches for any of these keywords individually.  Is there a way to turn off this behavior, or are we stuck with it?

R/read_ini.R
ashiklom: No specific reason, other than to avoid an external package dependency (or multiple recursive dependencies, in the case of `ConfigParser`) and maintain control of the parser in case the INI format or other aspects of the API change.

Neither of those packages seem like they're widely used (`ini` has one GitHub star, `ConfigParser` has none), and the task of parsing an INI file didn't seem particularly complicated, so I didn't have a good reason to trust others' implementation over our own.

R/hector.R
```diff
@@ -55,12 +55,14 @@ NULL
...
 #' in \code{\link{fetchvars}}.
 #'
-#' @param infile INI-format file containing the scenario definition
+#' @param infile INI-format file (as String or list) containing the
+#'     scenario definition. See \code{\link{newcore}}.
+#' @param ... Additional arguments to \code{\link{newcore}}
 #' @return Data frame containing Hector output for default variables
 #' @export
-runscenario <- function(infile)
+runscenario <- function(infile, ...)
```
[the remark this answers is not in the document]
ashiklom: I don't disagree, but I think passing the `...` to `newcore` adds minimal technical debt while improving flexibility in, in my opinion, a fairly intuitive way. For instance, the following seems like a very common use case that we should be able to support:

```r
runscenario(inifile, name = "my_custom_run", suppresslogging = TRUE)
```

ashiklom:
> it doesn't seem like a great way of doing runs. Generally speaking, I'd prefer to use ini files to represent base configurations and where possible to change things in the running core using the existing SETDATA interface.

I definitely agree that setting everything by modifying an existing Hector core is more efficient and conceptually more elegant. But...

>  If there are things that can't easily be changed in a running core

...as long as this is true, I think there is a place for modifying existing INI files programmatically. 

> it's not to hard to create base configs with those characteristics, and users who use alternate configs regularly will probably want to save them anyhow.

As an argument against this functionality, this is where I disagree. I personally like to be able to do 100% of my analysis from within an R session, without having to manually modify any files anywhere. Moreover, modifying INI files via script highlights the changes while hiding the constants. For example, it's tricky to immediately spot the difference between these two:

```
# file1.ini #
; Config file for hector model: RCP26
;------------------------------------------------------------------------
[core]
run_name=rcp26
startDate=1745
endDate=2300
do_spinup=1			; if 1, spin up model before running (default=1)
max_spinup=2000		; maximum steps allowed for spinup (default=2000)

;------------------------------------------------------------------------
[onelineocean]
enabled=0			; 'enabled=0' will disable any component
;output=0			; 'output=0' will disable any component's output
ocean_c=38000		; Pg C

# file2.ini
; Config file for hector model: RCP26
;------------------------------------------------------------------------
[core]
run_name=rcp26
startDate=1745
endDate=2300
do_spinup=1			; if 1, spin up model before running (default=1)
max_spinup=2000		; maximum steps allowed for spinup (default=2000)

;------------------------------------------------------------------------
[onelineocean]
enabled=0			; 'enabled=0' will disable any component
output=0			; 'output=0' will disable any component's output
ocean_c=38000		; Pg C
```

...whereas, it becomes immediately obvious (and much more concise) when looking at this code:

```
case1 <- read_ini("file1.ini")
case2 <- modifyList(case1, list(onelineocean = list(output = 0)))
```

In fact, having this functionality would allow us to replace the 4 base RCP scenarios, which are much more similar than they are different, with a single INI file in which we only modify the different emissions trajectories. That means we would only have to maintain one baseline INI file instead of 4. Similarly, storing INI changes as code rather than as separate INI files should make analyses more resilient to future API changes. For instance, the `case2` line above should, in theory, only break if we remove or change the `output` INI argument.

All that said, I have no problem closing this PR and keeping this code in a separate `hectorutils` repo to keep the Hector core code more focused.

man/shutdown.Rd
ashiklom: This was Hadley's doing -- https://github.com/klutometis/roxygen/issues/611. Doesn't seem to be configurable in any way. I could go through and remove each one manually, but then everyone working with the latest Roxygen would have to do it every time as well. 😞

R/hector.R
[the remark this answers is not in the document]
rplzzz: I suppose.  I think the question here is, do we really want to be encouraging people to use the training-wheels version of the interface for anything more than trying out the model?  Again, not saying we _can't_ do this, just saying that we should consider whether we want to support 1-1/2 partially overlapping APIs.

rplzzz:
> As an argument against this functionality, this is where I disagree. I personally like to be able to do 100% of my analysis from within an R session, without having to manually modify any files anywhere.

Well, in that case you're manually modifying R scripts, n'est-ce pas?  In any configurable system, somewhere there is always a configuration file that says, in essence, "This is the calculation I want to do," and that file is perforce manually produced because it serves as the interface between the developer's brain and the automated system.

> Moreover, modifying INI files via script highlights the changes while hiding the constants. For example, it's tricky to immediately spot the difference between these two:
> 

If they're both disk files, then it's easy to spot with `diff`.  More importantly, if the two files have informative names, you might not need to do a detailed comparison at all.  

> In fact, having this functionality would allow us to replace the 4 base RCP scenarios, which are much more similar than they are different, with a single INI file in which we only modify the different emissions trajectories. That means we would only have to maintain one baseline INI file instead of 4. Similarly, storing INI changes as code rather than as separate INI files should make analyses more resilient to future API changes. For instance, the `case2` line above should, in theory, only break if we remove or change the `output` INI argument.

IMO this impedes, rather than enhances, clarity.  Doing it your way, every Hector application has to start with a bunch of boilerplate code that loads a bunch of emissions and modifies the default configuration with them.  Compare that with the currently abundantly clear:
```R
hcore85 <- newcore('hector_rcp85.ini')
hcore26 <- newcore('hector_rcp26.ini')
```
One could even imagine a scenario in which changing something about the standard scenarios would require us to go change every R script that constructed those scenarios, rather than just changing 4 configuration files.

> 
> All that said, I have no problem closing this PR and keeping this code in a separate `hectorutils` repo to keep the Hector core code more focused.

As long as there are no plans to make this mode of operation mandatory (such as by removing the set of base scenarios we currently provide, such that people have to edit a single base configuration), I'm ok with it going in.  It's not something I would have specifically wanted, nor would I have marked it as a high priority relative to other things we might be doing, but the code looks solid, so if it has value for you, we should include it.

R/hector.R
```diff
@@ -93,6 +93,7 @@ newcore <- function(inifile, loglevel=0, suppresslogging=FALSE,
 {
     if (is.list(inifile)) {
         tmp <- tempfile()
+        on.exit(file.remove(tmp), add = TRUE)
```
rplzzz: :shipit:

man/shutdown.Rd
rplzzz: Yeah, let's not do that.

R/read_ini.R
rplzzz: Not sure that everyone who installs and uses a package from CRAN goes to github and stars the repository.  I certainly don't.  Speaking as someone who once devised a bunch of regexps to parse a bunch of XML inputs, I've come around to using other people's parsers (though not, perhaps, if they use R6 to pretend like they're Python packages, as ConfigParser appears to do).

tests/testthat/test_ini.R
```diff
@@ -0,0 +1,80 @@
...
+  expect_equal(raw_out[14], "tt = 72000000")
+  expect_equal(raw_out[15], "tu = 49000000")
+  expect_equal(raw_out[17], "tid = 200000000")
+
+  rcp45_test <- read_ini(rcp45_testfile)
+  expect_identical(rcp45_ini, rcp45_test)
+})
+
+
+test_that("Passing INI as file or list produces same results.", {
```
bpbond: Good test.

ashiklom:
Since that's two votes against, I'll close this and migrate the functionality to a separate package. Stay tuned for a link.

That said, to defend my 🛡honor🛡, a few counterpoints to @rplzzz :

> Well, in that case you're manually modifying R scripts...it serves as the interface between the developer's brain and the automated system

Fair enough, _but_ chances are if I'm using the Hector R package, I know my way around R lists and have several convenient ways to work with them (and R scripts in general). Keeping everything in INI files and providing no tools to work with them forces me to find my way around a new syntax (albeit an easy one) and format. In other words, I think the difference between "you can run Hector via R and configure it via the INI file" and "you can run and configure Hector via R" is not insignificant, especially to somebody who is new to coding. In my experience teaching Software Carpentry courses, every additional thing a learner has to think about comes at a cost, and every minor setback amplifies the frustration of using a new tool. Even dumb Windows messages like "You don't have a program installed that can open `.ini` files" can be demoralizing and provide an additional barrier to entry to using a tool.

> If they're both disk files, then it's easy to spot with diff.

Per my comment above, now we're introducing a _third_ utility -- one that not only requires knowing one's way around the command line, but that also has a different name and different output depending on which system you use (it's apparently [`compare-object`](https://serverfault.com/questions/5598/how-do-i-diff-two-text-files-in-windows-powershell) on Windows Powershell).

> every Hector application has to start with a bunch of boilerplate code

If they're default scenarios, then there's no reason we couldn't store them as R objects inside the package, as we do with the current variable names.

```r
RCP26 <- function() {
  modifyList(read_ini("hector_defaults.ini"), list(...))
}  

RCP45 <- function() {
  modifyList(read_ini("hector_defaults.ini"), list(...))
}
```

Also, in the current system, rather than maintaining <20 lines of R code per scenario (if we used `read_ini` + `modifyList`; possibly less with some clever use of `lapply`), we have to maintain 300-odd lines of INI file per scenario, which I would argue is no less "boilerplate" and, as I said before, obscures differences between files. Moreover, every time we need to add a new INI file, we invariably produce [large diffs that are hard to review](https://github.com/JGCRI/hector/pull/279/commits/c2bc7a1c911c01688f193af0c84d9c8144613845#diff-3b386e2f62d5c40623e5c67f3d9636a3R1); in that example, adding 3 scenarios which are only slightly different from the Hector defaults increases the size of the repository by over 1000 lines of code. When we eventually want to support the 5 (or however many) scenarios, we will need to add another 1500 lines. And that's hundreds of extra lines of text added _for every INI file changed_ to anyone's project using Hector. And any future changes to the Hector INI format will have to be propagated to each scenario file. Which, yeah, can be done with `sed` or `gsub` or a perl one-liner or whatever...but why add the extra challenge?

bpbond:
Honor well defended!

I definitely agree that having a programmatic way to modify INIs is great, for many of the reasons you outline above. I'm just not sure here is where we want that functionality.

ashiklom:
Functionality migrated to [hectortools](https://github.com/jgcri/hectortools). I'll try to flesh out that package a little more when I get the chance.

rplzzz:
> > If they're both disk files, then it's easy to spot with diff.

> Per my comment above, now we're introducing a third utility -- one that not only requires knowing one's way around the command line, but that also has a different name and different output depending on which system you use (it's apparently compare-object on Windows Powershell).

@ashiklom Fair enough, I guess that's a bit of a blind spot for me.  I've been using unix since 1988, and I sometimes forget that there are other OSes.  I'm not sure why, but there are.

For some of these other issues, I think that the interested parties need to sit down and have a think about how we want to use hector ourselves, how we think other people might want to use hector (and how much of that we want to support), and how we want to do configuration management into the future.  I definitely want to see everybody's preferred way of operating supported, provided we can do that without creating an undue burden for ourselves supporting it.

A github PR review is not really the ideal environment for having a discussion like that, and I should have saved my musings along those lines for a time when we could have a productive discussion on the subject, instead of sounding needlessly critical.  I am sorry for that.

PR #288, opened 2019-02-09
Add lookup table for units of settable variables.

rplzzz:
I found this unpushed branch when I was working on that other issue earlier today.  It provides a units lookup function for all settable quantities, so that users don't have to go find the unit they're looking for in the documentation every time they go to set something.


data-raw/units-data.R
```diff
@@ -0,0 +1,34 @@
...
+              EMISSIONS_HCF22(), EMISSIONS_HCF141B(), EMISSIONS_HCF142B(),
+              EMISSIONS_HALON1211(), EMISSIONS_HALON1301(),
+              EMISSIONS_HALON2402(), EMISSIONS_CH3CL(), EMISSIONS_CH3BR())
+halounits <- 'Gg'
+
+
+unitstable <- rbind(data.frame(variable=hvars, units=hunits),
+                    data.frame(variable=haloemis, units=halounits))
+
+devtools::use_data(unitstable, internal=TRUE, overwrite=TRUE)
```
bpbond: `use_data()` is deprecated as of `devtools` v2. Get ahead of this and change to `usethis::use_data()`?

PR #290, opened 2019-02-10
Automatically reset hector cores when their past has been altered

rplzzz:
This patch differs from the proposals advanced in #hector in two ways.
1. There is still just one single function (`setvar`) for setting variables in the API.  Whether a reset is needed or not is checked by looking at whether the `setvar` call will alter timesteps that have already been run.  If so, the core must reset to the last unchanged date.  Changes to parameters (variables whose time is `NA` are deemed to affect the run at all times and therefore require the spinup to be rerun.
2. The reset is not run immediately.  Instead, the core is marked as needing a reset, and `run` will reset any core so marked before proceeding to run it.  This allows us to set several parameters and only perform the reset and spinup rerun once we've made all the changes we wish to make.

### DIFF
--- a/R/messages.R
+++ b/R/messages.R
@@ -81,5 +81,9 @@ setvar <- function(core, dates, var, values, unit)
     unit[is.na(unit)] <- '(unitless)'
     sendmessage(core, SETDATA(), var, dates, values, unit)
 
+    if(any(dates <= getdate(core)) || any(is.na(dates))) {
+        core$clean <- FALSE
+    }
+
     invisible(NULL)
 }
--- a/R/messages.R
+++ b/R/messages.R
@@ -82,6 +82,15 @@ setvar <- function(core, dates, var, values, unit)
     sendmessage(core, SETDATA(), var, dates, values, unit)
 
     if(any(dates <= getdate(core)) || any(is.na(dates))) {
+        rdate <- min(dates) -1
+        if(is.na(rdate))
+            rdate <- 0
+
+        if(core$clean)
+            core$reset_date <- rdate
+        else
+            core$reset_date <- min(rdate, core$reset_date)
+
         core$clean <- FALSE
     }
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hector
 Title: Run the Hector Simple Climate Model
-Version: 0.2.0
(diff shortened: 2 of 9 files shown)


R/messages.R
```diff
@@ -81,5 +81,18 @@ setvar <- function(core, dates, var, values, unit)
...
+            rdate <- 0
+
+        if(core$clean)
+            core$reset_date <- rdate
+        else
+            core$reset_date <- min(rdate, core$reset_date)
+
+        core$clean <- FALSE
+    }
+
```
ashiklom: Not I defer to you (all) as the more experienced coders, but FWIW, both the [Google](https://google.github.io/styleguide/Rguide.xml#curlybraces) and [Tidyverse](https://style.tidyverse.org/syntax.html#indenting) R style guides recommend using braces around `if-else` statements, even if they are not technically required for single lines.

Both those style guides also suggest spaces after `if` (as in `if (condition)` rather than `if(condition)`).

Personally, I'm also generally wary of using `$` because it does partial matching. So, if we changed `clean` to `clean_core` at some later point, `core$clean` would still work. For internal code like this, it probably shouldn't matter, but I've been bitten by this multiple times in the PEcAn code base and have been wary of it ever since.

rplzzz:
> In general, looks good to me.
> 
> A functional programming purist might complain that this approach (1) requires keeping track of an additional, mutable state (`clean`), and (2) because it adds a hidden, conditional side-effect to the `setvars` function. But that's mostly food for thought, not a reason to hold this up.
> 
> Purely for the sake of discussion, here's what my implementation of `set_params` looked like: [ashiklom@6d91f9c](https://github.com/ashiklom/hector/commit/6d91f9c9bb68fa5c59fe53b2538276c118296ea7).

The hector cores are already a spectacular break with functional programming style because they store all kinds of state; it's just that most of that is hidden (from R users) in hector's  C++ code.  In theory I could have done the same here, adding two member variables to the `Hector::Core` class and checking it at the start of `Hector::run`, but this way was quicker to implement and more amenable to testing.

R/messages.R
bpbond: I noticed the brace thing too but thought I was being overly picky. Note there are cases where this makes a difference, e.g. I don't think `covr` can compute code coverage for the two cases without them.

PR #291, opened 2019-03-19
Add volcanic scaling factor

rplzzz:
As we worked through the Hector calibration project, when we looked at individual CMIP models it became clear that they vary wildly in the strength of their response to volcanic activity.  Here's an example that @kdorheim found of a model that has no apparent volcanic response:
image
Notice that Hector had to crank the diffusivity all the way up to 20.4 to fit this model because it had no other way to suppress its response to the Krakatoa (1883) and Mt. Agung (1963) eruptions.  This is probably not what we want.

We can use the difference between 1960 and 1965 temperatures as a measure of the volcanic response of a model.  For most models the difference is somewhere around a quarter of a degree, but for a few it is as high as half a degree, and there are some outliers with large negative values.
image

This patch creates a fourth temperature parameter, called `VOLCANIC_SCALE`.  This parameter works similarly to the `AERO_SCALE` parameter.  For `VOLCANIC_SCALE==1`, volcanic forcing has its normal effect.  Values greater than 1 enhance the effect, while values less than 1 suppress it.  This should allow us to fit ESMs with larger or smaller volcanic response without having to distort the other temperature parameters.

image
Comparison of Hector runs with `VOLCANIC_SCALE` set to its default value of 1.0 and an alternate value of 1.5, which represents an enhancement of volcanic effects.


rplzzz:
@JGCRI/hector It just occurred to me that this change probably merits a minor version bump, since it's adding a new feature.  The R package version number is still at 0.3, so I'm going increment that to 0.4.  Sometime soon we should probably sync the R package version to the Hector version, but I don't want to do that in the middle of a substantive revision.

The current Hector version is 2.1.0.  Therefore, I suggest we call this revision 2.2.0

bpbond:
>We can use the difference between 1960 and 1965 temperatures as a measure of the volcanic response of a model.

Why? Is there a particularly large eruption(s) and little other climate change in this period?

inst/input/hector-gcam.ini
```diff
@@ -139,6 +139,7 @@ enabled=1
 S=3.0 				; equilibrium climate sensitivity for 2xCO2, degC
 diff=2.3			; ocean heat diffusivity, cm2/s
 alpha=1.0     ; scaling factor for aerosol forcing
+volscl=1.0    ; scaling factor for volcanic forcing
```
bpbond: Perhaps both scaling factors' comments should note centered around 1, i.e. 1.0 = no scaling?

tests/testthat/test_parameters.R
```diff
@@ -124,3 +123,20 @@ test_that('Lowering aerosol scaling factor increases temperature',
     shutdown(hc)
 })
 
+test_that('Increasing volcanic forcing scaling factor increases the effect of volcanism', {
```
bpbond: 👍

rplzzz:
> > We can use the difference between 1960 and 1965 temperatures as a measure of the volcanic response of a model.
> 
> Why? Is there a particularly large eruption(s) and little other climate change in this period?

Yes, there is a particularly large eruption during that period.  The temperature in the years leading up to that time is relatively stable, for Hector, at least.  For other models it varies a bit.  Some are stable in that period, some less so.  If I cared a lot about creating an index of volcanic sensitivity, I would do something like I described in the discussion on the Hector calibration channel, viz., averaging over all of the big eruptions in the XIXth and XXth centuries, but I don't expect that we're going to be making extensive use of it, so I didn't bother.

rplzzz:
Derp, I forgot to push the commit that updates the changelog and version numbers.  Should be ready to go again as soon as Travis does its thing.

src/temperature_component.cpp
```diff
@@ -441,8 +449,8 @@ unitval TemperatureComponent::getData( const std::string& varName,
             returnval = heatflux;
         } else if( varName == D_ECS ) {
             returnval = S;
-        } else if(varName == D_AERO_SCALE) {
-            returnval = alpha;
```
cahartin: why did you delete the alpha here?

rplzzz: It was a duplicate.  Alpha is already processed on lines 434-435 (old numbering).

cahartin: ah, gotcha!

PR #295, opened 2019-04-05
Add API for "fraction" parameters

ashiklom:
This allows the parameters `f_nppv/d`, `f_litterd`, and `f_lucv/d` to be set and retrieved via the `sendMessage/getMessage` API (and therefore via the R package).

I also changed the unit of all of these parameters from `UNDEFINED` to `UNITLESS`, which seemed more appropriate and allowed these to be set more naturally through R (i.e. without having to explicitly set the unit to `(undefined)`).


src/simpleNbox.cpp
```diff
@@ -436,6 +441,28 @@ unitval SimpleNbox::getData(const std::string& varName,
     } else if( varName == D_RF_T_ALBEDO ) {
         H_ASSERT( date != Core::undefinedIndex(), "Date required for albedo forcing" );
         returnval = Ftalbedo.get( date );
+
+        // Partitioning parameters 
+    } else if(varName == D_F_NPPV) {
+        // For the time being, we are only supporting global beta, not biome-specific
```
rplzzz: I believe this comment, and the ones like it below, are left over from copying this code block.

ashiklom: Good catch. Addressed in 245d998.

bpbond:
Good catch @rplzzz . Looks good to merge?

PR #297, opened 2019-04-12
Set/retrieve biome-specific parameters via R API

ashiklom:
This PR includes the following:

1. Add biome-specific versions of terrestrial carbon cycle parameters `q10_rh`, `f_nppv/d`, `f_litterd`.
2. Allow vegetation, detritus, and soil C pools, the baseline NPP (`npp_flux0`), and the biome-specific warming factor to be set and retrieved by the R API.
3. More generally, allow biome-specific parameters and states to be set and retrieved by the R API (and as a side effect, remove the dependency on `hectortools` for testing).

The biome system in Hector is still a bit fragile, and this PR doesn't do much to change that, except that it creates a benchmark against which future revisions to the biome system can be tested.


src/simpleNbox.cpp
```diff
@@ -408,93 +433,110 @@ unitval SimpleNbox::getData(const std::string& varName,
...
+    } else if( varNameParsed == D_PREINDUSTRIAL_CO2 ) {
         H_ASSERT( date == Core::undefinedIndex(), "Date not allowed for preindustrial CO2" );
         returnval = C0;
-    } else if(varName == D_BETA) {
-        // For the time being, we are only supporting global beta, not biome-specific
+    } else if(varNameParsed == D_BETA) {
         H_ASSERT(date == Core::undefinedIndex(), "Date not allowed for CO2 fertilization (beta)");
-        returnval = unitval(beta[SNBOX_DEFAULT_BIOME], U_UNITLESS);
-    } else if(varName == D_Q10_RH) {
+        returnval = unitval(beta[biome], U_UNITLESS);
```
rplzzz: This is in `getdata()`, right?  So this means that if I send a message requesting a parameter for a biome that doesn't exist, an entry will be created under that biome, but _only_ for that parameter.  Is that going to create a problem.  Maybe it's harmless if nowhere else do we reference that biome, but if we ever infer the list of biomes by iterating over that parameter, then the calculation is going to turn into a charlie-foxtrot in a hurry.  

I think I've commented on this before.  If we're going to start using the code's biome capability, then we really need to be more intentional about creating biomes and keeping track of them.

src/simpleNbox.cpp
```diff
@@ -145,14 +155,20 @@ void SimpleNbox::setData( const std::string &varName,
         else if( varNameParsed == D_VEGC ) {
             H_ASSERT( data.date == Core::undefinedIndex(), "date not allowed" );
             veg_c[ biome ] = data.getUnitval( U_PGC );
+            // HACK: Set value at t=0
+            veg_c_tv.set(0, veg_c);
```
rplzzz: So, is this about making sure that setting `veg_c` works if it happens at some time after startup?  And before we were assuming that this would only be set from the ini file (not by message)?

src/simpleNbox.cpp
```diff
@@ -145,14 +155,20 @@ void SimpleNbox::setData( const std::string &varName,
         else if( varNameParsed == D_VEGC ) {
             H_ASSERT( data.date == Core::undefinedIndex(), "date not allowed" );
             veg_c[ biome ] = data.getUnitval( U_PGC );
+            // HACK: Set value at t=0
```
rplzzz: I think we can do better.

src/simpleNbox.cpp
```diff
@@ -220,7 +236,7 @@ void SimpleNbox::setData( const std::string &varName,
         }
         else if( varNameParsed == D_Q10_RH ) {
             H_ASSERT( data.date == Core::undefinedIndex() , "date not allowed" );
-            q10_rh = data.getUnitval(U_UNITLESS);
+            q10_rh[ biome ] = data.getUnitval(U_UNITLESS);
```
rplzzz: So, something like the case in `getData()`, setting _any_ of these with a biome that doesn't already exist will create it without ensuring that it is created consistently across all the other variables.  Maybe it's harmless, maybe not.  I know that we do occasionally iterate over these maps indiscriminately.  Can we show that having a bogus biome in one or more of these arrays will never cause a problem?  Can we ensure that nobody somewhere down the line will create a situation where it _will_ cause a problem?

src/simpleNbox.cpp
ashiklom: Without this line, `veg_c` is overwritten during reset by a line like `veg_c = veg_c_tv.get(t)`, which eliminates the newly-created biome.

rplzzz: Thanks for clarifying.

src/simpleNbox.cpp
ashiklom: > Maybe it's harmless, maybe not.

I'm fairly certain it's _not_ harmless. I do know there are some runtime checks that can throw errors about mismatched sizes, but it's not immediately obvious that biome specification is what causes them.

src/simpleNbox.cpp
ashiklom: > So this means that if I send a message requesting a parameter for a biome that doesn't exist, an entry will be created under that biome, but only for that parameter.

This is true. 

>  Is that going to create a problem?

Almost definitely.

Will using the `.at` method alleviate this? I've seen comments elsewhere in the code that `.at(x)` will throw an error if `x` is not in the object, but I haven't checked.

ashiklom: > then we really need to be more intentional about creating biomes and keeping track of them

Definitely agree. I mostly did this to have something working for my permafrost analysis. I think this PR can still be a useful starting point for discussions about how to proceed _properly_, even if this version of the code never sees the light of `master`.

src/simpleNbox.cpp
ashiklom: It's worth noting that this particular line, along with other changes related specifically to Q10, are needed to make Hector behave as documented. Prior to this change, Hector actually had _absolutely no mechanism_ for setting or using biome-specific Q10, even though the INI file indicates that it can (not that the INI file is an authoritative source of documentation, as we've discussed before, but even so).

src/simpleNbox.cpp
ashiklom: I agree.

Does it make sense to use `t = data.date` if date is provided and `t = 0` if `date` is missing? Or do we just force users to provide the first year of the simulation?

bpbond: My gut reaction is they have to provide the date.

bpbond:
Thanks @ashiklom . Looking through things, FWIW I agree with @rplzzz and you that if we're going to expose this via API, let's do it right, or at least right-er than the current mechanism, which as you guys note is fragile and awkward.

src/simpleNbox.cpp
rplzzz: I agree with @bpbond on this.  If something is time variable, then you should have to supply a date to get or set it.  The only places where we allow a default date for something that varies over time are places where it was done to maintain backward compatibility (specifically, variables where we used to keep only a current value, and model components relied on that behavior when exchanging data internally).

However, my remark referred to the comment itself.  Please don't put comments like "this is a hack" in your code; it's completely unhelpful to anyone else who comes along later and reads it.  It only takes a few minutes to make a note of what you were trying to do and why it had to be done in a hacky way.  Take those few minutes.

PR #303, opened 2019-05-02
Fix bugs in reporting for forcings.

kdorheim:
Change the bindings between R and C++ code so that the RF_H2O() refer to H2O RF.


rplzzz:
@kdorheim I can't add you as a reviewer, since this is originally your PR, but if you would, please, take a look at my changes (you might want to filter to just the commit that I added), and let me know if you see anything that could use some improvement.

rplzzz:
The code won't build with GCC because I used a C++-11 feature, and g++ is configured to do C++-0x compatibility.  That should be an easy fix to the makefile.  

I'm not sure why the R build succeeds, whether it's because w'ere using clang on the travis R build, or whether the default flags for g++ are different under Rcpp.  Either way, while we're at it, we should probably configure the Rcpp build explicitly to specify C++-11.  We've ben getting compiler warnings about C++-11 features the whole time, and we wouldn't want someone's build to fail because their compiler didn't get the memo.

I probably won't get to any of this until after my telecon tomorrow afternoon.  In the meantime, this branch should be usable, just not quite ready to merge.

ashiklom:
> configure the Rcpp build explicitly to specify C++-11

I _think_ the [recommended](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Using-C_002b_002b11-code) way to do this is to add the following line to the `src/Makevars` file:

```
CXX_STD = CXX11
```

> easy fix to the makefile

I think this is just a matter of adding `-std=c++11` to the end of the `CXXFLAGS` list in `src/makefile.standalone`. Though my understanding of exactly which variables control what in the C compilation process is a bit hazy, so I defer to the experts.

rplzzz:
Well, on the windows side we still get warnings about C++-11 features, but at least it builds.  I guess any compile you can walk away from is a good one.

On the mac R builds we're now building cleanly, with no warnings at all, so I'm calling this a win.

R/aadoc.R
```diff
@@ -76,7 +76,9 @@ NULL
 #' These identifiers specify forcing values that can be provided by hector via
 #' one of the myriad halocarbon components.  All of the values corresponding to
 #' these identifiers are read-only (\emph{i.e.}, they can only appear in
-#' \code{\link{GETDATA}} messages.)
+#' \code{\link{GETDATA}} messages.)  The forcings returned are the
+#' \emph{adjusted} forcings, with the base year (typically 1750) values
```
bpbond: "Adjusted" isn't very clear. Perhaps change to "relative"?

changelog.txt
```diff
@@ -1,3 +1,19 @@
+Hector 2.2.2
+============
+* Fix bug that was causing requests for H2O forcing in the R interface
+  to return N2O forcing instead (the model internals were
+  unaffected). 
+* Fix bug that was causing API requests for halocarbon forcing to
+  return absolute forcing values, rather than values relative to the
+  base year (which is what is done for all other forcings). 
+* Add missing RF_VOL dependency to forcing component.  In practice the
```
bpbond: "RF_VOL (volcano forcing)"

inst/include/component_data.hpp
```diff
@@ -82,6 +82,41 @@
 #define D_RF_halon2402          D_RF_PREFIX halon2402_COMPONENT_BASE
 #define D_RF_CH3Cl              D_RF_PREFIX CH3Cl_COMPONENT_BASE
 #define D_RF_CH3Br              D_RF_PREFIX CH3Br_COMPONENT_BASE
+
+// Adjusted halocarbon forcings
```
bpbond: Here too; "adjusted" or "relative"? To me, the latter is more precise and clearer.

tests/testthat/test_scenarios.R
```diff
@@ -20,7 +20,7 @@ test_that('RCP scenarios are correct', {
 
         ## Get the comparison data
         sampleoutfile <- sprintf('sample_outputstream_rcp%s.csv', rcp)
-        sampledata <- read.csv(file.path(sampledir, sampleoutfile), comment.char = '#')
+        sampledata <- read.csv(file.path(sampledir, sampleoutfile), comment.char = '#', stringsAsFactors=FALSE)
```
bpbond: For consistency, spaces around `=`

R/aadoc.R
rplzzz:
```suggestion
#' \emph{relative} forcings, with the base year (typically 1750) values
```

changelog.txt
rplzzz:
```suggestion
* Add missing RF_VOL (volcanic forcing) dependency to forcing component.  In practice the
```

tests/testthat/test_scenarios.R
rplzzz:
```suggestion
        sampledata <- read.csv(file.path(sampledir, sampleoutfile), comment.char = '#', stringsAsFactors = FALSE)
```

inst/include/component_data.hpp
rplzzz:
```suggestion
// Relative halocarbon forcings
```

rplzzz:
```suggestion
// Adjusted (i.e., relative) halocarbon forcings
```

Issue #306, opened 2019-05-18
Initial atmospheric concentration isn't equal to preindustrial co2 setting


rplzzz:
During the spin-up process the atmosphere has some net absorption from the other carbon pools in the model, to the tune of about 11.7 ppmv.  
![image](https://user-images.githubusercontent.com/7209964/57969536-ff091380-7945-11e9-9231-7c3e0a820647.png)

It's not totally clear that the 1745 value _should_ be equal to the preindustrial value, since the definition of "preindustrial" is a bit nebulous.  However, it also seems like our discrepancy between the preindustrial and initial concentrations is not the result of a considered assessment of the likely change in concentration from preindustrial times to the start of the simulation.  Rather, it seems to be a byproduct of trying to equilibrate the model before the run begins in earnest.

I can see a couple of ways of dealing with this situation.
1. Ignore it.  We calibrate the parameter to match ESM outputs, so as long as we recognize that its value shouldn't be taken as literally representative of the concentration at some particular past time it doesn't really present (much of) a problem.
2. Apply an adjustment at initialization.  Since the discrepancy is pretty constant, we could initialize the atmospheric carbon pool to a value 11.7 ppmv less than the input parameter.  Then, by the end of the spinup the value should be pretty close to the value specified by the preindustrial concentration parameter.  The disadvantage to this method is that it is brittle.  Changes to the model will probably affect the offset, so we would have to monitor it.  (That said, it wouldn't be hard to write a test that would do this.)
3. Apply an adjustment during spinup.  At each time step we could add or remove carbon as required to bring the atmospheric concentration back to the intended initial value.  This is more work, but it is probably the closest to the right way to handle the matter.

Is this something we want to tackle?

bpbond:
>Rather, it seems to be a byproduct of trying to equilibrate the model before the run begins in earnest.

Well, it's not a _byproduct_; after all the whole point of the spinup is to bring the C cycle to equilibrium. But you raise a good point, and one that we should at the very least be clearer about to model users: the "initial (preindustrial) pools and fluxes" given in an `ini` file is _not_ what the model actually starts with in 1745 or whenever. 

[...]

PR #307, opened 2019-05-19
Constrain CO2 conc to be equal to preindustrial during spinup.

rplzzz:
Constrain CO<sub>2</sub> concentrations during spinup to be equal to the preindustrial values.  This will result in the concentration at the start of the simulation being equal to the `PREINDUSTRIAL_CO2` parameter, which was not previously the case.

See further discussion in #306.

Since this changes the results slightly, but does not otherwise affect the design of the model, I called it a minor version increment to 2.3.0.


rplzzz:
I didn't include the code and data for the validation runs I did in the repository, since the comparison data is rather large, and probably won't be of interest once these changes are decided on.  You can find the results of those experiments here:  http://rpubs.com/Robert_Link/497501

PR #309, opened 2019-05-31
Update version of Boost on Travis

ashiklom:
As Travis upgraded to Ubuntu 16 (xenial), the version and source of Boost libraries changed. The new default version is 1.58, and it is not in the core Ubuntu repositories (i.e. does not require a PPA).

Previously, this was causing build errors like the following:

```
Package libboost-filesystem1.55-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
Package libboost-system1.55-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
E: Package 'libboost-system1.55-dev' has no installation candidate
E: Package 'libboost-filesystem1.55-dev' has no installation candidate
```

I also removed the dependencies on Boost libraries from the R test since they have not been required since #267.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,8 @@ matrix:
       addons: 
         apt: 
           packages: 
-            - libboost-system1.58-dev
-            - libboost-filesystem1.58-dev
+            - libboost-system-dev
+            - libboost-filesystem-dev
 
       # BOOSTROOT is set to look for header files in the default place
       #where Ubuntu apt will install them --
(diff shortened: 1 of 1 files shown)


rplzzz:
Hmm.  apt-get can't find `libboost-system1.58-dev`.  Why do we even have to give a version number?  Is there really not a `libboost-system-dev` package that gives whatever the latest version is?

ashiklom:
Super weird, because it had no trouble with my fork: https://travis-ci.org/ashiklom/hector/jobs/539843941

I figured we had a reason for hard-coding the version, but I'm happy to remove it. Let's see if it works with the versionless package.

ashiklom:
That seems to have worked. https://travis-ci.org/JGCRI/hector/jobs/539863372

What's interesting is that the JGCRI remote is building on Ubuntu `trusty`, whereas my fork was building on `xenial`. _Why_ that's happening is beyond me -- maybe because of caching?

rplzzz:
Hard to say.  Let's call it a mystery for another time.

PR #311, opened 2019-05-31
Eliminate `hectortools` dependency from biome tests

ashiklom:
Our multi-biome unit tests required my `hectortools` package to read, modify, and write the INI file. It was only `Suggests`-ed, but even so, Travis insisted on installing it during our builds. But `hectortools` itself depends on `hector`, so, to install `hectortools`, Travis actually installed an older version of `hector` from CRAN before installing ours. This was my fault, from PR #285, and while it's not a _major_ problem, the added build delay and the conceptual problem of circular dependency has been irritating me for a while.

This PR performs the same tests, but replaces `hectortools` functions with base-R string manipulation of the INI files.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,10 +22,7 @@ Suggests:
     rmarkdown,
     ggplot2,
     tidyr (>= 0.7),
-    dplyr (>= 0.7),
-    hectortools
-Remotes:
-    jgcri/hectortools
+    dplyr (>= 0.7)
 SystemRequirements: Boost C++ libraries
 Collate:
     'aadoc.R'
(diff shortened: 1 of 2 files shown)


rplzzz:
> Travis actually installed an older version of `hector` from CRAN before installing ours.

Is hector on CRAN?  I didn't think we ever submitted it.  At one time we thought about it, but ended up kicking it down the road because we expected them to give us static about the package notes due to the size of the sample outputs.  Did we manage to get around this somehow?

ashiklom:
> Is hector on CRAN?

You're right, it's not. `hectortools` is installing it from GitHub via the `DESCRIPTION -> Remotes:` mechanism. I'll need to update the documentation in #310 accordingly.

PR #317, opened 2019-06-27
Revamped multi-biome system

### LINKED ISSUE #20: Global versus biome pools
Minor   If there are any 'global' pools, should not also be biomes! N   simpleNbox.cpp      Ben     Open

### LINKED ISSUE #286: Have a canonical biome list defined in INI file
that then can be iterated over generally. Proposed by @rplzzz here https://github.com/JGCRI/hector/pull/285#discussion_r251205251

ashiklom:
- There is now a global `biome_list` (`std::vector<std::string>`) that stores the current list of biomes in use by the Hector core. All loops related to biome-specific operations now loop over this list instead. Consistency of biome-specific variables against this list is rigorously checked before running, and high-level user-facing functions have additional protections that make it very difficult to create an incomplete or nonsensical biome. The biome list can be accessed from the R interface via the `get_biome_list` function.
- Functions related to messages (`setData`, `getData`) are now equipped to handle biome-specific versions of variables. The C function `setData` can technically create new, incomplete biomes (for compatibility with the way we currently load data from the INI file), but trying to run Hector with an incomplete biome will throw a loud, early error. The R function `setvar` will _not_ let you set a variable for a biome that does not exist.
- Biome specific capability identifiers (e.g. `BETA()`, `VEG_C()`) have gained an optional `biome` argument (e.g. `BETA("permafrost")`, `VEG_C("tropical")`). For vegetation, detritus, and soil pools, omitting the biome in `fetchvars` will retrieve the sum of that pool across all biomes. For parameters, the biome is required unless the only biome is `"global"` (this preserves back-compatibility).
- I wrote several new higher-level C functions for working with biomes: `createBiome`, `deleteBiome`, and `renameBiome`. These in turn are wrapped by R functions that automate common biome manipulations while preserving Hector's internal mass balance: `split_biome` and `merge_biomes`.
- All of this comes with fairly comprehensive unit tests that also double as a useful reference for how all this can work. In addition, I wrote up a vignette showing two practical examples of this in action ([rendered version](https://ashiklom.github.io/hector/articles/multiple-biomes.html)).

Probably 75% of the work on this PR was figuring out how everything fit together, what order and under what conditions various `SimpleNbox` methods are called, etc. So I'm very open to your suggestions for alternative designs, especially at the C level, even if such designs require basically re-writing much of this from `master`.

Closes #286 
Closes #20


R/biome.R
```diff
@@ -0,0 +1,181 @@
...
+#' @return Hector core, invisibly
+#' @author Alexey Shiklomanov
+#' @export
+create_biome <- function(core, biome,
+                         veg_c0 = 0,
+                         detritus_c0 = 0,
+                         soil_c0 = 0,
+                         npp_flux0 = 0,
+                         warmingfactor = 1,
+                         beta = 0.36) {
```
ashiklom: Note that Q10 is _not_ currently a biome-specific paramater, _even though we say it is_. Unlike these changes, which are all new features and do not affect existing behavior, making Q10 a biome-specific parameter would be a breaking change, so I deliberately left it out of this PR. If we agree on this design, then I'll make that change in a PR hot on the heels of this one.

inst/include/simpleNbox.hpp
```diff
@@ -198,6 +205,15 @@ class SimpleNbox : public CarbonCycleModel {
     double sum_map( double_stringmap pool ) const;      //!< sums a double map (collection of data)
     void log_pools( const double t );                   //!< prints pool status to the log file
     void set_c0(double newc0);                          //!< set initial co2 and adjust total carbon mass
+
+    template <class T_data>
+    void add_biome_to_ts(tvector<std::map<std::string, T_data> >& ts, const std::string& biome, T_data init_value); 
+    template <class T_map>
+    void remove_biome_from_ts(tvector<T_map>& ts, const std::string& biome); 
+    template <class T_map>
+    void rename_biome_in_ts(tvector<T_map>& ts, const std::string& oldname, const std::string& newname);
```
ashiklom: These are all templates so that I can re-use them for both `unitval` time-series (`veg_c_tv` and friends) and regular `double` time-series (like `tempfertd` and `tempferts`).

src/simpleNbox.cpp
```diff
@@ -18,6 +18,8 @@
 #include "simpleNbox.hpp"
 #include "avisitor.hpp"
 
+#include <algorithm>
```
ashiklom: Needed for `std::find`.

src/simpleNbox.cpp
```diff
@@ -113,9 +127,36 @@ void SimpleNbox::setData( const std::string &varName,
...
+                "' because you cannot have both 'global' and biome data. " << std::endl;
+            // We don't use the `deleteBiome` function here because
+            // when `setData` is used to initialize the core from the
+            // INI file, most of the time series variables that
+            // `deleteBiome` modifies have not been initialized yet.
+            // This should be relatively safe because (1) we check
+            // consistency of biome-specific variable sizes before
+            // running, and (2) the R interface will not let you use
+            // `setData` to modify the biome list.
+            biome_list.erase( it_global );
```
ashiklom: This is currently failing on Travis because of a known bug in old versions of GCC/clang:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=57158
https://stackoverflow.com/questions/19559235/missing-const-iterator-overload-of-stdvectorerase-with-g-4-8/19559316

The root cause of this is probably that the JGCRI/Hector Travis instance is still using [Ubuntu Trusty](https://travis-ci.org/JGCRI/hector/jobs/551480463#L9); the same code built without issue on my fork because my Travis containers have been [upgraded to Ubuntu Xenial](https://travis-ci.org/ashiklom/hector/jobs/551467449#L9).

I'm perfectly happy to change this from a `const_iterator` to just a plain `iterator`, which should fix this specific issue, but I'm not totally clear on the consequences of that, so I defer to you all for advice.

bpbond: Is there a reason _not_ to change the Hector repo to use Xenial?

R/biome.R
bpbond: Why have default values for these parameters? Shouldn't the user be required to specify them?

R/biome.R
```diff
@@ -0,0 +1,181 @@
...
+#'
+#' @param core Hector core
+#' @param old_biomes Names of biomes to be merged
+#' @param new_biome Name of new merged biome
+#' @param ... Additional parameters for the new biome, as set by
+#'   [create_biome()].
+#' @return `core`, invisibly
+#' @author Alexey Shiklomanov
+#' @export
+merge_biomes <- function(core, old_biomes, new_biome, ...) {
```
bpbond: Maybe I'm not imaginative enough...what's the use case for this? I.e. when would the user ever set up biomes, and then merge then?

src/simpleNbox.cpp
```diff
@@ -143,16 +184,33 @@ void SimpleNbox::setData( const std::string &varName,
...
+            // provided, set only the "current" model pool, without
+            // touching the time series variable. This is to
+            // accommodate the way the INI file is parsed. For
+            // interactive use, you will usually want to pass the date
+            // -- otherwise, the current value will be overridden by a
+            // `reset` (which includes code like `veg_c = veg_c_tv.get(t)`).
+            //
+            // TODO: Should we create special variables `veg_c0`,
+            // `soil_c0`, etc. that are set in the INI file and
+            // require a date for `veg_c`, `soil_c`, etc.?
```
bpbond: I.e. matching the approach taken with NPP?

src/simpleNbox.cpp
```diff
@@ -998,4 +1131,201 @@ void SimpleNbox::set_c0(double newc0)
...
+    }
+}
+
+// Remove a biome from a time-series map variable
+template <class T_map>
+void SimpleNbox::remove_biome_from_ts(tvector<T_map>& ts,
+                                      const std::string& biome) {
+    // We don't need to check for presence of `biome` here because the
+    // `<std::map>.erase()` method is effectively a no-op when given a
+    // non-existing key.
```
bpbond: "non-existent"

tests/testthat/test_biome.R
```diff
@@ -81,4 +85,154 @@ test_that("Hector runs with multiple biomes.", {
...
+  permafrost_veg <- sendmessage(core, GETDATA(), VEG_C("permafrost"), 0, NA, "")[["value"]]
+  expect_equivalent(default_veg, (1 - fsplit) * global_veg)
+  expect_equivalent(permafrost_veg, fsplit * global_veg)
+  expect_equivalent(default_veg + permafrost_veg, global_veg)
+  invisible(run(core))
+  r_biome <- fetchvars(core, 2000:2100)
+  expect_equivalent(r_global, r_biome)
+
+  # Sum of biome-specific pools should exactly equal global results at
+  # each time step. This is a robust test!
```
bpbond: 👍

vignettes/multiple-biomes.Rmd
```diff
@@ -0,0 +1,135 @@
...
+Before we begin, let's load the R package we will need: the `hector` package to interact with Hector, and the `ggplot2` package for plotting.
+
+```{r}
+library(hector)
+library(ggplot2)
+```
+
+# Example 1: Accelerated high-latitude warming
+
+Observational and modeling studies have found that high-latitudes, particularly in the northern hemisphere, are warming significantly faster than the global average.
```
bpbond: "high-latitude ecosystems" or "high latitudes"

vignettes/multiple-biomes.Rmd
```diff
@@ -0,0 +1,135 @@
...
+In this example, we will explore the implications of this effect with Hector.
+
+First, lets run a reference case.
```
bpbond: "let's"

vignettes/multiple-biomes.Rmd
```diff
@@ -0,0 +1,135 @@
...
+                                c("low-latitude", "high-latitude"))
+warming_details$variable <- vapply(variable_split, "[[", character(1), 2)
+ggplot(warming_details) +
+  aes(x = year, y = value, color = biome) +
+  geom_line() +
+  facet_wrap(vars(variable), scales = "free_y") +
+  theme_bw()
+```
+
+# Example 2: C3 and C4 plants
```
bpbond: Nice example.

R/biome.R
ashiklom: > Why have default values for these parameters? Shouldn't the user be required to specify them?

I usually air on the side of minimizing the number of things users _have_ to set to minimize user-side friction. The defaults for `warmingfactor` and `beta` are sensible values for which the intuition is not obvious, so I'm inclined to keep those. However, you're probably right that it makes sense to remove the defaults for the initial C pools and NPP, since the current defaults don't actually make sense either.

R/biome.R
ashiklom: In the long term, I was thinking it might be possible to alter the number of biomes dynamically (e.g. have biomes that only exist during a certain point in time). But that's not possible the way this is currently implemented, and might not make sense as an idea anyway, so I think you're right that this is unnecessary. I'll remove it.

vignettes/multiple-biomes.Rmd
ashiklom: Addressed in 551e47f366caff261080a4476d3d556960a14ee8.

vignettes/multiple-biomes.Rmd
ashiklom: Addressed in 551e47f366caff261080a4476d3d556960a14ee8.

vignettes/multiple-biomes.Rmd
ashiklom: Addressed in 551e47f366caff261080a4476d3d556960a14ee8.

vignettes/multiple-biomes.Rmd
ashiklom: Addressed in 551e47f366caff261080a4476d3d556960a14ee8.

src/simpleNbox.cpp
ashiklom: Exactly. I clarified this comment in 55c9341f43ef24026de736c22dd66db8fbc390db.

src/simpleNbox.cpp
ashiklom: Addressed in 55c9341f43ef24026de736c22dd66db8fbc390db.

R/biome.R
ashiklom: Removed in 70bca145c61ce0a0d91764c1383140fbecb94006.

src/simpleNbox.cpp
ashiklom: > Is there a reason not to change the Hector repo to use Xenial?

If we want to support the older OS and compiler. GCC 4.8 is only ~5 years old, and is actually later than the default GCC on, say, PIC (4.4.7). There's no way I would have caught this issue if Travis had upgraded the shared repo to Xenial, but it could have been a nasty surprise for someone trying to install Hector on PIC.

@rplzzz @pralitp -- Do you have any thoughts on this? Should we just abandon support for GCC < 4.9 (I think that's where this was fixed, but I'm not 100% sure), or should I implement some of the casting hacks/workarounds described in the SO question above?

rplzzz: GCC 4.8 may be only 5 years old, but its C++-11 support is a little spotty, which is a good reason not to support it.  We already have to use gcc 6 or later for GCAM anyway.

BTW, if you've been habitually using one compiler on a particular system and you switch to another, you probably need to reinstall all of your R packages (which probably means it's just a good time to upgrade to a new version of R).  Having packages built with a mix of different compilers can lead to weird, random crashes.

ashiklom: Thanks! I explicitly set the Travis build environment distribution to Xenial in 974685f485b27f04313d49f1278f44fdf0354b58.

R/biome.R
ashiklom: Removed defaults for `create_biome` in 2e77808f8f1f5f7da2b1c83d6be4528d2cfb60f5.

src/simpleNbox.cpp
```diff
@@ -998,4 +1132,201 @@ void SimpleNbox::set_c0(double newc0)
...
+    for ( double i = ts.firstdate(); i < ts.lastdate(); i++ ) {
+        if (ts.exists(i)) {
+            currval = ts.get(i);
+            currval.erase(biome);
+            ts.set(i, currval);
+        }
+    }
+}
+
+// Remove a biome from a time-series map variable
```
rplzzz: Comment is inaccurate; this is the rename function, not the remove function.
```suggestion
// Rename a biome in a time-series map variable
```

src/simpleNbox.cpp
```diff
@@ -998,4 +1132,201 @@ void SimpleNbox::set_c0(double newc0)
...
+            currval[newname] = currval.at(oldname);
+            currval.erase(oldname);
+            ts.set(i, currval);
+        }
+    }
+}
+
+// Check if `biome` is present in biome_list
+bool SimpleNbox::has_biome(const std::string& biome) {
+    return std::find(biome_list.begin(), biome_list.end(), biome) != biome_list.end();
```
rplzzz: How often do we search `biome_list` for a value (vs. just iterating over whatever happens to be there)?  If it's something we do often, then we might want to use a `std::set<string>` instead of a vector.  If not, then what we have is fine.

src/simpleNbox.cpp
```diff
@@ -998,4 +1132,201 @@ void SimpleNbox::set_c0(double newc0)
...
+    biome_list.erase( i_biome );
+
+    H_LOG(logger, Logger::DEBUG) << "Finished deleting biome '" << biome << ",." << std::endl;
+
+}
+
+// Create a new biome called `newname`, transfer all of the parameters
+// and pools from `oldname`, and delete `oldname`. Note that the new
+// biome and all of its associated values will be at the end of their
+// corresponding vectors.
```
rplzzz: The associated values are in maps, anyhow, aren't they?  In that case, the last added value will generally not be at the end (and I'm not sure those structures even provide a `back` method in any case).

src/rcpp_constants.cpp
```diff
@@ -881,10 +882,30 @@ return D_Q10_RH;
...
+//' @param biome Biome for which to retrieve parameter. If missing or
+//'   `""`, default to `"global"`.
 //' @export
 // [[Rcpp::export]]
-String BETA() {
-return D_BETA;
+String BETA(String biome = "") {
+  if (biome == "") return D_BETA;
+  // `Rcpp::String` has a `+=` method, but no `+` method, so have to use
+  // this clunky workaround.
```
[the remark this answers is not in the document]
ashiklom: I originally had this as `std::string`, which apparently Rcpp supports directly (translates to `String`?), precisely because it supports the `+` operator. I changed it to `Rcpp::String` for consistency with the other calls, but perhaps we can change to `std::string` here and everywhere else?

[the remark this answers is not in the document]
rplzzz: I suspect that what's happening is that `Rcpp::string` has a constructor for initializing with a `std::string`, and vice versa, so it doesn't matter which one you use, since it will get converted as necessary.  If that's the case, then I'd say use whichever one you like.  I probably used `Rcpp::string` in the other functions because I never thought to check whether a `std::string` would work.

src/simpleNbox.cpp
ashiklom: Is this a basically free performance boost? I.e. The only consequence of changing to `std::set` is that the operation becomes faster (and we get the assurance that entries are unique)? Or are there other side effects I should keep eye out for?

ashiklom: So, as I understand it (which may definitely be wrong), one issue with using a `set` here is that we lose the history of biome creation. The main place where this matters is that the current implementation of `SimpleNbox::createBiome` will initialize a biome with the parameters from the most recently set biome. However, I think this could also introduce some minor surprises on the user side if the position of biomes in the biome list cannot be predicted unless you know the contents of the existing biome list.

I'm inclined to keep this as is right now, but I think it's something we could revisit in the future.

rplzzz: It's a performance boost for searching for a specific biome, but it's a performance hit for iterating over whatever biomes happen to be there.  And you are correct that a set will not allow you to save the history of the biome creation.  I'm ok with the current implementation.  I was just calling out the alternative for you to evaluate.

src/simpleNbox.cpp
ashiklom: Ah, good catch, thanks! This is technically still true of `biome_list` (see my [comment above](https://github.com/JGCRI/hector/pull/317#discussion_r301640644)), but I'll remove the part about the values.

src/rcpp_constants.cpp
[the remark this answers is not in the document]
ashiklom: For consistency and (IMHO) clarity, I'm gonna leave this as `Rcpp::String`. Marking this thread as resolved.

src/rcpp_hector.cpp
```diff
@@ -195,6 +195,61 @@ double getdate(Environment core)
...
+    std::vector<std::string> biome_list = hcore->getBiomeList();
+    return biome_list;
+}
+
+//' Create a biome
+//'
+//' @param core Handle to the Hector instance that is to be run.
+//' @param biome (character) Name of new biome
+// [[Rcpp::export]]
+Environment c_create_biome(Environment core, std::string biome)
```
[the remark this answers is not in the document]
ashiklom: Addressed in c06b34ce906e89322d6323efda5bd6f98ece5b55.

tests/testthat/test_biome.R
```diff
@@ -81,4 +85,148 @@ test_that("Hector runs with multiple biomes.", {
...
+  core <- newcore(system.file("input", "hector_rcp45.ini",
+                              package = "hector"),
+                  suppresslogging = TRUE)
+  b1 <- fetchvars(core, NA, BETA())
+  b2 <- fetchvars(core, NA, BETA("global"))
+  expect_equal(b1$value, b2$value)
+  expect_error(fetchvars(core, NA, BETA("fake")),
+               "Requested biome missing from biome list")
+  expect_error(setvar(core, NA, BETA("permafrost"), 0.5, NA),
+               "Biome 'permafrost' is not in current biome list")
```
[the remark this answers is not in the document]
ashiklom: Addressed in f4d8f1cf57272ddbb5725596d1e40c21ef22a966.

tests/testthat/test_biome.R
```diff
@@ -81,4 +85,148 @@ test_that("Hector runs with multiple biomes.", {
...
+test_that("Correct way to create new biomes", {
+
+  core <- newcore(system.file("input", "hector_rcp45.ini",
+                              package = "hector"),
+                  suppresslogging = TRUE)
+  gbeta <- fetchvars(core, NA, BETA())
+  expect_equal(get_biome_list(core), "global")
+  # Make sure these low-level operations on their own don't throw an error.
+  expect_silent(invisible(c_create_biome(core, "testbiome")))
+  expect_silent(invisible(c_delete_biome(core, "testbiome")))
```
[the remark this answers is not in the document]
ashiklom: Addressed in d9056692c348a0df9a4c057fdb658c30d7fbced5.

src/simpleNbox.cpp
```diff
@@ -408,89 +478,140 @@ unitval SimpleNbox::getData(const std::string& varName,
                             const double date) throw ( h_exception )
 {
     unitval returnval;
+
+    std::string biome = SNBOX_DEFAULT_BIOME;
+    std::string varNameParsed = varName;
+    std::string biome_error = "Requested biome missing from biome list. "
+        "Hit this error while trying to retrieve variable: '" + varName + "'.";
```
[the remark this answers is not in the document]
ashiklom: I think I originally removed the `+` because I was hitting an error that concatenating string literals via `+` was not defined (basically, the problem described [here](https://stackoverflow.com/questions/6061648/concatenate-two-string-literals)). However, adding it back does not throw an error, so I guess that was a different situation 🤷‍♂ 

The corresponding change was in f4d8f1cf57272ddbb5725596d1e40c21ef22a966.

src/simpleNbox.cpp
```diff
@@ -998,4 +1132,201 @@ void SimpleNbox::set_c0(double newc0)
         
 }
 
+// Add a biome to a time-series map variable (e.g. veg_c_tv)
+template <class T_data>
```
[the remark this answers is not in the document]
ashiklom: Good to know, thanks! Addressed in 53de23c293e781409873d68da2e3200b25a80664.

src/simpleNbox.cpp
ashiklom: Good catch! Addressed in 53de23c293e781409873d68da2e3200b25a80664.

src/simpleNbox.cpp
ashiklom: Addressed in dc8f5f0c6186ecfbe51aa1d0b8c97ed34fa54150.

src/simpleNbox.cpp
```diff
@@ -709,9 +838,9 @@ unitval SimpleNbox::npp( std::string biome ) const
 unitval SimpleNbox::sum_npp() const
 {
     unitval total( 0.0, U_PGC_YR );
-    for( unitval_stringmap::const_iterator it = veg_c.begin(); it != veg_c.end(); it++ ) {
-        total = total + npp( it->first );
-    }
+    std::vector<std::string>::const_iterator it;
```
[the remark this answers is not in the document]
ashiklom: Thanks! Using `auto` is much cleaner, and will make it easier to change the type of `biome_list` if we decided to do that later.

Addressed in 1c4bc03aec20de893238fb43e84dfb1c75e46f8b.

src/simpleNbox.cpp
```diff
@@ -408,89 +478,140 @@ unitval SimpleNbox::getData(const std::string& varName,
...
-        else
-            returnval = sum_map(detritus_c_tv.get(date));
-    } else if( varName == D_SOILC ) {
-        if(date == Core::undefinedIndex()) 
-            returnval = sum_map( soil_c );
-        else
-            returnval = sum_map(soil_c_tv.get(date));
-    } else if( varName == D_FFI_EMISSIONS ) {
+    } else if( varNameParsed == D_VEGC ) {
+        if(biome == SNBOX_DEFAULT_BIOME) {
```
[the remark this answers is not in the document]
ashiklom: So the subtle (arguably ill-advised? Definitely insufficiently documented) design choice here is that `global.veg_c` in the case where there are multiple biomes will _still work_, and is a special case that returns the sum of vegetation C pools across all biomes. For instance, if you have `veg_c["default"]` and `veg_c["arctic"]`, then asking for `VEG_C()` (same as `VEG_C("global")`) will return `veg_c["default"] + veg_c["arctic"]`.

This may or may not be good design, and while I think the ability to do this is useful in some cases, I also definitely see the rationale for dropping it.

src/simpleNbox.cpp
```diff
@@ -623,12 +750,14 @@ void SimpleNbox::stashCValues( double t, const double c[] )
...
-        std::string biome = it->first;
+
+    std::vector<std::string>::const_iterator it;
+    for( it = biome_list.begin(); it != biome_list.end(); it++ ) {
+        std::string biome = *it;
         const double wt     = ( npp( biome ) + rh( biome ) ) / npp_rh_total;
-        veg_c[ biome ]      = veg_c[ biome ] + veg_delta * wt;
-        detritus_c[ biome ] = detritus_c[ biome ] + det_delta * wt;
-        soil_c[ biome ]     = soil_c[ biome ] + soil_delta * wt;
+        veg_c[ biome ]      = veg_c.at( biome ) + veg_delta * wt;
```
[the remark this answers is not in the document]
ashiklom: I switched to `at` to ensure that this statement threw an error if `biome` didn't exist, now and at all points in the future (i.e. if the earlier checks in this block were removed). But I'm happy to change it back if there's a performance hit of `.at` vs. `[]`.

Changing this to `+=` fails because of the error `no viable overloaded +=`. I took a stab at implementing the overloaded `+=` operator in `unitval.hpp`, but it's sufficiently different from the binary operators that I wasn't totally sure how to proceed.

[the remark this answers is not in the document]
rplzzz: `at` and `[]` do the same thing; it's just that `[]` creates the entry if it doesn't exist already.  `x[y]` seems more natural to me than `x.at(y)`, if you don't need the existence check, but it's fine either way.  

It would be good to have a `+=` operator for unitval objects.  I might have time to crank that out sometime this afternoon.  If not, then we can just leave it as it is.

src/core.cpp
```diff
@@ -754,4 +767,62 @@ void Core::delcore(int idx)
     // If core is null, it's already been shutdown, so do nothing. 
 }
 
+//! Retrieve the current biome list
+std::vector<std::string> Core::getBiomeList() const
+{
+    IModelComponent* snbox_i = getComponentByName( SIMPLENBOX_COMPONENT_NAME );
+    SimpleNbox* snbox = dynamic_cast<SimpleNbox*>(snbox_i);
```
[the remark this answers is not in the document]
ashiklom: I agree that this is probably the better way to go. However, one noteworthy caveat is that I think we have to introduce code like the following to the `ocean_component.hpp`, which is _also_ a `CarbonCycleModel`:

```
    inline
    void createBiome(const std::string& biome) {
        H_THROW("Cannot call `createBiome` for ocean component.")
    }
    inline
    void deleteBiome(const std::string& biome) {
        H_THROW("Cannot call `deleteBiome` for ocean component.")
    };
    inline
    void renameBiome(const std::string& oldname, const std::string& newname){
        H_THROW("Cannot call `renameBiome` for ocean component.")
    }
```

Otherwise, we get a compile-time error because, when we go to initialize the ocean component, we end up allocating a "virtual" class, which I guess is a bad thing.

Is this an OK workaround, or are there better alternatives? Should I just implement these directly in the `CarbonCycleModel` (modifying the error slightly to say something like `createBiome` not implemented for this model)? I guess we could also consider deepening the class hierarchy such that `SimpleNbox` inherits from something like `LandCCycleModel` that in turn inherits from `CarbonCycleModel`, but that seems like overkill.

src/simpleNbox.cpp
```diff
@@ -143,16 +184,34 @@ void SimpleNbox::setData( const std::string &varName,
...
-            H_ASSERT( data.date == Core::undefinedIndex(), "date not allowed" );
+            // For `veg_c`, `detritus_c`, and `soil_c`, if date is not
+            // provided, set only the "current" model pool, without
+            // touching the time series variable. This is to
+            // accommodate the way the INI file is parsed. For
+            // interactive use, you will usually want to pass the date
+            // -- otherwise, the current value will be overridden by a
+            // `reset` (which includes code like `veg_c = veg_c_tv.get(t)`).
+            //
+            // TODO: Should we create special variables `veg_c0`,
```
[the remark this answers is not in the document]
ashiklom: Converted to an issue in case we want to discuss further: https://github.com/JGCRI/hector/issues/318

Removed the comment in 10d3ac2f353bc946bcfc5560802aff6a67f9f9e6.

src/simpleNbox.cpp
ashiklom: Conversation moved to #318.

src/simpleNbox.cpp
```diff
@@ -880,14 +1010,16 @@ void SimpleNbox::slowparameval( double t, const double c[] )
     Ca.set( c[ SNBOX_ATMOS ] * PGC_TO_PPMVCO2, U_PPMV_CO2 );
 
     // Compute CO2 fertilization factor globally (and for each biome specified)
-    double_stringmap::const_iterator itd;
-    for( itd = beta.begin(); itd != beta.end(); itd++ ) {
+    std::vector<std::string>::const_iterator it;
+    std::string biome;
```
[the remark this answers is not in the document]
ashiklom: Addressed in 1c4bc03aec20de893238fb43e84dfb1c75e46f8b.

src/core.cpp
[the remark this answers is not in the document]
rplzzz: Hmm.  I hadn't realized that the ocean component was also a `CarbonCycleModel`.  If I were going to add a new class, I'd do it using multiple inheritance rather than deepening the hierarchy.  But maybe the best course for now is just to have those methods throw a not-implemented error in the ocean component.

> Should I just implement these directly in the CarbonCycleModel (modifying the error slightly to say something like createBiome not implemented for this model)?
 
Yes, this is the best way to implement a design where the subclasses are allowed to not implement part of the interface.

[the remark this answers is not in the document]
ashiklom: OK, I implemented this in 87665ee8b3e8169e1b23dd46c068b9d2cf241199.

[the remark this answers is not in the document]
rplzzz: 💯

src/core.cpp
```diff
@@ -754,4 +767,62 @@ void Core::delcore(int idx)
     // If core is null, it's already been shutdown, so do nothing. 
 }
 
+//! Retrieve the current biome list
+std::vector<std::string> Core::getBiomeList() const
+{
+    IModelComponent* snbox_i = getComponentByName( SIMPLENBOX_COMPONENT_NAME );
```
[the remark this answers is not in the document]
ashiklom: Good point. I changed this to `getComponentByCapability(D_VEGC)` in 133497a0216e6eb4be7aaf162041a998c3ba27ae.

I considered adding a new `"biomes"` capability, but that seemed like overkill. I can't imagine a situation where a C cycle model would use biomes and _not_ have a vegetation carbon term.

ashiklom:
@bpbond @rplzzz I think I addressed (one way or another) all of your comments.

> Check the comments in the getData method to make sure that this is working as intended, and delete special case code if appropriate.

This was the only one I didn't change -- see [my explanation here](https://github.com/JGCRI/hector/pull/317#discussion_r301660619). As I said there, I think there is some utility of leaving it as is, but I'm also fine changing it as suggested.

The only other unresolved item is the issue of the `+=` operator discussed [here](https://github.com/JGCRI/hector/pull/317#discussion_r301714043). Since this PR is already pretty huge, I would probably punt that to a dedicated future PR, but I'm happy to include it here if necessary.

I'll bump the version number once all the other changes have been approved, to make the git commit history clear about which changes are included in this version.

src/core.cpp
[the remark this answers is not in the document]
rplzzz: Sounds good.  That's the same strategy we applied in the dump-to-deep-ocean message.

src/simpleNbox.cpp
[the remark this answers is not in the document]
rplzzz: Will it work that way?  I thought that earlier in that method we check to see if the asked-for biome exists, and we throw an error if it doesn't (I'm going from memory here, but I think that's how it worked).  If so, then we _can't_ ask for `VEG_C("global")` when multi-biome is in effect because you aren't allowed to have a global biome when you have other biomes.

But if I'm mistaken, and we do let "global" through in the multi-biome case, then I'm ok with it as is.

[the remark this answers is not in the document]
ashiklom: Good catch!

Interestingly, what I had was a situation where `VEG_C()` (`"veg_c"`) would work as expected, but `VEG_C("global")` (`"global.veg_c"`) would not, because the `has_biome` assertion was only run if the split character was detected.

However, having that assertion there was redundant because we run it _again_ before every biome-specific parameter retrieval. So I removed it in 58469cb7875c12ce92d832974b6d02069a05ee44 and added a quick test to make sure that both cases work as expected.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: hector
 Title: The Hector Simple Climate Model
-Version: 2.3.0
+Version: 2.3.1
```
rplzzz:
```suggestion
Version: 2.4.0
```

The version string will also need to be changed in `h_util.hpp`.

ashiklom: Resolved in 1f4bbc4.

DESCRIPTION
```diff
@@ -26,6 +26,7 @@ Suggests:
 Collate:
     'aadoc.R'
     'RcppExports.R'
+    'biome.R'
```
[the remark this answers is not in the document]
ashiklom: Resolved in 1f4bbc4.

rplzzz:
I'm satisfied.  Nice work, Alexey.

ashiklom:
Thanks! @bpbond Your change request is still blocking this -- let me know if you'd like to see any other changes or if this is ready to merge.

PR #319, opened 2019-07-30
Expose npp

kdorheim:


kdorheim:
I've added two tests let me know if you think if you think we need more.

rplzzz:
@kdorheim The tests look good to me.  Go ahead and select a reviewer so we can get this merged.

rplzzz:
@kdorheim The tests look good to me.  I'll select a reviewer.

PR #320, opened 2019-08-08
Biome-specific partitioning fractions and Q10

ashiklom:
This adds biome-specific versions of the following terrestrial C cycle variables:
- Heterotrophic respiration temperature Q10 (`q10_rh`). (NOTE: We claim to be able to do this already, but we actually can't until this change.)
- Fraction of NPP to vegetation (`f_nppv`) and detritus (`f_nppd`)
- Fraction of litter flux to detritus (`f_litterd`)

I bumped the version 2.4.0 -> 2.4.1, but this is technically a breaking change to multi-biome runs configured entirely via the INI file because the code will error if `f_nppv` and `f_nppd` are not set for each biome. Default, global runs are unaffected, and I'm probably the only person doing anything with multiple biomes right now, but I'm open to making this a bigger version bump or adding some more back-compatibility (e.g. assign defaults for missing biome-specific parameters, like we do for `warmingfactor`).


rplzzz:
I would call this a minor version change.  It's _technically_ a breaking change, but arguably those multi-biome files that parsed in the past never should have.  So, I'd call it adding new capabilities, plus fixing a bug that was causing input files that should have produced errors to parse successfully.  So, 2.5.0?

ashiklom:
@bpbond Thanks!

@rplzzz Did you want to take a closer look at this, or should I proceed with the merge?

rplzzz:
@ashiklom I won't be able to get to it until next week, so it depends on if you mind waiting that long.

ashiklom:
Did anyone else want to have a look at this, or can I merge it in?

PR #336, opened 2019-12-18
Expose and test more emissions inputs and capabilities

ashiklom:
Basically, this allows more variables that should be settable via `setvar` and gettable via `fetchvars`. I've also added a test that runs through every emissions variable and makes sure it can be set and retrieved.


ashiklom:
FYI -- this is currently failing on AppVeyor because of an [upstream problem](https://github.com/eddelbuettel/bh/issues/67) with a slightly outdated version of the BH package. The problem has been resolved and pushed to CRAN ([1.72.0-1](https://cran.r-project.org/web/packages/BH/ChangeLog), 2019-12-16), but I don't think the change has propagated over to wherever AppVeyor gets its packages from yet.

bpbond:
Nice new test. 👍

Issue #330, opened 2019-11-13
Enable CSV output stream from R package (or a helper to fetch all outputs)


ashiklom:
I can imagine plenty of scenarios where one might want to do a bunch of Hector simulations (e.g. for a sensitivity or uncertainty analysis) without knowing the full list of desired output variables in advance. For these situations, the old Hector capability of dumping all of its output variables to a CSV file would be really useful. However, as far as I can tell, there isn't a way of turning on the CSV output stream from the R interface or INI file.

Off the top of my head, I see two possibilities:

(1) An R helper function (or set of helper functions) containing all of the variables Hector knows how to provide (e.g. `all_variables`), and/or the ability to retrieve all of the variables via a separate function `fetchvars_all` or a special argument like `all` to `fetchvars`. This should be pretty straightforward to implement, but will only work for variables that the R interface knows about, which is currently not all of them (which may be its own separate issue? I feel like the R interface should be able to retrieve any Hector variable...).

(2) The ability to enable/specify/configure the CSV output stream via `sendMessage` (which in turn would allow this to be modified by the INI file or the R interface). This solution is more difficult to implement, but is also more versatile (e.g. works with R or the INI file, and to the extent that pyhector uses `sendMessage` calls, could be easily integrated there as well).

rplzzz:
Regarding option 1, `fetchvars` already takes a vector of variables to return, so all we need is a function that returns a list of everything that's available.  That's pretty easy to add and equally easy to use, so I don't think we need another function.  If we want to support the fetch everything use case, this is the best way to do it.

Personally, I really don't like option 2, as it presses Hector's internal messaging mechanism into service to send messages to something (the CSV output visitor, presumably) that isn't a model component.  In my opinion, it's better to keep the messaging system simple and clean, particularly since it already provides a way to retrieve output from the model.

[...]

PR #337, opened 2019-12-19
Enable CSV output stream from R package

mnichol3:
Addresses #330 by adding the function `fetchvars_all` and 3 helper functions. Similar to `fetchvars`, this function retrieves all available variables by searching through the list of available Hector library functions for variable capability names and then passes them to `sendmessage` to retrieve their model output values. The resulting data frame can then be written to csv if desired. 

RStudio changed a lot of whitespace and I tried to revert those commits with no luck. Any advice is appreciated


rplzzz:
I guess we decided on this approach, instead of the simpler tack of creating a package function (`ALL_HECTOR_VARS()` or something like that) that listed all the available variables, so as to make it so that when we add new variables we don't have to rely on developers remembering to add them to the all-variables list.  However, I wonder if in doing so we have introduced some brittleness into the design.  

Now we have to ensure that newly added variables conform to one of several regular expressions for variable names.  In some cases, such as `"^OCEAN_C(_\\w{2})?$|^ATM_OCEAN_+|\\w{1,5}_[HL]L$"` the convention is pretty restrictive ("OCEAN_C" may be followed by an underscore and exactly two word constituent characters, or by none at all, but not by, say "_X").  We also have to make sure that we don't add any other kind of functions that matches one of these patterns.  The capital letters make this somewhat unlikely, but Murphy's Law never sleeps.

This patch is impressive in its cleverness at tracking down the variables to fetch, but I wonder if it might not better to go with the simpler, yet more robust, solution just enumerating what we mean by "all" variables.

R/messages.R
```diff
@@ -110,3 +110,229 @@ setvar <- function(core, dates, var, values, unit)
...
+    # Get variables from the helper functions, combine results into one dataframe
+    vars_inpt  <- cum_vars_input(core, dates, lib_funcs)
+    vars_param <- cum_vars_params(core)
+    vars_outpt <- cum_vars_output(core, dates, lib_funcs)
+
+    vars_all <- rbind(vars_inpt, vars_param, vars_outpt)
+
+    # If 'write' is TRUE, reshape the resulting data frame to a wide format
+    # and write to csv
+    if (write) {
```
rplzzz: Instead of a logical, consider making write a string with a default value of `NA`.  `NA` means don't write the output do a file; a non-NA value means write to the indicated filename.

R/messages.R
```diff
@@ -110,3 +110,229 @@ setvar <- function(core, dates, var, values, unit)
...
+
+        # Set the column names for the re-ordered dataframe
+        colnames(vars_all) <- col_names
+
+        # Check if the absolute path of the output csv file is valid
+        if (!is.null(outpath)) {
+            write.table(vars_all, outpath, col.names=col_names, row.names=F, sep=',')
+            cat("Output written to ", outpath, sep="")
+        } else {
+            # Write to a default location in the Hector R library output directory
```
rplzzz: I am very leery of writing stuff into R's package directories outside of the package install mechanism.  If nothing else, we don't even know for sure that the user has write privilege there.  It also puts us in the position of having "installed" files that the package doesn't know anything about.  If we need a default output path, we should put it in the current working directory (better still, make specifying the output file the thing that turns on the file output)

R/messages.R
bpbond: Agreed.

src/core.cpp
```diff
@@ -371,7 +371,7 @@ bool Core::run_spinup()
         spunup = true;
         for( NameComponentIterator it = modelComponents.begin(); it != modelComponents.end(); ++it )
             spunup = spunup && ( *it ).second->run_spinup( step );
-        
+
```
bpbond: It's going to be a big drag if every PR now has tab/spacing wars with lots of changes like this. Ideally please make sure your RStudio settings match others'.

mnichol3: I think this was due to RStudio removing leftover whitespace on save rather than the classic tab/space schism. I believe I fixed it in Rstudio's setting but I'm not sure how to undo the changes to the files.

mnichol3:
> I guess we decided on this approach, instead of the simpler tack of creating a package function (`ALL_HECTOR_VARS()` or something like that) that listed all the available variables, so as to make it so that when we add new variables we don't have to rely on developers remembering to add them to the all-variables list. However, I wonder if in doing so we have introduced some brittleness into the design.
> 
> Now we have to ensure that newly added variables conform to one of several regular expressions for variable names. In some cases, such as `"^OCEAN_C(_\\w{2})?$|^ATM_OCEAN_+|\\w{1,5}_[HL]L$"` the convention is pretty restrictive ("OCEAN_C" may be followed by an underscore and exactly two word constituent characters, or by none at all, but not by, say "_X"). We also have to make sure that we don't add any other kind of functions that matches one of these patterns. The capital letters make this somewhat unlikely, but Murphy's Law never sleeps.
> 
> This patch is impressive in its cleverness at tracking down the variables to fetch, but I wonder if it might not better to go with the simpler, yet more robust, solution just enumerating what we mean by "all" variables.

My 'Plan B' solution was to have a .csv file in `/R`, or perhaps `/inst/input`, containing a single column of variable names that would be read by the function and possibly validated by comparing these variables to the list of Hector's variables from `lsf.str("package:hector")`. It would be slightly more maintenance intensive when it comes to adding new variables, but would be easier than the regex to exclude certain variables users don't want. It would also have a smaller chance of breaking (unless the csv gets nuked, but that's an easy fix)/having unintended side effects down the road. Thoughts on this solution?

R/messages.R
ashiklom: Agree with the general idea, but I usually prefer `NULL` to `NA` for missing values as default arguments. `NA` can be of any `type`, but `NULL` is its own, very specific type. Moreover, `is.null()` is always a scalar boolean so it's guaranteed to play nice with `if` statements, whereas a vector of `NA` values is perfectly valid and therefore `is.na()` is vectorized (which becomes problematic in the context of `if` statements).

See also: [NA vs. NULL](https://www.r-bloggers.com/r-na-vs-null/) and [this Stack Overflow discussion](https://stackoverflow.com/questions/29619431/should-i-use-null-or-na-as-default-values-for-optional-function-parameters).

In practice, the `NULL` vs. `NA` distinction is pretty pedantic -- either one is fine, as long as we are fairly consistent throughout Hector.

rplzzz:
> 
> My 'Plan B' solution was to have a .csv file in `/R`, or perhaps `/inst/input`, containing a single column of variable names that would be read by the function and possibly validated by comparing these variables to the list of Hector's variables from `lsf.str("package:hector")`. It would be slightly more maintenance intensive when it comes to adding new variables, but would be easier than the regex to exclude certain variables users don't want. It would also have a smaller chance of breaking (unless the csv gets nuked, but that's an easy fix)/having unintended side effects down the road. Thoughts on this solution?

You can't put a csv file in the `R` subdir directly, though you could create the table using an appropriate script and use `dput` to create a block of R code to recreate the structure.  It would look something like this:
```
structure(list(name = structure(1:3, .Label = c("a", "b", "c"
), class = "factor"), value = c(1, 2, 3)), class = "data.frame", row.names = c(NA, 
-3L))
```
And you could assign that to a package variable in a .R file in the `R` subdir.  Alternatively, you could use `usethis::use_data` to install the table as package data in the `data` subdir.  The latter is cleaner, but less transparent (since the files in `data` are binary).

I don't recommend installing the csv in `inst`, since to make that work you would have to include some package-startup shenanigans to get the table loaded for use.  Better just to arrange for the table to be part of the package from the start.

src/core.cpp
ashiklom: Personally, I would prefer to keep this version of the files -- it's good practice to trim trailing whitespace from our code. As far as project consistency, I think this is the whole point of the project-specific [`hector.Rproj`](https://github.com/JGCRI/hector/blob/master/hector.Rproj) file, which already specifies the removal of trailing whitespace, the enumber of spaces for indenting, etc.

I think the issue is that many of these C++ source files predate the Hector R package and therefore did not have these settings specified. @mnichol3 just happened to be the first person to open and re-save them since then.

For the contrarians among us who don't use RStudio (I think just me? Emacs!), perhaps we should also create a Hector [EditorConfig](https://editorconfig.org/) file to codify these settings?

ashiklom:
> Alternatively, you could use usethis::use_data to install the table as package data in the data subdir. The latter is cleaner, but less transparent (since the files in data are binary).

The `data` directory actually supports not just binary `.Rdata` files, but also plain-text tables as `.txt` and `.csv` as well as plain-old `.R` scripts ([source](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Data-in-packages); see also [`?data`](https://www.rdocumentation.org/packages/utils/versions/3.6.2/topics/data)). That, combined with adding `LazyData: True` to the `DESCRIPTION` file, should provide for a nice interface while still having plain-text, easily VC-able and editable files for developers.

(Note that the CSV file format R assumes for the `data` directory is, annoyingly, not quite standard because it assumes a `;` separator. That will prevent it from being correctly rendered by GitHub and other places. A better option may be a tab-separated `.txt` or `.tab` file).

rplzzz:
> The `data` directory actually supports not just binary `.Rdata` files, but also plain-text tables as `.txt` and `.csv` as well as plain-old `.R` scripts ([source](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Data-in-packages); see also [`?data`](https://www.rdocumentation.org/packages/utils/versions/3.6.2/topics/data)). That, combined with adding `LazyData: True` to the `DESCRIPTION` file, should provide for a nice interface while still having plain-text, easily VC-able and editable files for developers.
> 
> (Note that the CSV file format R assumes for the `data` directory is, annoyingly, not quite standard because it assumes a `;` separator. That will prevent it from being correctly rendered by GitHub and other places. A better option may be a tab-separated `.txt` or `.tab` file).

Good to know.  It appears that csv and tsv files in the data directory are read in with `as.is=FALSE`, so if we go with one of those, we should also double check that having all the strings converted to factors won't cause any problems.

Also, I don't think there is any reason to have this table exposed as part of the public interface, so if we store it in the data directory we should look into storing it as internal data.

PR #341, opened 2020-01-15
Store and retrieve atmosphere-land and atmosphere-ocean C flux

### LINKED ISSUE #328: Retrieve atmosphere-land and atmosphere-ocean flux time series
Currently, neither `LAND_CFLUX` nor `OCEAN_CFLUX` accept date values -- only the values for the current time step. I have a personal stake in this because "total C sequestration" is tier 1 variable for RCMIP. However, both of these seem like useful diagnostic variables and could be good targets for calibrating Hector's C cycle to CMIP models, etc.

I think the following changes (on a branch of my Hector fork) accomplish this, but I may be missing something:

https://github.com/ashiklom/hector/compare/concentration-constraint..track-cfluxes

ashiklom:
- Adds a new time-series variable to `simpleNbox` -- `atmosland_flux_ts` -- that tracks the atmosphere -> land C flux. Update this variable every time `simpleNbox::stashCValues` is called.
- Require `D_LAND_CFLUX()` and ``D_OCEAN_CFLUX()` to take dates in `getData`. This is a (small) **breaking change**.
- New unit test checks that these fluxes can be retrieved, and return reasonable values.

Closes #328.


ashiklom:
@bpbond Thanks for the review!

(1) I added `NPB()` as an R alias for `LAND_CFLUX()`. I think replacing `LAND_CFLUX()` altogether would be a but too much of a backward incompatible change relative to the gain. For a future major Hector version (3.0?), it may be worth revisiting existing variable names to make them more consistent with the literature and with each other.

(2) I added a note to the SimpleNBox documentation, and a comment in the source code, about fire.

bpbond:
@ashiklom That seems like a sensible compromise, thanks. Note however that it's NBP not NPB.

ashiklom:
Oops. That's what I meant, and is actually what's in the code.

src/rcpp_constants.cpp
```diff
@@ -839,6 +839,14 @@ String LAND_CFLUX() {
...
 }
 
+// An alias for LAND_CFLUX that may be more familiar to carbon cycle people
+//' @rdname carboncycle
+//' @export
+// [[Rcpp::export]]
+String NBP() {
+return D_LAND_CFLUX;
+}
+
```
kdorheim: I am slightly confused by the NBP terminology. So NBP = Land_C_Flux but this term is more familiar to the carbon cycle community. However because Hector has multiple biomes or will have multiple biomes does NBP reflect the global land C flux or is it biome specific? Are we opposed to spelling out net biome productivity here in the comments? Am I correct in understanding that both NBP() and LAND_CFLUX() will work as functions in R and return the same string?

tests/testthat/test_hector.R
```diff
@@ -228,3 +228,20 @@ test_that('Test RF output.', {
...
+                  suppresslogging = TRUE)
+    run(hc, 2100)
+
+    yrs <- 1750:2100
+    out_land <- fetchvars(hc, dates = yrs, LAND_CFLUX(), "Pg C year-1")
+    out_nbp <- fetchvars(hc, dates = yrs, NBP(), "Pg C year-1")
+    expect_identical(out_land, out_nbp)
+    out_ocean <- fetchvars(hc, dates = yrs, OCEAN_CFLUX(), "Pg C year-1")
+
+    # After 1960, land is consistently a C sinks
```
rplzzz:
```suggestion
    # After 1960, land is consistently a C sink
```

vignettes/manual/SimpleNBox.Rmd
```diff
@@ -12,10 +14,13 @@ Contents:
 Hector's land C sink is divided into three pools: vegetation ($C_v$), detritus ($C_d$), and soil ($C_s$).
 These sinks are global by default, but they can be divided into $n$ "biomes" with unique parameters and initial pool sizes.
 
-The total C flux from atmosphere to land at time $t$ ($F_{L}(t)$) is the net of C sequestration by photosynthesis (net primary productivity, $NPP_i(t)$) and C release by heterotrophic respiration ($RH_i(t)$) summed across each biome $i$, and global land use change ($LUC(t)$):
+The total C flux from atmosphere to land at time $t$ ($F_{L}(t)$; also known as Net Biome Production, NBP) is the net of C sequestration by photosynthesis (net primary productivity, $NPP_i(t)$) and C release by heterotrophic respiration ($RH_i(t)$) summed across each biome $i$, and global land use change ($LUC(t)$):
 
 $$ F_{L}(t) = \sum_{i=1}^{n} \left( NPP_i(t) - RH_i(t) \right) - LUC(t)  $$
 
+(Another significant component of land-atmosphere C exchange is emissions from fire.
```
rplzzz: If it were me, I'd leave this out of this section, as a piece of our development roadmap is not really what users are looking for here, nor is this the place where users would generally go to look for information on future development.    We could, however,  just mention that fires are not included, if we think that users might expect it to be.  If we do that, we should mention it in the paragraph above, not as a parenthetical down here.

ashiklom: Addressed in 0d4ed96. @bpbond You were the one that originally requested I mention fire -- let me know if it looks acceptable to you.

src/rcpp_constants.cpp
```diff
@@ -846,6 +846,14 @@ String LAND_CFLUX() {
 return D_LAND_CFLUX;
 }
 
+// An alias for LAND_CFLUX that may be more familiar to carbon cycle people
+//' @rdname carboncycle
```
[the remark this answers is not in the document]
ashiklom: Addressed in 0d4ed96.

src/rcpp_constants.cpp
bpbond: >Am I correct in understanding that both NBP() and LAND_CFLUX() will work as functions in R and return the same string?

Yes as far as I see. C cycle terminology:
* NEE: exchange between land and atmosphere measured in a single place, negative = uptake
* NECB: full carbon balance, so NEE + a bunch of other things not measured by towers
* NBP: carbon balance over a big area, so including things like fire and LUC

Canonical reference is [Chapin et al. 2006](http://dx.doi.org/10.1007/s10021-005-0105-7).

vignettes/manual/SimpleNBox.Rmd
bpbond: That's a good point–normally NBP would include fire. This looks fine for now, thanks.

rplzzz:
Thanks, @ashiklom; everything looks good now.

vignettes/manual/SimpleNBox.Rmd
cahartin: @bpbond is fire coming into Hector in the form of LUC emissions?

bpbond: No; it's not there at all currently afaik.

PR #342, opened 2020-01-15
Prescribe atmospheric CO2, v2

ashiklom:
This PR is an updated version of #302. I rebased that PR's commits onto the `master` branch, and resolved the [problem](https://github.com/JGCRI/hector/pull/302#issuecomment-493242815) that @kdorheim was having. The specific fix was to (1) only apply `Ca_constraint` when `Ca_constrain.exists(time)` (previously, it was just less than `lastdate`), and (2) set `Ca_constrain.allowPartialInterp(true)` during `setData` (because the usual place where this is enabled -- `prepareToRun` -- is not called by `run`).


src/simpleNbox.cpp
```diff
@@ -261,6 +263,11 @@ void SimpleNbox::setData( const std::string &varName,
             H_ASSERT( data.date != Core::undefinedIndex(), "date required" );
             H_ASSERT( biome == SNBOX_DEFAULT_BIOME, "atmospheric constraint must be global" );
             Ca_constrain.set( data.date, data.getUnitval( U_PPMV_CO2 ) );
+            // This is necessary because the usual place where this is set --
+            // `prepareToRun` -- is not called by `run`, so a sequence like
+            // `newcore` -> `setvar` -> `run` will result in `Ca_constrain` that
+            // cannot use interpolation.
```
bpbond: Great comment 👍

vignettes/manual/Constraints.Rmd
```diff
@@ -39,3 +39,6 @@ Ca_constrain=csv:path/to/constraint/file.csv
 then the model's atmospheric carbon pool (given in ppmv CO2) will follow this contents of this file (which should have two columns, `Date` and `Ca_constrain`). The model will extrapolate within this data as necessary, but not beyond its endpoints. If the constraint [CO2] is lower than the calculated [CO2] at that time step, any excess is moved to the deep ocean; conversely, any necessary deficit is transferred from there.
 
 Once Hector's internal date passes the last date in the constraint file, the model's atmospheric CO2 becomes *unconstrained*.
```
bpbond: Maybe note that discontinuous constraints are OK.

src/simpleNbox.cpp
ashiklom: Thanks!

vignettes/manual/Constraints.Rmd
ashiklom: Revised in c2fdb83.

tests/testthat/test_parameters.R
```diff
@@ -256,3 +256,87 @@ test_that('All "fraction" parameters can be set and retrieved', {
...
+                    out_ca[out_ca$year == 1900, "value"])
+
+    # You should be able to retrieve constraints that were set
+    expect_equal(fetchvars(hc, min(ca_years_1), CA_CONSTRAIN())$value, ca_vals_1[1])
+    expect_equal(fetchvars(hc, max(ca_years_1), CA_CONSTRAIN())$value, ca_vals_1[1])
+    expect_equal(fetchvars(hc, min(ca_years_2), CA_CONSTRAIN())$value, ca_vals_2[1])
+    expect_equal(fetchvars(hc, max(ca_years_2), CA_CONSTRAIN())$value, ca_vals_2[1])
+
+    err <- "No CO2 constraint for selected date"
+    # You should NOT be able to retrieve constraints before...
```
rplzzz: As noted above, this is kind of a user-unfriendly behavior.  It would be much better if we could contrive a way to return `NA` in these cases.

vignettes/manual/Constraints.Rmd
```diff
@@ -36,6 +36,9 @@ then the model's ocean carbon uptake (in Pg C) will follow this contents of this
 Ca_constrain=csv:path/to/constraint/file.csv
 ```
 
-then the model's atmospheric carbon pool (given in ppmv CO2) will follow this contents of this file (which should have two columns, `Date` and `Ca_constrain`). The model will extrapolate within this data as necessary, but not beyond its endpoints. If the constraint [CO2] is lower than the calculated [CO2] at that time step, any excess is moved to the deep ocean; conversely, any necessary deficit is transferred from there.
+then the model's atmospheric carbon pool (given in ppmv CO2) will follow this contents of this file (which should have two columns, `Date` and `Ca_constrain`). If the constraint [CO2] is lower than the calculated [CO2] at that time step, any excess is moved to the deep ocean; conversely, any necessary deficit is transferred from there.
```
rplzzz: This is not correct.  The atmospheric carbon pool is different from the atmospheric CO<sub>2</sub> concentration.  The latter is what we are actually constraining.

vignettes/manual/Constraints.Rmd
```diff
@@ -36,6 +36,9 @@ then the model's ocean carbon uptake (in Pg C) will follow this contents of this
...
+Otherwise Hector's atmospheric carbon pool can be set using the `setvars` function and `CA_CONSTRAIN` variable.
```
rplzzz: Same here.

R/aadoc.R
```diff
@@ -63,6 +63,7 @@ NULL
 #' \describe{
 #' \item{FFI_EMISSIONS}{Fossil fuel and industrial emissions (\code{"Pg C/yr"})}
 #' \item{LUC_EMISSIONS}{Land use change emissions (\code{"Pg C/yr"})}
+#' \item{CA_CONSTRAIN}{Prescribed atmospheric carbon pool (\code{"ppmv CO2"})}
```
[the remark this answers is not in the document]
ashiklom: Good catch. I'll revise the documentation accordingly. 

Should we also change the variable to `CO2_CONSTRAIN()` (and `CO2_constrain` in the C++ code) for clarity (and consistency with how I've implemented the other constraints in #347)?

src/simpleNbox.cpp
```diff
@@ -261,6 +263,11 @@ void SimpleNbox::setData( const std::string &varName,
...
+            Ca_constrain.allowPartialInterp( true );
```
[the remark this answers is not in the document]
ashiklom: At one point, I kept hitting errors where the code would try to retrieve interpolated values for non-integer years, so I assumed something about the sub-annual timestep of the C cycle required this interpolation. But I just tried disabling this and all my tests still passed, so I guess it's not necessary after all. Good catch!

Addressed in 4638fde.

src/simpleNbox.cpp
```diff
@@ -595,6 +602,10 @@ unitval SimpleNbox::getData(const std::string& varName,
     } else if( varNameParsed == D_LUC_EMISSIONS ) {
         H_ASSERT( date != Core::undefinedIndex(), "Date required for luc emissions" );
         returnval = lucEmissions.get( date );
+    } else if( varNameParsed == D_CA_CONSTRAIN ) {
+        H_ASSERT( date != Core::undefinedIndex(), "Date required for atmospheric CO2 constraint" );
+        H_ASSERT( Ca_constrain.exists(date), "No CO2 constraint for selected date" );
```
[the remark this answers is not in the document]
ashiklom: > Is there a way we could return a sentinel value that the R interface could translate to an NA value?

Is `NaN` OK? Or, more formally, how important is it that we maintain the distinction between `NaN` (the result of an illegal math operation, like dividing by zero) and `NA` (a true missing value)? Blurring the distinction makes this easy -- it's trivial to set `std::NAN` as the return value, and R has native support for those. Preserving the distinction would be more challenging (though could ultimately be rewarding down the line), since it would require more substantial modifications of `sendmessage` and/or `unitval`.

My proposal would be to stick with `NaN` for now and discuss implementing a true missing value class later on.

tests/testthat/test_parameters.R
```diff
@@ -256,3 +256,87 @@ test_that('All "fraction" parameters can be set and retrieved', {
...
+test_that('Atmospheric CO2 constraint affects RF and temperature', {
+
+    # Checks for the regression reoported here:
+    # https://github.com/JGCRI/hector/pull/302#issuecomment-493242815
+    hc <- newcore(rcp45, suppresslogging = TRUE)
+    ca_years <- 1850:2100
+    ca_low <- rep(278, length(ca_years))
+    setvar(hc, ca_years, CA_CONSTRAIN(), ca_low, getunits(CA_CONSTRAIN()))
+    invisible(run(hc))
+    out_ca_lo <- fetchvars(hc, ca_years, CA_CONSTRAIN())
```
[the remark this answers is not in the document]
ashiklom: Good catch. Addressed in 2b24788.

vignettes/manual/Constraints.Rmd
ashiklom: Good catch. Addressed in c841d8f.

vignettes/manual/Constraints.Rmd
ashiklom: Blegh -- good idea to read all comments before committing. Addressed in 88a78e9.

src/simpleNbox.cpp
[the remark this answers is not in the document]
ashiklom: Actually, I think I found a workaround (4d348ba): If Rcpp is available, we can use its `NA_REAL`, which R (and compatible Rcpp code) understands to be `NA` but other C++ code will treat as `NAN`. At least for the R package, this lets us have our cake and eat it too, without (AFAICT) any major drawbacks.

tests/testthat/test_parameters.R
ashiklom: Addressed in 4d348ba.

vignettes/manual/Constraints.Rmd
rplzzz: I feel your pain.  I just submitted the same set of revisions 4 (!) times because I kept finding minor mistakes and typos right at the final review stage.

R/aadoc.R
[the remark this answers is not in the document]
rplzzz: That's a good idea, especially since this set of capabilities hasn't been out in the wild under the old name yet.

src/simpleNbox.cpp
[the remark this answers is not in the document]
rplzzz: Sounds great.  I was going to say return a `NaN` and have the R code change it to `NA`, but this is even better.

inst/include/unitval.hpp
```diff
@@ -19,6 +19,19 @@
...
+// ("not-a-number"), which is conceptually different (the result of an illegal
+// math operation, like the square root of a negative number) but tends to have
+// the same practical effect on calculations (in fact, R's NA is a special case
+// of NaN).
+#ifdef USE_RCPP
+#include <Rcpp.h>
+#define MISSING_FLOAT NA_REAL
+#else
+#define MISSING_FLOAT NAN
+#endif
```
rplzzz: Smooth!

src/simpleNbox.cpp
```diff
@@ -595,6 +596,15 @@ unitval SimpleNbox::getData(const std::string& varName,
     } else if( varNameParsed == D_LUC_EMISSIONS ) {
         H_ASSERT( date != Core::undefinedIndex(), "Date required for luc emissions" );
         returnval = lucEmissions.get( date );
+    } else if( varNameParsed == D_CA_CONSTRAIN ) {
+        H_ASSERT( date != Core::undefinedIndex(), "Date required for atmospheric CO2 constraint" );
+        if (Ca_constrain.exists(date)) {
+            returnval = Ca_constrain.get( date );
+        } else {
+            H_LOG( logger, Logger::WARNING ) << "No CO2 constraint for requested date " << date <<
```
rplzzz: If it were me, I'd log this at no higher than `NOTICE` (and I'd probably opt for `DEBUG`).  The reason why is that as things stand, the easiest way to figure out what dates have constraints is to just ask for all dates and check to see which ones do or don't come back with `NA`.  That's a legitimate use case that shouldn't generate a bunch of warnings (which, to me mean "probably wrong, but not necessarily or fatally so").  

Alternatively, we could implement a function in the API to check what dates have valid constraints, but I don't think that will necessarily be any easier for users (particularly R users, for whom it's pretty natural to pass big blocks of dates and drop `NA`s).  Plus, that will get kind of ugly when we add all of the other constraint options in #347 .

vignettes/manual/Unitvals.Rmd
```diff
@@ -53,3 +53,18 @@ if( varName == D_PREINDUSTRIAL_N2O ) {`
...
+Missing values
+--------------
+Briefly, when you need to return a missing value, use the `MISSING_FLOAT` macro defined in `inst/include/unitval.hpp`.
+This macro will return `NA_REAL` (from `Rcpp`) if Hector is compiled as an R package (technically, if the `USE_RCPP` macro is defined) and `NAN` (from `std`) otherwise.
+
+Why this complexity?
+R (and, by extension, Rcpp) distinguish between two kinds of "placeholder" numeric values:
+A _missing_ value (`NA_real_` in R; often abbreviated to `NA`) represents an unobserved value; for instance, if you take annual observations from 1990 to 2000 but failed to make a measurement in 1996.
+On the other hand, _not a number_ (`NaN`) represents the result of an invalid mathematical operation, such as taking the square root of a negative number.
+Both of these values propagate through any calculations in which they are involved, both in R and C++, and both can be detected and removed by R's `is.na()` function and removed by `na.omit`, `na.rm = TRUE`, and friends.
```
rplzzz: Note also that `is.na()` returns true for both `NA` and `NaN` values, while `is.nan()` returns true for `NaN` values, but false for `NA`.

R/aadoc.R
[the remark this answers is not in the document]
ashiklom: Addressed in 83aeeab.

src/simpleNbox.cpp
ashiklom: Changed to `Logger::NOTICE` in 80eb6b7.

Re: API for missing constraints -- I agree that it's probably easiest just to check for `NA` on the R side.

vignettes/manual/Unitvals.Rmd
ashiklom: Added a note to this effect in 919fa30.

ssmithClimate:
@ashiklom For documentation purposes, what happens if a constraint is set only up to a certain year? Are concentrations discontinuous after that, or will concentrations smoothly change to being calculated values?

rplzzz:
> @ashiklom For documentation purposes, what happens if a constraint is set only up to a certain year? Are concentrations discontinuous after that, or will concentrations smoothly change to being calculated values?

The way I read the code, when a constraint ends the system will continue evolving from its last constrained state with no discontinuity.  On the other hand, if you run unconstrained for a while and then start a constraint, the system will be abruptly forced into the constraint, possibly with a large step if the last unconstrained state was very inconsistent with the new constraint.

bpbond:
>The way I read the code

Right. Obviously we _could_ have a transition period entering a constraint, if we cared about smoothness above all, but that seems unnecessarily complicated imo.

ashiklom:
@ssmithClimate I added a note clarifying this in the documentation in 2fb39f2.

@rplzzz You still have this marked as "Changes requested". Was there anything else you (or anyone else) needed done on this PR, or can it be merged?

ashiklom:
@ssmithClimate I don't fully follow your questions, but I tried to clarify exactly what happens in https://github.com/JGCRI/hector/pull/342/commits/03506715e30530908276a975bb18d596db5c2177.

> Can the system suddenly switch to a different concentration?

If you go from an unconstrained year to a constrained one, then yes, that's exactly what happens. Hector will force the atmospheric CO2 to match whatever the user specified, even if it's much higher/lower than expected.

On the other hand, the first unconstrained year after a CO2 constraint should be a (relatively) smooth transition because that year's C pools are based on the previous (constrained) year's value. Of course, the more different the prescribed CO2 is from equilibrium, and the shorter the constraint period, the more aggressively the carbon-climate system should try to course-correct itself.

PR #346, opened 2020-01-16
Add function to output all variables to csv

### LINKED ISSUE #330: Enable CSV output stream from R package (or a helper to fetch all outputs)

mnichol3:
This PR aims to resolve #330 by introducing the function `fetchvars_all`, which retrieves all available Hector variables and optionally writes them to a .csv file. 

A tab-delimited text file in `/data-raw` holds the variable capability strings as well as a flag for whether or not the variable's `getData` method is able to handle a date parameter. This file will need to be updated as new variables are introduced, renamed, or existing variables are given time series capabilities. 

Additionally, a test was added to `test_hector.R` that compares the values of variables retrieved from Hector by the `fetchvars` function to those retrieved by `fetchvars_all`. 

As with last time, numerous cpp files had trailing whitespace trimmed by Rstudio settings defined in `hector.Rproj`.


R/messages.R
```diff
@@ -110,3 +110,85 @@ setvar <- function(core, dates, var, values, unit)
...
+    # Get variables that DO NOT use date arg
+    rslt_nodate <- do.call(rbind,
+                       lapply(vars_nodate, function(v) {
+                            sendmessage(core, GETDATA(), v, NA, NA, '')
+                       }))
+    
+    rslt <- rbind(rslt_date, rslt_nodate)
+    rslt$scenario <- scenario
+    
+    if (!is.null(outpath)) {
```
ashiklom: All of the code from here on is useful enough that I might consider making it its own exported function. I can think of _lots_ of circumstances where we would want to export output in a wide, pretty CSV format. If you wanted to preserve this interface, you could then just call that function here. In other words...

```r
# ...
  if (!is.null(outpath)) {
    write_wide_output(rslt, outpath, ...)
    message("Output written to ", outpath)
  }
  invisible(rslt)
}

# R/write_wide_output.R
#' Write Hector output to a wide CSV file
#' @param results Hector results `data.frame` such as the one produced by [fetchvars()]
#' @param outpath Path for writing output
#' @param ... Additional arguments to [utils::write.table()]
write_wide_output <- function(results, outpath, ...) {
  # ...
  # Note the use of ... below to allow users to modify other arguments of `write.table`
  write.table(rslt_pretty, outpath, col_names = col_names, row.names = FALSE, ...)
}
```

R/messages.R
```diff
@@ -110,3 +110,85 @@ setvar <- function(core, dates, var, values, unit)
...
+
+        # Move last column (initial_value) into 4th col position & scenario into 1st col
+        rslt_pretty <- rslt_pretty[, c(3, 1, 2, num_cols, 5:num_cols-1)]
+
+        # Set the column names for the re-ordered dataframe
+        colnames(rslt_pretty) <- col_names
+
+        # Construct output filename and absolute path
+        f_name <- "fetchvars_all.csv"
+        abs_path <- file.path(outpath, f_name)
```
ashiklom: Personal preference, but I would just make `outpath` the output file name (you could change to `outfile` to make that clearer). I could imagine users wanting to write out multiple files from different Hector simulations to the same directory.

R/messages.R
```diff
@@ -110,3 +110,85 @@ setvar <- function(core, dates, var, values, unit)
...
+
+    strt <- startdate(core)
+    end <- getdate(core)
+
+    if (is.null(dates)) {
+        dates <- strt:end
+    } else {
+        valid <- dates >= strt & dates <= end
+        dates <- dates[valid]
+    }
```
ashiklom: If you use `fetchvars` (as indicated below), most of these lines should no longer be necessary.

tests/testthat/test_hector.R
```diff
@@ -228,3 +228,38 @@ test_that('Test RF output.', {
 
     shutdown(hc)
 })
+
+
+test_that('Test fetchvars_all.', {
```
ashiklom: 👍 Nice test!

R/messages.R
```diff
@@ -110,3 +110,85 @@ setvar <- function(core, dates, var, values, unit)
...
+#'
+#' @param core Hector core object
+#' @param dates Vector of dates, optional
+#' @param scenario Scenario name, optional str
+#' @param outpath Absolute path of the output csv. Output will only be written to
+#' csv if this is given
+#' @return vars_all    Dataframe containing all Hector variables
+#' @family main user interface functions
+#' @export
+fetchvars_all <- function(core, dates=NULL, scenario=NULL, outpath=NULL) {
```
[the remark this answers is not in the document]
rplzzz: Not everybody agrees with that prescription.  There is something to be said for putting closely related terms close together and separating them from less closely related terms with spaces.  For example, to my eye `a*c - b*d` is more readable than `a * c - b * d`.  The goal should always be readability, not adherence to some arbitrary convention.

Personally, I think the argument list is more readable _without_ spaces around the `=` signs because it's easier to see at a glance the number and names of the arguments.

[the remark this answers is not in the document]
ashiklom: Fair enough. I usually try to follow the [Tidyverse style guide's recommendations](https://style.tidyverse.org/syntax.html#infix-operators), which, AFAICT, is the most widely used R style guide. But I think the No.1 rule of every style guide is (or at least should be) "If breaking this style guide makes your code more readable, break it."

I've gotten pretty accustomed to seeing `parameter = value` with the spaces, so `parameter=value` to me looks a bit cramped and less readable 🤷‍♂

R/messages.R
```diff
@@ -110,3 +110,85 @@ setvar <- function(core, dates, var, values, unit)
...
+    rslt_date  <- do.call(rbind,
+                      lapply(vars_date, function(v) {
+                            sendmessage(core, GETDATA(), v, dates, NA, '')
+                      }))
+    
+    # Get variables that DO NOT use date arg
+    rslt_nodate <- do.call(rbind,
+                       lapply(vars_nodate, function(v) {
+                            sendmessage(core, GETDATA(), v, NA, NA, '')
+                       }))
```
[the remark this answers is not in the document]
mnichol3: Ah yes that's much cleaner. I wasn't using `fetchvars` because it returned an error when trying to pass `dates = NULL`. I forgot about R's `NA`

R/messages.R
mnichol3: Good idea...I'm going to add a snippet that checks for the `.csv` extension at the end of `outfile`. 
Would it be better to stop execution if its not there or raise a warning, append `.csv` to `outfile`, and continue?

ashiklom: > checks for the .csv extension at the end of outfile

Personally, I wouldn't. Sometimes, you might want a different suffix (e.g. `.dat`, `.txt) or no suffix at all, and I tend to prefer such things not be forced on people.

My preference would be to pass arguments to `write.table`, with common arguments for `csv` as the default. I.e. Something like:

```r
write_hector_table <- function(x, file, sep = ",", row.names = FALSE, col.names = TRUE, ...) {
  pretty_table <- make_pretty(x) # all the stuff you did
  # Provide different default values for `sep`, `row.names`, and `col.names` so the default format is CSV
  # But, because they are arguments, they can still be modified if the user wants
  # Also, the <...> means "pass any other named argument down to write.table
  write.table(pretty_table, file, sep = sep, row.names = row.names, col.names = col.names, ...)
}
```

With that function, the user has a lot of flexibility about the output format. E.g.:

```r
# Tab-separated file
write_hector_table(results, "mydata.tsv", sep = "\t")
# CSV with row names
write_hector_table(results, "mydata.csv", row.names = TRUE)
# CSV without column names and a different NA value
# NOTE -- The `...` allows write_hector_table to inherit the `na` arg from write.table
write_hector_table(results, "mydata.csv", col.names = FALSE, na = -9999)
```

data-raw/units-data.R
```diff
@@ -31,4 +31,16 @@ halounits <- 'Gg'
 unitstable <- rbind(data.frame(variable=hvars, units=hunits),
                     data.frame(variable=haloemis, units=halounits))
 
-devtools::use_data(unitstable, internal=TRUE, overwrite=TRUE)
+###### Get all available hector variables (as defined by data-raw/vars_all.txt)
+vars_all <- read.table("data-raw/vars_all.txt", header=T, as.is=T)
```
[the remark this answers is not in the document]
bpbond: Seconded

R/messages.R
```diff
@@ -110,3 +110,64 @@ setvar <- function(core, dates, var, values, unit)
...
+        }
+
+        # Pivot the dataframe "wide" so years run along the x axis
+        # Catch warning from reshape()
+        suppressWarnings(rslt_pretty <- reshape(rslt, direction="wide",
+                                        idvar=c("scenario", "variable", "units"),
+                                        timevar="year"))
+
+        # Clean up the column names (value.YYYY --> YYYY)
+        col_names <- colnames(rslt_pretty)[-1:-3]
```
rplzzz: This sequence of transformations is complex and hard to follow, and I think there's an easier way to do it.  First, it isn't necessary to chop off the first three columns to rename the remaining ones.  Also, `sub` can operate on a vector, so you should be able to just write:
```R
colnames(rslt_pretty) <- sub('value\\.', "", colnames(rslt_pretty))
```

After that, if you want to rearrange the columns in the data frame, you can just access the data frame as if it were a list
```R
nc <- ncol(rslt_pretty)
rslt_pretty <- rslt_pretty[c(1,2,3, nc, seq(4, nc-1))]
```
And I think those three lines together give you what you want (though, I confess, I haven't tested it).

R/messages.R
```diff
@@ -110,3 +110,64 @@ setvar <- function(core, dates, var, values, unit)
...
+        col_names <- sapply(col_names, function(x) sub("value.", "", x), USE.NAMES=FALSE)
+        col_names <- col_names[-length(col_names)]
+
+        col_names <- c("scenario", "variable", "units", "initial_value", col_names)
+
+        num_cols <- length(col_names)
+
+        # Move last column (initial_value) into 4th col position & scenario into 1st col
+        rslt_pretty <- rslt_pretty[, c(1, 2, 3, num_cols, 5:num_cols-1)]
```
rplzzz: Using `5:num_cols-1` is error-prone and confusing to anyone reading the code.  It's hard to know just by looking at it whether you really meant 4 ... nc-1, or whether you meant 5 ... nc-1 and made an error.  It's much clearer to write `seq(4, nc-1)` because that leaves no doubt.

data-raw/units-data.R
```diff
@@ -31,4 +31,16 @@ halounits <- 'Gg'
...
+###### Get all available hector variables (as defined by data-raw/vars_all.txt)
+vars_all <- read.table("data-raw/vars_all.txt", header=TRUE, as.is=TRUE)
+
+# Split the variables into two lists: Those who can handle a date parameter
+# and those who cannot
+vars_date <- vars_all[vars_all$date_param == TRUE,][,1]
+vars_nodate <- vars_all[vars_all$date_param == FALSE,][,1]
+
+# Convert the strings representing variable names to capability strings
+vars_date <- lapply(vars_date, do.call, args=list())
```
rplzzz: Hmm.  We're taking whatever is in that file and executing it as a function.  That makes me nervous, even if we are theoretically controlling the contents of the file we read the strings from.  Could we explicitly fetch names from the Hector namespace?  That would at least limit the amount of mischief that could be gotten into.  Something like
```R
vfuns <- lapply(vars_date, getFromNamespace, ns='hector')
vars_date <- sapply(bar, do.call, args=list())
```

R/messages.R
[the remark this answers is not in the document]
rplzzz: The tidyverse developers are just these guys, you know?  We've disagreed with them about things in the past and probably will do again in the future, and that's ok.  As long as the code is readable, I don't think we should be dogmatic about spacing.  And it's even ok if the spacing varies a bit from place to place in the code, depending on who wrote it.

All those statements are just my opinions, of course.

R/messages.R
```diff
@@ -110,3 +110,64 @@ setvar <- function(core, dates, var, values, unit)
...
+
+    if (!is.null(outfile)) {
+        # Check that outfile ends with the .csv file extension
+        if (substr(outfile, nchar(outfile)-3, nchar(outfile)) != ".csv") {
+            stop("'outfile' must end in '.csv'")
+        }
+
+        # Pivot the dataframe "wide" so years run along the x axis
+        # Catch warning from reshape()
+        suppressWarnings(rslt_pretty <- reshape(rslt, direction="wide",
```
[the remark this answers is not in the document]
ashiklom: 👍 to namespace-qualified `stats::reshape`.

I looked into this and the warning is that there are records with missing time values. Basically, `reshape` doesn't like the fact that all the `rslt_nodate` rows have `NA` for the time variable. The easiest (albeit hacky) way to just circumvent the warning is to give all the `rslt_nodate` rows a fake date value -- e.g. `rslt[is.na(rslt$year), "year"] <- -999`. However, the underlying issue is that there isn't a good wide tabular data structure to efficiently store both time-dependent and time-independent variables without a bunch of `NA`s. In a very wide format like this one, where each row is a variable and each column is a year, you end up with hundreds of columns of `NA`s in rows with time-independent variables. On the other hand, if you spread the variables instead...

```r
reshape(rslt[, colnames(rslt) != "units"], direction = "wide", timevar = "variable", idvar = c("scenario", "year"))
```

...you end up with each parameter column having only one non-`NA` value. If we refactor this as a separate `hector2wide` function (per Robert's suggestion below), it may be worth including a more informative warning about trying to reshape data with missing dates.

Re: `tidyr::pivot_wider` -- I definitely think it's a much better interface, but I'm somewhat reluctant to make it a dependency for just one function, especially given the propensity of the Tidyverse folks to introduce breaking changes (and given that it has 12 dependencies of its own, including some pretty beefy ones). `stats::reshape` is clunky and confusing, with counterintuitive argument names and limited flexibility...but it does get the job done for this specific and relatively-straightforward case.

data-raw/units-data.R
ashiklom: This is a great idea. I didn't know about `getFromNamespace` -- that's very useful! We should do this.

As an aside, if you wanted extra type- and structure-safety, you could write the second line of that as...

```r
vars_date <- vapply(vfuns, do.call, character(1), args=list())
```

...which ensures that every applied call returns a length-1 character vector and throws an informative error if not. That might be particularly useful here in case the wrong thing slips into the variable list file.

R/messages.R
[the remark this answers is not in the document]
rplzzz: Interesting, thanks.  I think this is a case where the warning is definitely telling us that we're doing something we shouldn't be doing, viz., defining a wide format where many of the variables don't have the coordinate that will be running across the columns.  We should give some thought to what it is that we really want from this operation.  I suspect that what we are producing now will not fully satisfy the users who want wide-format data.

R/messages.R
```diff
@@ -110,3 +110,64 @@ setvar <- function(core, dates, var, values, unit)
...
+#' variables (e.g., concentrations & forcings).
+#'
+#' @param core Hector core object
+#' @param dates Vector of dates, optional
+#' @param scenario Scenario name, optional str
+#' @param outfile Absolute path, including the name, of the output .csv file.
+#' @return vars_all    Dataframe containing all Hector variables
+#' @family main user interface functions
+#' @export
+fetchvars_all <- function(core, dates=NULL, scenario=NULL, outfile=NULL)
```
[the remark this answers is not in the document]
mnichol3: How would I go about coaxing roxygen to put the documentation into the same file as `fetchvars`? 

Adding `@rdname fetchvars` to the `fetchvars_all` documentation produced a `fetchvars.Rd` file with a mixture of the documentation for the two functions which was very confusing to read

data-raw/units-data.R
mnichol3: Addressed in ec76fb94303245b1785bde35fa1c6c15a31cbae1

data-raw/units-data.R
```diff
@@ -31,4 +31,22 @@ halounits <- 'Gg'
...
+# Convert the strings representing variable names to capability strings.
+vfuns_date <- lapply(vars_date, getFromNamespace, ns='hector')
+vfuns_nodate <- lapply(vars_nodate, getFromNamespace, ns='hector')
+
+# Throws an error if the result of each do.call call is not a character vector
+# of length 1
+vfuns_date <- lapply(vfuns_date, do.call, character(1), args=list())
+vfuns_nodate <- lapply(vfuns_nodate, do.call, character(1), args=list())
+
+usethis::use_data(unitstable, vars_date, vars_nodate, internal=TRUE, overwrite=TRUE)
```
ashiklom: Typo.

```suggestion
usethis::use_data(unitstable, vfuns_date, vfuns_nodate, internal=TRUE, overwrite=TRUE)
```

data-raw/units-data.R
```diff
@@ -31,4 +31,22 @@ halounits <- 'Gg'
...
+vars_nodate <- vars_all[vars_all$date_param == FALSE,][,1]
+
+# Convert the strings representing variable names to capability strings.
+vfuns_date <- lapply(vars_date, getFromNamespace, ns='hector')
+vfuns_nodate <- lapply(vars_nodate, getFromNamespace, ns='hector')
+
+# Throws an error if the result of each do.call call is not a character vector
+# of length 1
+vfuns_date <- lapply(vfuns_date, do.call, character(1), args=list())
+vfuns_nodate <- lapply(vfuns_nodate, do.call, character(1), args=list())
```
ashiklom: Need to use `vapply` here, not `lapply`.

mnichol3: Does `lapply` not check the return value? (I 100% trust your feedback, just curious)

ashiklom: `lapply` always returns a list the same length as the thing being mapped over.

```r
result <- lapply(x, FUN, a = a, b = b)
# ...is basically the same as...
result <- list()
for (i in seq(1, length(x)) {
  # In R, lists can hold anything, so this always works
  result[[i]] <- FUN(x[i], a=a, b=b)
}
```

`vapply` takes an _additional_ argument, confusingly called `FUN.VALUE`, that is the same length and type as the expected output of `fun(x[i], ...)`.

```r
# If...
y <- FUN(x[1])
class(y)
#> numeric
length(y)
#> 1
# ...then...
result <- vapply(x, FUN, numeric(1), a = a, b = b)
# ...is roughly the same as...
# Pre-allocate a numeric vector with `length(x)` elements
result <- numeric(length(x))
for (i in seq(1, length(x)){
  r <- FUN(x, a = a, b = b)
  # Check that the output is the specified length and type
  stopifnot(length(r) == 1, class(r) == "numeric")
  result[i] <- r
}
```

Read more by running `?lapply` in the R console (they're both documented on the same page). I also like [this SO answer](https://stackoverflow.com/a/7141669/2477097), though I think it undersells `vapply`. It's true that knowing the output type in advance makes it a bit faster, but I really like the added predictability and explicit error when something unexpected happens.

ashiklom:
@mnichol3 Looks like there's a merge conflict here that you'll have to resolve by hand. You'll have to `git pull <jgcri/hector> master` and then manually fix the conflict in `units-data.R`, and then re-run it to generate a new, clean `sysdata.rda`. This is an annoying, but expected, consequence of having a bunch of PRs open in parallel (I've had to do this a bunch of times across all my open PRs). That said, this should be a quick fix since it only affects two files, one of which is generated programmatically.

ashiklom:
I just merged #341, which introduces a few more conflicts. You should be able to address them as follows:

- `NAMESPACE` can just be regenerated via Roxygen.
- `src/ocean_component.cpp` -- The more recent update, with the `H_ASSERT` conditioned on `varName != D_OCEAN_CFLUX`, is the correct one.
- `test_hector.R` -- Both tests should be included; they are not mutually exclusive.
- In `data-raw/vars_all.txt`, `LAND_CFLUX` and `OCEAN_CFLUX` now _do_ expect dates. Because of this, you'll also need to rerun the `units-data.R` script to (again) regenerate `sysdata.rda`. Note that there is also a [bug](https://github.com/JGCRI/hector/issues/354) that prevents at least the `LAND_CFLUX` from being retrieved in year 1745, so you'll have to adjust your `fetchvars_all` test to read values from 1746.

R/messages.R
```diff
@@ -110,3 +110,64 @@ setvar <- function(core, dates, var, values, unit)
...
+
+#' Fetch all the available Hector variables
+#'
+#' This function is similar to fetchvars, except it fetches all available
+#' Hector variables, including exogenous inputs (e.g., emissions & pre-industrial
+#' concentrations), parameters (e.g., alpha, beta, ECS), and output
+#' variables (e.g., concentrations & forcings).
+#'
+#' @param core Hector core object
+#' @param dates Vector of dates, optional
```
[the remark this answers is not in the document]
mnichol3: Addressed in 14f96fc8c31edaea17a1143597cb9468d0981d76 & 1f998858caea5b25a559df8635e6eb83e78a15c2

R/messages.R
```diff
@@ -110,3 +110,64 @@ setvar <- function(core, dates, var, values, unit)
...
+#' @param scenario Scenario name, optional str
+#' @param outfile Absolute path, including the name, of the output .csv file.
+#' @return vars_all    Dataframe containing all Hector variables
```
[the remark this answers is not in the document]
mnichol3: Addressed in 14f96fc8c31edaea17a1143597cb9468d0981d76

R/messages.R
```diff
@@ -110,3 +110,64 @@ setvar <- function(core, dates, var, values, unit)
...
+
+    rslt <- rbind(rslt_date, rslt_nodate)
+
+    if (!is.null(outfile)) {
+        # Check that outfile ends with the .csv file extension
+        if (substr(outfile, nchar(outfile)-3, nchar(outfile)) != ".csv") {
+            stop("'outfile' must end in '.csv'")
+        }
+
+        # Pivot the dataframe "wide" so years run along the x axis
```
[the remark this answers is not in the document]
mnichol3: Data Frame reshaping and file writing functionality removed in 14f96fc8c31edaea17a1143597cb9468d0981d76. The function looks a lot cleaner now!

cahartin:
@rplzzz can you take a look at this one and see that @mnichol3 addressed your comments?

cahartin:
@mnichol3 where are we on this one?

PR #347, opened 2020-01-17
Concentration constraints for CH4, N2O, and halocarbons

ashiklom:
This PR adds concentration constraints for CH4, N2O, and halocarbons. The general approach is similar to CO2 constraint (#342) and is as follows:

- At any given time step, check if the constraint has a value for that time step (e.g. `CH4_constrain.exists(t)`)
  - If it does, just set the corresponding atmospheric concentration value and do basically nothing else (e.g. `CH4.set(t, CH4_constrain.get(t))`
  - If it doesn't, proceed with Hector's standard atmospheric chemistry of figuring out what the concentration should be based on emissions and concentrations of other relevant species.

All of the constraints have a consistent naming scheme: `<component>_CONSTRAIN` from R (e.g. `CH4_CONSTRAIN()`, `HFC4310_CONSTRAIN()`) and `<component>_constrain` in C++/

NOTE: CH4 and N2O have a special variable for their pre-industrial concentration that could conflict with the constraint. I reconciled this by assuming that, if you provide a concentration constraint for the timestep corresponding to the pre-industrial, _that_ is the concentration you want, _regardless of the prescribed pre-industrial value_. In terms of implementation, this means code like the following:

https://github.com/ashiklom/hector/blob/10ea19ec3def3b341d2fa02f73a49c38b1ccf645/src/ch4_component.cpp#L128-L137

NOTE: This PR is a superset of #342. That one is smaller and has already been reviewed, so it should probably be merged first. Once that is merged, this one will have a slightly smaller diff.


rplzzz:
@ashiklom If it's ok with you, I'm going to wait until #342 is merged before I tackle this one.

ssmithClimate:
@ashiklom Similar questions as I had on CO2 regarding smooth transitions to/from constraints (or not).

ashiklom:
@ssmithClimate The constraint text you wrote for #342 basically applies to these constraints as well. I moved that paragraph into its own section in the Constraints documentation, and tweaked the wording to make it general ([link](https://github.com/JGCRI/hector/pull/347/files#diff-d4b78cac323653e83e50164616d3e806R91)). Let me know if you think it needs further clarification.

cahartin:
@ssmithClimate can you take a quick look?

ssmithClimate:
@ashiklom the new text looks good, thanks! One question still. It says they work similar to Co2 constraints, but for CO2 there's the deep ocean to take up the slack in the carbon cycle. How is this done for the other gases? Would this lead to more discontinuous behavior as compared to CO2?

ashiklom:
@ssmithClimate As far as I can tell, the main reason we move carbon around in the CO2 constraint is to preserve the C mass balance. The reason we dump the carbon into the deep ocean is because doing that has minimal impact -- in fact, we've had discussions about dumping into the "Earth C" pool instead, which has _no_ impact on the C cycle by definition. 

We don't track the mass balance of any of these other gases, so similar techniques aren't necessary from a programming/math perspective. Moreover, based on Hector's current structure, I don't think there's anything fundamentally different about passing emissions for these gases vs. concentrations from an atmospheric chemistry perspective. In fact, the way we _used_ to do concentration constraints was to [numerically solve for the emissions that produced that pathway](http://jgcri.github.io/hector/articles/hector_apply.html). At one point, I worked out that there is a simple analytical relationship between the emissions and concentrations of any Halocarbon at any timestep. CH4 and N2O are more complicated because of their interactions with OH and with each other, but based on my understanding of the code, there should still be a one-to-one relationship between emissions and concentrations (conditioned on the _previous_ time step's concentrations of the other gases; this is distinct from the carbon cycle, where we solve the system of differential equations for the current time step). So I think any "discontinuity" a user could introduce with a concentration constraint could also be introduced with a emission year.

ssmithClimate:
@ashiklom good points! (N2O doesn't depend on OH - it does depend on its own concentration in a non-linear way. Same result w.r.t. your note however...) So how about this text modification:

"These all work in the same manner, similar to the atmospheric CO2 constraint (above), except that there is no equivalent to the carbon pool adjustment for these gases."

ashiklom:
@rplzzz Did you still want to take a look at this, or are you satisfied with Corinne and Steve's reviews?

rplzzz:
> @rplzzz Did you still want to take a look at this, or are you satisfied with Corinne and Steve's reviews?

If they're satisfied, so am I.

Issue #351, opened 2020-01-24
Trailing whitespace in cpp source files


mnichol3:
The majority of the lines of cpp source code have trailing white space. When files are saved in RStudio, the trailing white space is removed per `hector.Rproj`:
```
StripTrailingWhitespace: Yes
```
Git detects these changes and presents them in pull requests along with meaningful changes. As many have noted, this is quickly becoming a nuisance in PRs as it adds tens of extra changed files.

rplzzz:
Echoing my comment on #353, would it not be easier to change the `StripTrailingWhitespace` setting so that Rstudio does not make these changes?

mnichol3:
That certainly is the simpler solution. However, the whitespace serves no purpose and has drawn numerous comments in PRs, so why not remove it from the existing code in addition to adding an IDE config file that removes any new trailing whitespace on save? 

In addition to making the code cleaner (clean code is good code), it'll make navigation via cursor/arrowkeys much easier in vim (I use vim a good bit for tweaking code in my Hector repo on `pic`) as well as more popular IDEs.

rplzzz:
It serves no purpose, but it also causes no harm, and removing it does not make the code cleaner in any meaningful way because the spaces are invisible to humans.  

Meantime, the price you're paying is that you're demanding that people use an editor that conforms to that standard.  From the look of the diffs there is at least one editor in use that inserts indentation even on blank lines.  So, now we have to track down which one that is and reconfigure it or make the user change editors.  It seems like a lot of annoyance for not much benefit.

Finally, there is no way to ensure that new developers joining the team are using a compliant editor, so now we have to check the trailing whitespace on each PR, or we have to occasionally fix the trailing whitespace that will inevitably creep in.  It really is just a lot easier to have editors not automatically adjust whitespace.

BTW, trailing whitespace is significant in some dialects of Markdown.

mnichol3:
> the price you're paying is that you're demanding that people use an editor that conforms to that standard

We already ask that people conform to the [Hector style guide](https://jgcri.github.io/hector/articles/manual/StyleGuide.html), which includes many common C++ standards & conventions. I fail to see how the addition of another [common convention](http://google.github.io/styleguide/cppguide.html#Horizontal_Whitespace) is an issue.



> Finally, there is no way to ensure that new developers joining the team are using a compliant editor, so now we have to check the trailing whitespace on each PR, or we have to occasionally fix the trailing whitespace that will inevitably creep in

[...]

PR #353, opened 2020-01-24
Remove trailing whitespace from source files

mnichol3:
Addresses #351 by stripping trailing whitespace from `.cpp`, `.hpp`, and `.R` files. This should put an end to PRs being bloated with whitespace changes when files are edited and saved in RStudio.


ssmithClimate:
Should probably coordinate with others working on active PRs. Not sure if its easier on them to commit this now or after their PRs are in.

bpbond:
@kdorheim OK to merge with #366 to follow up shortly?

kdorheim:
👍👍 sounds good @bpbond

bpbond:
Thanks @mnichol3

PR #358, opened 2020-02-20
Correct RCP60 emissions filename

### LINKED ISSUE #357: RCP60 emissions filename is incorrect
The name of the emissions file for RCP60 is `RCP6_emissions.csv`, instead of `RCP60_emissions.csv`.

Hector still runs properly due to the `RCP6` name appearing in the `hector_rcp60.ini` file 
(Ex: `ffi_emissions=csv:emissions/RCP6_emissions.csv`), which may be why this has slipped through the cracks. 

It appears that the file has had this name since at least v2.3.0

mnichol3:
Change the name of `RCP6_emissions.csv` to `RCP60_emissions.csv` to avoid confusion. Also correct any occurrence of the `RCP6` filename within the RCP60 `.ini` files.

Closes #357


cahartin:
@mnichol3 did we do this for RCMIP?

kdorheim:
@bpbond and @mnichol3 actually I think I goofed up on this one. This PR will not change any R hector code because the typo was the label of the emissions and not the ini file so this change should not impact any code for users on the R interface side. @bpbond will it change the ability to run hector directory from c++? If not I think it can go ahead and be merged.

bpbond:
@kdorheim Ah I see what you mean...ok!

PR #379, opened 2020-06-09
Provide link to HectorUI

bpbond:
@evanoffPNNL requested a link to the Shiny HectorUI, which seems like a good idea.


bpbond:
Thanks @kdorheim

PR #388, opened 2020-06-19
Update DESCRIPTION

bpbond:
So the Hector package DESCRIPTION file needs updating:
* Given that @rplzzz has moved on but this is PNNL property, I propose changing the package maintainer ("cre") to @kdorheim , who is one of two repo admins
* However, as far as I know the original package code was almost entirely Robert's work. To reflect this I've changed him to be the only "aut" (author), while the rest of us are "ctb" (contributors). Does this make sense to folks? 
* Updated Corinne's email
* Added @ashiklom as a contributor
* Added ORCID IDs for all
* Tweaked language

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,7 +20,11 @@ Authors@R: c(person("Robert", "Link",
 	     person("Pralit", "Patel", 
 	        email = "pralit.patel@pnnl.gov", 
 	        role = "ctb",
-	        comment = c(ORCID = "0000-0003-3992-1061")))
+	        comment = c(ORCID = "0000-0003-3992-1061")),
+	     person("Alexey", "Shiklomanov",
+	        email = "alexey.shiklomanov@nasa.gov",
+	        role = "ctb",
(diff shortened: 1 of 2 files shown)


rplzzz:
@bpbond ,

R uses the "creator" role to denote the package maintainer, which is a little weird, since it means that in any long-running project the "creator" will likely end up being someone who didn't create the project and maybe wasn't even part of the team at its inception, but there you have it. 
 In R's taxonomy, the maintainer is specifically the person who fields questions and bug reports (http://r-pkgs.had.co.nz/description.html#author), so whoever goes in that role should be prepared to take responsibility for those things.  

I'm indifferent to how my name appears in the list.  Considering everything that has happened over the past year, I'd be just as happy not to have it appear at all, but I'll leave it to your discretion.

bpbond:
@rplzzz Yeah the "creator" thing is weird, agreed. OK, if you don't object, will leave as is. Thanks.

kdorheim:
Looks good to me!

Issue #304, opened 2019-05-02
Should components use specific land/ocean temperature instead of GMT?


ssmithClimate:
Now that hector separately reports separate temperature values for land and ocean air, and ocean SST, should components use the appropriate temperatures instead of global mean surface air temperature (GMT)?

Changing what temperature value is used would alter results in two ways. 1) The numerical value of these vary for any given year, and 2) also the time trend is different as compared to GMT.

These components currently use GMT:
slr_component.cpp
simpleNbox.cpp
ocean_component.cpp
oceanbox.cpp
temperature_component.cpp

Mechanically this is easily done. 

However, would we need to change parameter values to accommodate this? That is, care the current parameters (heat transport, Q10, etc.) derived relative to GMT, so they would need to be re-derived if something other than GMT was used?

kdorheim:
Addressed in #562

Issue #68, opened 2015-02-06
Distinguish between atmosphere-land and atmosphere-ocean feedbacks


bpbond:
Some climate feedbacks operate mostly over land, which warms faster than the ocean. We'd like to account for this.

Not yet sure if this should be in milestone 1.0 or not.

cahartin:
with the addition of DOECLIM we should be able to "easily" separate temperature over land and ocean to account for the differential warming.

kdorheim:
Adding DOECLIM in Hector v < 2.5 and its integration #562 accounts for the warming rates being different over the ocean & land.

PR #389, opened 2020-08-17
Land ocean warming ratio

skygering:
This change adds land-ocean warming ratio as a tunable parameter; this change is backwards-compatible. Users can now set land-ocean warming ratio, which will overwrite the average land, ocean surface, and ocean air temperatures to values that preserve the land-ocean warming ratio. Additionally, I modified the code so that `ocean_component.cpp` uses the ocean air temperature and `simpleNbox.cpp` uses land temperature where they previously used global temperature. (This partially addresses #304 but may need more work as referenced in the issue).These downstream changes affect the carbon cycle calculations and thus affect the global temperature calculations. I have also exposed the land-ocean warming ratio parameter within R for ease of use. As can be seen below, changing the land-ocean warming ratio within Hector has the potential to greatly change the results. 
Land-ocean warming ratio

I believe that this change could be even more pronounced. The values calculated for land temperature and sea surface temperature are not fed directly back into the global temperature calculations due to a potential bug specified in issue ___.

Documentation has been updated with the exception of adding `LO_WARMING_RATIO()` to [this](https://jgcri.github.io/hector/reference/index.html) page as I could not find the code for it.


bpbond:
This related to (would close) #304 #68

kdorheim:
Skylar's awesome work was incorporated & merged into Hector with #562!!

PR #395, opened 2020-10-16
Krd pr materials

kdorheim:
**Description** 

This PR 
1. adds information about contributing to Hector
2. fixes a documentation change requested by @crvernon (see read me)

POTENTIAL problem, I am not sure about some of the hyperlinks that I've added. They might not be necessary, if that is the case let's remove them. If we want to include them I think some of them will have to be added after the PR is merged... 

**Type of Development** 
Patch change, these documentation changes make no impact on Hector outputs.


bpbond (approved):
Looks solid. 👍

PR #397, opened 2020-10-26
NBP (land-atmosphere C flux) constraint

bpbond:
## DO NOT MERGE YET

**Background.** This PR adds a net biome production, the land-atmosphere exchange, constraint to Hector (output `land_atm_flux`). This capability is needed for a current project of Han Qiu and Min Chen, and will be required in the future for GCIMS SFA work. 

The trick is that NBP is the sum of net primary production (NPP) and heterotrophic respiration (RH), so if we're imposing a constraint on it these need to be adjusted, and this needs to happen _during_ the solver, not after it, so that the land fluxes stay roughly consistent. It's not simply a matter of substituting in a new value, e.g. like the existing Tgav constraint.

**Method.** After NPP and RH have been computed in `calc_derivs`, we compare NPP-RH-LUC to the constraint value, if there is one; any difference is eliminated by increasing NPP by `diff/2` and subtracting the same value from RH. The NPP and RH sub-fluxes are then changed as well. After the solver finishes, we don't always end up at the exact constraint value we want, so basically the same thing happens again: NPP and RH are adjusted.

**Compatability.** This should be fully backward-compatible with existing Hector inputs and outputs. It adds a new setting `nbp_constrain` to the INI files and R interface.

**Testing.** We test three constraints: the unconstrained NBP output values (`rcp26_nbp_constrained`), and constraints that are lower (`rcp_han6`) and higher (`rcp_han8`) for the 2015-2100 period.
##### Fig 1
constraints

The good news is that the han6 and han8 cases function as expected–the model hits the constraint exactly:
##### Fig 2
constraint_success

...changing atmospheric C and temperature in the direction we'd expect:
##### Fig 3
constraint_effects

(Note that when the han8 constraint goes away, there's a big pulse of C out to the atmosphere. This is expected.)

**Problems.** There are some problems still:
1. I don't understand why the rcp26 'constraint', that should exactly mimic the unconstrained run, doesn't. (See above.) It's very weird and I don't know if there's a problem in my R plotting routines or something numeric in the C++ code.
2. The `simpleNbox` output files (both csv output stream and log) currently spit out the _unadjusted_ versions of NPP and RH when operating under a constraint. That is, the `npp` - `rh` - `luc` won't equal `land_atm_flux`. Not quite sure how to solve this at the moment, but either the npp and rh calculating functions need to know about the constraint (ugh), or we store the adjusted values so that the visitor and access them (better?).


inst/include/component_data.hpp
```diff
@@ -287,7 +287,8 @@
 #define D_F_LUCV                "f_lucv"
 #define D_F_LUCD                "f_lucd"
 #define D_NPP_FLUX0             "npp_flux0"
-#define D_CO2_CONSTRAIN          "CO2_constrain"
+#define D_CO2_CONSTRAIN         "CO2_constrain"
```
kdorheim, line 290: Good catch!

PR #399, opened 2020-11-18
Krd actions issues

kdorheim:
Trying to figure out what is going on with the different issues we've run into with CI and package down, ideally, there would be a better way to resolve this than just opening PRs all the time


bpbond:
It's really frustrating to try to debug this kind of thing. As you say, nonstop PRs.

kdorheim:
@bpbond  DO NOT APPROVE, I think we will end up deleting this PR & branch

PR #404, opened 2020-12-09
Fix typo in makefile.standalone re LDFLAGS and -Wl,-rpath

payerle:
See issue ticket #403
The original had a space between 
-Wl,-rpath and $(BOOSTLIB)
This causes the linker to get a -rpath without an argument, and the
$(BOOSTLIB) is passed as an argument to the compiler.

What is wanted is either
-Wl,-rpath,$(BOOSTLIB)
or
-Wl,-rpath -Wl,$(BOOSTLIB)
to prevent linker from complaining about -rpath w/out an argument.

PR #406, opened 2020-12-10
GitHub Actions and Logger Fixes

crvernon:
This pull request adds/fixes the following:
- Fixed the `pkgdown.yaml` workflow.  This now automatically builds and deploys what is being pushed to the `gh-pages` branch.  No additional building necessary.  The push branches are set to `master` or `main` so this will not execute in the `github_actions` branch.  I have tested this on my fork but did not want to trigger a documentation build for Hector until the base branch of this PR was ready to go into the `main`.
- Renamed the `R-CMD-check.yaml` workflow to `rcmd.yml`
- Fixed the RCMD workflow to run on R 3.3-5 + release for Ubuntu 16.04. Using 16.04 for testing is a common practice.  However, the `ubuntu-latest` image will be migrating to 20.04 by the end of 2020.  So we should plan on testing with that image as well when it passes muster.
- Fixed the RCMD workflow to run on R 3.3-6, 4.0, and release using the `windows-latest` image
- Fixed the RCMD workflow to run on R 3.6, 4.0, and release using the `macOS-latest` image
- **Note** the install of XQuartz for MacOS images.  This was done to allow the vignettes to build properly.
- **Note** the change from fail on warning to fail on error for the `Check` block in the `rcmd.yml` workflow
- @pralitp updated `logger.cpp` to use the Boost filesystem for `Logger::chk_logdir`.  The original code for the conditional statement used when Windows was detected did not build the `logs` directory; it only returned 1.  The new setup now works on both Unix and Windows.
- Fixed the "Write out logs" test in `test_wrapper.R`.  The original test used a Unix specific command and file path which caused a fail on Windows.  The code now will execute locally, or via CI for Unix and Windows.  It detects the platform and applies the appropriate path to the logs directory using `here::here()` to fetch the directory name.

Since this is coming in off of a fork, the following needs to be changed upon arrival:
- Remove `github_actions` as a push branch in `command-line.yaml` once an action cycle has completed for testing
- A subsequent PR will follow to fix and re-enable the coverage tests


crvernon:
@kdorheim @bpbond the use of `appveyor` for Windows should now be disabled.  This PR allows all Windows builds.

kdorheim:
Do we need to continue to use `appveyor`? Is using both `appveyor` and `github actions` unnecessary, a good, or bad practice?

crvernon:
@kdorheim  No we do not.  Appveyor needs to be taken down.  It was likely in existence because Travis-CI had trouble with Windows builds.  GitHub actions now solves this.

crvernon:
@kdorheim it looks like you already removed the appveyor config file.  Since the original YAML file did not have a branches > only specification, appveyor initiates on all branches, regardless of the presence of the file in a current branch.  This is done to ensure that the main will not be overridden by an erroneous build.  All that to say, when you pull this branch into the master, the failed appveyor test will cease to exist; but until then, we are stuck with it.

crvernon:
Also, it looks like #404 will fix the coverage tests as well.  So I will not be submitting the subsequent PR for that.  See success on action https://github.com/JGCRI/hector/actions/runs/413130899

bpbond (approved):
Fantastic @crvernon — thanks.

DESCRIPTION
```diff
@@ -43,7 +43,8 @@ Suggests:
     rmarkdown,
     ggplot2,
     magrittr (>= 1.5), 
-    ragg
+    ragg,
```
kdorheim, line 46: should we add the package versions?

kdorheim (approved):
This is great! Thank you so much!

DESCRIPTION
crvernon: Sure, see https://r-pkgs.org/description.html#versioning

I am OOO today. Could you add these in? You may want to check the actions install log to see what versions the VMs are using too. 

Thanks!

kdorheim: Sure thing, although now that I think about it the package version dependencies might be a larger question... @bpbond do we want to think about a package manager or something? Should we punt this to that decision?

bpbond: I think specifying reasonable minimum versions is almost always a good idea, both for R and the package dependencies. 

>package manager

I would vote punt for now.

PR #408, opened 2020-12-10
Add a CCEXTRA variable to CFLAGS to allow users to easily add flags

payerle:
This is useful, e.g. to add -fPIC if building both standalone and R
package.

See e.g. #407


bpbond:
Hi @payerle I'm having trouble finding much information about this flag. Can you explain its use a bit, and/or provide a documentation link? Thanks.

bpbond:
😆  ...should have clicked through to #407 first, sorry!

PR #401, opened 2020-11-28
Shift from Travis to GitHub Actions

kdorheim:
@crvernon and @bpbond so while I don't think the pkgdown github action is ready yet I think that the github actions for R-CMD-check and Command Line _might be ready_ the changes for pkgdown will probably have to happen in a different PR but if possible I'd like to get this one merged in. 

As part of this PR we are disabling Travis (cause it won't be open source) and also appveyor so we have all tests located in one place. 

Anyway, I would really appreciate if someone more familiar with github actions could take a look at this. Also, I am not sure if this is the case on your end but there are some old workflows on github actions that I have distabled. I am trying to delete them but I might not have the correct permissions to do that.

PR #409, opened 2020-12-16
codecov github actions

kdorheim:
As @crvernon pointed out there was a mix up in #401 and #406, and code coverage workflow was deleted. This PR adds the yaml file to re-enable this Github actions workflow. @crvernon please take a look to make sure I am adding the correct work flow, it is the one you [committed here](
https://github.com/crvernon/hector/blob/b993c3a4c37ef1970c2ac5a5b566ff29beac4852/.github/workflows/test-coverage.yaml).


bpbond (approved):

bpbond:
Thanks @kdorheim

crvernon:
@kdorheim I'll see what's up with that failing coverage test likely tomorrow.

kdorheim:
@crvernon it looks like the code coverage tests are failing related to the log tests 

``` test_check("hector")
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure (test_wrapper.R:71:5): Write out logs ───────────────────────────────
dir.exists(log_dir) is not TRUE

`actual`:   FALSE
`expected`: TRUE 
── Failure (test_wrapper.R:74:5): Write out logs ───────────────────────────────
length(list.files(log_dir, pattern = ".log")) not equal to 41.
1/1 mismatches
[1] 0 - 41 == -41

[ FAIL 2 | WARN 0 | SKIP 0 | PASS 240 ]
Error: Error: Test failures

```

which is confusing because shouldn't these failures also be triggered during ruing the R-CMD tests as well? Could you take a look at this?

crvernon:
@kdorheim let me take a look on Monday and I'll see what is up

kdorheim:
@crvernon I've done a bit of digging into this and it looks like we are running into an issue with `test_wrapper.R` the `Write out logs` test when running on codecov platform, we have run into issues with the logs/log tests on other platforms but I think you and @pralitp were able to come up with a workaround. Do you think that setting up the codecov to use a non-macosx platform for the test would resolve the issue? Or does the log dir need to be set up to `/private/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/RtmpetRAo1/R_LIBS57996308e1dc/hector/hector-tests/` when running on codecov? 

Why is `Rscript -e 'remotes::install_github("r-lib/covr@gh-actions")'` pulling from `gh-actions` instead of master? 

Also looking at [hector's codecov](https://app.codecov.io/gh/JGCRI/hector/) it looks like everything is 2 months old but there has been some activity on the repo since then...

bpbond:
>Why is Rscript -e 'remotes::install_github("r-lib/covr@gh-actions")' pulling from gh-actions instead of master?

I don't know why Jim Hester has it set up that way, especially since the `gh-actions` branch hasn't been updated in 16 months. 😕

bpbond:
Note that we had a _lot_ of R CMD tests — 3.3, 3.4, 3.5, 3.6, 4.0, release on three different platforms. I have trimmed this back to just testing release and 3.6 for each.

bpbond:
For future reference, both the `test_wrapper.R` and `codecov.yaml` scripts were unnecessarily complicated in crucial parts. (Kalyn provided a crucial insight [above](https://github.com/JGCRI/hector/pull/409#issuecomment-756199524)—why wasn't the test suite failing in the R CMD checks too?)

@kdorheim if you think everything looks good, merge when ready!

PR #413, opened 2021-01-21
Tracer pools

skygering:
At this point, earth_c and atmos_c are both trackedvals, and fossil fuel flux from earth_c to atmos_c is tracked. The outputvisitor is set up, but it prints unitvals for reasons I will explain below. I have a few questions on the code for @bpbond , @dawnlwoodard , and @kdorheim . If you have time please comment on any of the following! 😄 

- [ ] Turning tracking off and then back on has inconsistent effects if things are done to the code in between. Some of the methods basically wipe the map, while others leave the map as it was when tracking was turned off. Should we make this consistent? We could wipe the map when tracking is turned off. However, this would cause problems if the map was turned back on as the initialized key (i.e. "atmos" for the atmosphere pool) would no longer exist. I am okay with just calling this undefined behavior (i.e. no guarantees on how things work if you turn tracking off then back on.) Thoughts?
- [ ] Similarly, the way I integrated trackedval to have unitvals in the ODE solver means that the time series of each of the pools still holds unitvals. Thus, when these are used to reset Hector, I am not currently able to reset the map to how it was. The other major problem with this is that getData uses the time series, which is how the data for the outputvisitor is retrieved. Thus, I am going to have to fix this. I am not sure how yet as time series has some interpolation functions (I assume having to due with the ODE solver), which don't seem like they will work too well with trackedvals. I might need to look into changing tseries.hpp. Any thoughts welcome! 
- [ ] Should we make the setTracking() and isTracking() functions static? It is already getting confusing in testing to have to turn tracking on for each of the pools. Do we have a reason for having a separate track for each pool? Is there a case where we would only want to track a subset of the pools?
- [ ] Do we want to make trackedval namespace Hector like unitval is? 
- [ ] Adding negatives (actually subtraction) and subtracting negatives (addition) don't work right now since there is such different behavior for addition and subtraction and the sign of the flux doesn't affect that behavior. I need to add that functionality. This is especially confusing for RCP 2.6 where some of the fluxes change signs (e.g. the fossil fuel one) when warming is mitigated. That is why I am using set() for earth_c right now. 
- [ ] I had to add two set() functions. One of them (with value, units, and pool name) is for initial setup. The way simpleNBox is set up, I needed a default constructor (seems to be called implicitly at the beginning), so this is for right after that. I also have a set function that takes in a unitval for reset, and for a few small situations when tracking is off. Again, this is a use at your own risk function, although the problems are widely covered by the risks of reset and turning tracking off. 
- [ ] Note: I added a flux_from_pool() method and a adjust_pool() method. The flux_from_pool() makes a trackedval flux of a specified value from a specific pool. The adjust_pool function is basically taking the place of the temporary addition of unitvals we talked about. It will add or subtract the needed amount of flux to get the pool to the desired value. It takes advantage of the preexisting operator+ and operator- functions. This was better than the temporary addition function as it also takes care of finding the deficit flux, which would be a lot of repeated code if we had to do that for every single pool. 
- [ ] I am going to add output CSVs to the gitIgnore.


skygering:
@bpbond I have sort of an interesting idea for +/-. I don't think the solution we talked about yesterday (01/28)  is going to work. If you add a negative flux to one pool (which we determined that we could turn into subtraction) that necessitates that you are subtracting that negative from another pool as the carbon had to come from somewhere. Since the second thing will throw an error like we discussed yesterday (since we don't know the composition of the pool), that won't work. However, I currently have an adjust_pools method that adjusts a pool up or down to the unitval value of the pool from calcDerivs. I could do something similar with the flux_from_pool method where it is called like 'poolFluxIsFrom.(poolFluxGoesTo, fluxValueUnitVal)` and then just do all the addition and subtraction within this function as I would have the value of the flux, and depending on if it is positive or negative I would know how to adjust the two pools. This takes away the explicit addition and subtraction of pools within stashC. I don't know how you feel about it. I am going to try it this weekend.

inst/include/h_util.hpp
```diff
@@ -28,6 +28,6 @@
  */
 #define MODEL_VERSION "2.5.0"
 
-#define OUTPUT_DIRECTORY "output/"
+#define OUTPUT_DIRECTORY "inst/output/"
```
bpbond, line 31: We need to fix this.

bpbond:
I think we can close this now.

PR #414, opened 2021-01-26
Remove deprecated output visitor

kdorheim:
**Describe development**

Remove depreciated output visitor, the output stream visitor has been used in place of the output visitor for several years now. Also deleted some unneeded include header file calls in C++. Note that I did not clean up all of the include header file calls only in a handful of files that where we noticed this. 

**Type of development** 

* PATCH: minor changes (e.g., documentation or bug fixes)

**Checklist**
- [NA] [Roxygen](https://roxygen2.r-lib.org/) where applicable but also thorough and coherent inline documentation 
- [NA] [Unit tests](https://r-pkgs.org/tests.html) when applicable
- [NA] Complies with the [Hector code style guide](https://jgcri.github.io/hector/articles/manual/StyleGuide.html)
- [x] The PR passes automated testing suitev

**Other Notes** 

I did run into some issues when rebuilding Hector in Xocde after the cpp file as removed, it turns out the [build phases](https://stackoverflow.com/questions/10167442/whats-the-xcode-no-such-file-or-directory-error) had to be updated. I suspect that something similar would have to be done with VS sln. After looking into it I realized that the VS sln is extremely out of date, see #415 for more details.


inst/include/forcing_component.hpp
```diff
@@ -15,17 +15,8 @@
  */
 
 #include "imodel_component.hpp"
-#include "ch4_component.hpp"
```
bpbond, line 18: 👏 awesome, thank you

bpbond (commented):
Do we want to remove the `comparison_data/` entirely? Would that information be useful to e.g. @dawnlwoodard 's `hectoreval`?

inst/include/forcing_component.hpp
kdorheim: > Do we want to remove the `comparison_data/` entirely? Would that information be useful to e.g. @dawnlwoodard 's `hectoreval`?

That is a great question for @dawnlwoodard if they are using it in `hectoreval` then we should leave it alone. Once upon a time CH had replacing the comparison data on my to-do list because we weren't entirely sure exactly where the data came from...

kdorheim:
@bpbond I spoke with @dawnlwoodard about `hectoreval` you are correct it does use the `comparison_data`. I've added those files back in the same location. However, speaking with @dawnlwoodard it sounds like `hectoreval` will eventually switch to updated data produces and may process its own data to be consistent, at that point in time we can go ahead and remove the `comparison_data` files.

bpbond (approved):
Looks good, then. Anything else you want to do here @kdorheim ?

PR #417, opened 2021-02-03
Reformat R code and enforce style checks on PRs and commits

bpbond:
This PR reformats the R codebase to follow `tidyverse` style (using the [styler](https://styler.r-lib.org) package) and adds a GitHub Actions integration to check that PRs and commits follow this style (using [lintr](https://github.com/jimhester/lintr)).


kdorheim (commented):
@bpbond I think that this looks good and is a great idea! Something that is unclear clear to me is with the github actions if/what developers will have to make sure the PR passes the CI. Will they need to make sure to run some specific R code locally? Or will the github actions styler calls take care of it? 

Could you add some details clarifying what is being enforced? And what needs to be done to pass the actions to the beginning of the lint.yaml file?

bpbond:
Thanks @kdorheim . The tidyverse style guide is [here](https://style.tidyverse.org). Based on previous repos I've worked on, I think `lintr` will give developers an informative error message telling them how to fix things. I will test this by opening a new PR against the `rstyle` branch with a style error.

Note we could also set up pre-commit hooks to run `lintr` on our individual machines; cf #418 .

kdorheim:
> Thanks @kdorheim . The tidyverse style guide is [here](https://style.tidyverse.org). Based on previous repos I've worked on, I think `lintr` will give developers an informative error message telling them how to fix things. I will test this by opening a new PR against the `rstyle` branch with a style error.

Cool, I am curious to see what error messages are returned! 

> Note we could also set up pre-commit hooks to run `lintr` on our individual machines; cf #418 .

What is the downside to setting up the pre-commit hooks? Would each of us have to do it individually? If a contributor didn't have the pre-commit hook set up would the burn of making sure the code passes fall on them? Or would it mean that we would have to check out their branch and run the pre-commit hook?

bpbond:
>Cool, I am curious to see what error messages are returned!

There were quite a few, and I've actually just loosened several of the `lintr` check defaults: a max line length of 100 instead of 80; don't check object names (not sure about this, but Alexey tended to use camel case a lot so we'd have to change all of those; and ignore the Rcpp export file.

>What is the downside to setting up the pre-commit hooks? Would each of us have to do it individually? If a contributor didn't have the pre-commit hook set up would the burn of making sure the code passes fall on them? Or would it mean that we would have to check out their branch and run the pre-commit hook?

We could have the pre-commit hook be part of the repository, but yes, we would individually have to install e.g. the `lintr` package (or in #418 the clang-format utility). Well, the burn of making sure the code passes ALWAYS falls on the contributor, right?

kdorheim:
> We could have the pre-commit hook be part of the repository, but yes, we would individually have to install e.g. the `lintr` package (or in #418 the clang-format utility). Well, the burn of making sure the code passes ALWAYS falls on the contributor, right?

True so long as it is transparent, if it obscure/unclear that doesn't really seem fair

bpbond:
Note to self: seems like the lintr step should just be combined into one of the existing yaml files? I don't see why it needs its own stand-alone check...?

bpbond:
@kdorheim Here's what a lint error looks like. Gives file, line number, and problem: 
Screen Shot 2021-02-05 at 8 04 21 AM

bpbond:
This is ready to merge afaik.

kdorheim (approved):

PR #418, opened 2021-02-04
Enforce C code style    ** DO NOT MERGE **

### LINKED ISSUE #92: Future pull requests from non-PNNL folks

### LINKED ISSUE #366: Add systematic white space tests
PR #353 removes the trailing white space that is added to scripts in different editing tools. @bpbond and myself think that it might be worth while to add some sort of github actions, unit tests, or something of that nature to prevent PRs with the trailing white space from being merged into Hector. 

Something we discussed was using https://github.com/jimhester/lintr to create a github actions issue, but this might only work for the `.R` files ideally we would want this test to also check the C++ files.

bpbond:
This branch tests using [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) to reformat files in the Hector codebase. The actual command line tool can be installed using `brew install clang-format` for example on your local machine.

ClangFormat offers a variety of built-in default styles, and you can of course modify these to define your own requirements. So far I have tried out three different styles, and in this PR:
* `forcing_component` (cpp and app) is reformatted using the [WebKit style](https://webkit.org/code-style-guidelines/)
* `SimpleNBox` is reformatted using the [Google style](https://google.github.io/styleguide/cppguide.html)
* `bc_component` is reformatted using the default [llvm style](https://llvm.org/docs/CodingStandards.html)

([Mozilla](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style) and [Chromium](https://chromium.googlesource.com/chromium/src/+/master/styleguide/c++/c++.md) styles are also available.)

So step 1 - do we prefer a particular style? Please take a look at the diffs for the files above and see if there's something you particularly like or don't. Comments welcome.
Step 2 - once we decide I'll reformat the entire codebase (are we ok with this? note rpl's comment about this in #92 )
Step 3 - ...and then enable a GitHub Action enforcing this on PRs
Step 4 - we will probably then want to set up GitHub pre-commit hooks to run `clang-format` on any modified code before it gets committed. Otherwise we'll have constantly failing PRs.

Does this sound good? Please take a look at the styles and let me know your thoughts. Thanks.

Closes #366 
Closes #92


kdorheim:
Thanks for doing this! The consistency in coding style will make the code more readable in my opinion! Comparing the three styles they all look pretty dang close to me, I don't have a particularly strong preference towards one. 


A brief glance at the documentation I get an impression that there are two levels of styles, the 
llvm and WebKit style appear to be somewhat superficial. Dealing more with spacing, braces and so on. The Google style is interesting because the includes call seems to be somewhat managed, which sounds nice especially because we currently don't do that. But apparently, this can also hide dependencies which sets off some alarm bells in my head, would this affect Hector's dependency finder?

bpbond:
@kdorheim I don't think it will affect the dependency finder, no. Are you referring to this [this part](https://google.github.io/styleguide/cppguide.html#Header_Files) of the spec?

kdorheim:
I was talking about the pro/cons list [under this part](https://google.github.io/styleguide/cppguide.html#Forward_Declarations) of the documentation.

PR #420, opened 2021-02-04
Rstyle test fail

bpbond:
This should fail the lintr test; let's see what kind of error message is available for the developer, per KD's question in #417 .


kdorheim:
@bpbond wait is this supposed to fail?

bpbond:
Yes but I fixed the issue it wasn't failing over in #417 — sorry for any confusion.

PR #423, opened 2021-02-06
New 'fluxpool' class guaranteeing fluxes and pools are always >=0

bpbond:
Hi @skygering I jumped on this quickly so you don't get delayed.

This branch implements a new `fluxpool` class, that inherits from `unitval` but imposes the restriction that _its value may not be negative_. You can do math with `fluxpool`s just like with `unitval`s, and in fact can mix and match them in equations, but the result will always be a `fluxpool`. I wrote some basic test code for the new class, temporarily located in the `simpleNbox` constructor.

All the pools and fluxes in `simpleNbox` have been changed over to use this new class, with a few exceptions:
* `atmosland_flux` is not really a flux but rather an exchange (i.e. can be positive or negative); note that this is only an output,  computed in `stashCValues()` but not used after that
* The `ffiEmissions` input data stream is still a `unitval`, so can be positive or negative. It's split into `ffi_flux_current` and `ccs_flux_current` in `simpleNbox::calcDerivs()`

~**Currently all of this is bfb** (bit for bit): Hector inputs and outputs are unchanged.~

I haven't yet changed anything in `ocean_component`, `ocean_box`, etc., but will.

In summary, all the carbon pools and fluxes in `simpleNBox` are guaranteed not to be negative. You should be able to simplify your code and life!

**Also** `simpleNbox.cpp` was over 1300 lines 😱 and I have split it into two files, hopefully sensibly but lmk.


bpbond:
I realize that I lied above: this is not bfb, because as `earth_c` is now a `fluxpool`, it is initialized to 4000 Pg C (see documentation in code) instead of zero. So its values in the output stream do change; but nothing else does.

bpbond:
This is superseded by #448

Issue #332, opened 2019-11-26
Remove onelineocean


ashiklom:
I personally have never gotten the `onelineocean` component to successfully run, and it seems to be deprecated anyway in favor of an ocean model with multiple lines. Per @cahartin 's suggestion, we should probably just remove it from the code altogether.

Some additional context:

- The component header file: https://github.com/JGCRI/hector/blob/master/inst/include/ocean_component.hpp
- The component source file: https://github.com/JGCRI/hector/blob/master/src/onelineocean_component.cpp
- Line initializing the `onelineocean` component in all cores by default (possibly its own issue): https://github.com/JGCRI/hector/blob/e97ff4ef4697dbcdddc25aed02214a29452992e5/src/core.cpp#L117
  - Related discussion on Confluence for GCAM-Hector update PR: https://stash.pnnl.gov/projects/JGCRI/repos/gcam-core/pull-requests/86/overview?commentId=28645
- #167 is a related issue

bpbond:
👍  My only comment here is that `onelineocean()` is a window into the past–Hector's past, that is. It was one of the very first things we coded.

kdorheim:
Was this replaced by DOEClim? Awesome that this is a window into Hector's first soul but even if we remove it shouldn't we be able to revisit it via version control? 🤔

bpbond:
Yes, we should be able to. It was replaced by Corinne's ocean compartment model.

PR #424, opened 2021-02-07
Eliminate all clang warnings in the Hector build

bpbond:
[Clang](https://clang.llvm.org) generates quite a few unneeded (from our pov) warnings when building Hector, which is not ideal—it makes it hard to see more serious issues. This PR
* Tells clang to ignore all warnings generated by boost headers; they're not our problem.
* Explicitly marks dead code in `oceanbox.cpp`, removing another warning.
* Adds explicit casts in `core.cpp` to get rid of the rest.

After this there are zero warnings.

### DIFF
--- a/src/carbon-cycle-solver.cpp
+++ b/src/carbon-cycle-solver.cpp
@@ -15,7 +15,12 @@
 
 #include <math.h>
 #include <string>
+
+// some boost headers generate warnings under clang; not our problem, ignore
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Weverything"
 #include <boost/numeric/odeint.hpp>
+#pragma clang diagnostic pop
 
 #include "carbon-cycle-solver.hpp"
 #include "avisitor.hpp"
--- a/src/oceanbox.cpp
+++ b/src/oceanbox.cpp
@@ -275,7 +275,7 @@ void oceanbox::compute_fluxes( const unitval current_Ca, const double yf, const 
             closs_total = closs_total + compute_connection_flux( i, yf );
         } // for i
 
-        if( 0 /* osc */ ) {
+        if( /* DISABLES CODE */ (0) /* osc */ ) {
             const double mean_past_loss = vectorHistoryMean( carbonLossHistory, 10 );
             unstable_box_flux_adjust = mean_past_loss / closs_total.value( U_PGC );
             
--- a/src/core.cpp
+++ b/src/core.cpp
@@ -557,7 +557,7 @@ void Core::registerCapability(const string& capabilityName, const string& compon
     H_ASSERT( !isInited, "registerCapability not available after core is initialized")
 
     // check whether the capability already exists
-    int ncap = componentCapabilities.count(capabilityName);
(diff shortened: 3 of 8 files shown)


kdorheim:
@bpbond I think that is a great capability, is the dead code warning from `oceanbox.cpp` informative? If so you could you post the details? I suspect that it could be useful for #332

src/carbon-cycle-solver.cpp
```diff
@@ -15,7 +15,12 @@
 
 #include <math.h>
 #include <string>
+
+// some boost headers generate warnings under clang; not our problem, ignore
+#pragma clang diagnostic push
```
kdorheim, line 20: Does the pragma call silence the warnings?

src/oceanbox.cpp
```diff
@@ -275,7 +275,7 @@ void oceanbox::compute_fluxes( const unitval current_Ca, const double yf, const
             closs_total = closs_total + compute_connection_flux( i, yf );
         } // for i
 
-        if( 0 /* osc */ ) {
+        if( /* DISABLES CODE */ (0) /* osc */ ) {
```
kdorheim, line 278: is all the code within this if statement disabled? when would we want to / need to enable it?

src/carbon-cycle-solver.cpp
bpbond: `#pragma` is an open-ended extension to the C++ preprocessor through which each compiler can implement custom functionality.

Here, the first `#pragma` says "I'm about to send you a command to push onto your stack" (where the top of the stack represents the active settings).

Second command says "ignore all warnings"

Then the `#include`

And then finally we tell clang to pop off (remove) this instruction, restoring previous behavior.

kdorheim (approved):
@bpbond looks good, only a few suggestions for documentation

bpbond:
I disabled the Travis web hook, not sure why that's still there.

src/oceanbox.cpp
bpbond: Yes. It was debugging code used by Corinne. We could remove entirely, but might want for future issues?

bpbond:
>@bpbond I think that is a great capability, is the dead code warning from oceanbox.cpp informative? If so you could you post the details? I suspect that it could be useful for #332

Sorry, missed this. It's just in a comment in the code; doesn't appear anywhere else: `        if( /* DISABLES CODE */ (0) /* osc */ `

@kdorheim

src/oceanbox.cpp
kdorheim: Wait so this is still part of the ocean component code that is used (cause there is some ocean stuff that is floating around not used or is this is) but the code in the `if statement` is helpful when debugging?

bpbond: Yes, the `if` block is helpful (was helpful, anyway) when debugging instabilities in the ocean boxes.

kdorheim: ahhh I thought that it was legacy code from back in the day pre doeclim ocean. 

If it is helpful code to use in debugging it would probably be helpful to have a comment stating that

Issue #374, opened 2020-05-22
Tracer pool


bpbond:
@dawnlwoodard and I had a cool idea: a tracer pool data type that would let us arbitrarily track where C (say) goes in the model. For example, we might want to know where all the SOC emitted over a run ends up.

This would be so cool!

It seems like it would require a tracked-pool class to handle things under the hood, so that Hector components could just mostly multiply, add, compute fluxes, etc., as they do now.

bpbond:
Also we don't think any other simple climate model has anything like this?!?

ssmithClimate:
If anyone does it would be the Bern model. (probably not the simplified version though).

kdorheim:
That's an awesome idea!

bpbond:
It would be nontrivial to implement, but we were like, wow would open the door to some super cool analyses.

kdorheim:
something kinda like this? 

https://twitter.com/RARohde/status/1131224334402555905

dawnlwoodard:
@kdorheim yes! Though of course with the ability to trace every pool, not just fossil fuels. Which means we would want to treat fossil fuel emissions like a carbon pool in terms of this capability.

bpbond:
Long drive today and I had a chance to think this through. I think actually it'll be straightforward to implement for the carbon cycle.

Basically a `tracked_pf` (tracked pool/flux) will take the place of `unitval` for the carbon cycle elements. Each `tracked_pf` is a map of origin names (key) and unitvals (values), with the keys giving the ultimate origin of the 'package' of carbon. A subtraction of two `tracked_pf` objects (or a `tracked_pf` and a unitval) debits from each pool proportionally; an addition operation means combining the maps of two objects. This setup will let us do a drop-replacement wrt the carbon cycle calculations; that code will be unchanged.

```c++
class tracked_pf {
    std::map<string, unitval> origin_map;

public:
   
    static std::string unitsName( const unit_types );
        
    tracked_pf();
    
    void set( initial, unit_types, double );

    friend tracked_pf operator+ ( const tracked_pf&, const tracked_pf& );
      // use something like set_union to calculate a joint origin map of the two tracked_pf objects
    friend tracked_pf operator- ( const tracked_pf&, const unitval& );
      // debug the second argument (a flux) proportionally from the tracked pools of the first
    friend tracked_pf operator- ( const tracked_pf&, const tracked_pf& );
      // this just gets the unitval total of the second argument and calls method above
}
```

[...]

PR #425, opened 2021-02-10
Tracking fluxpools

bpbond:
Your trackedval code from #374 reimplemented into the `fluxpool` class. Has all the functionality needed to start changing the `simpleNbox` runtime code.

### DIFF
--- a/inst/include/csv_fluxpool_visitor.hpp
+++ b/inst/include/csv_fluxpool_visitor.hpp
@@ -1,0 +1,64 @@
+/* Hector -- A Simple Climate Model
+   Copyright (C) 2014-2015  Battelle Memorial Institute
+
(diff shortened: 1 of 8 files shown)


bpbond:
OK @skygering documenting where we are:

* The branch is now correctly (I think) tracking the transfer of carbon from earth to atmosphere.

abs
rel

* Code usually follow your logic and design but I did make some changes as I went to algorithms, the output visitor's behavior, and other things.
* There's a fair amount of test code in the `simpleNbox` constructor
* There are some oddities we're going to need to consider

But anyway, for the moment, it's in a good state for you to do work I think. Keep us posted!

bpbond:
This is superseded by #448

PR #428, opened 2021-02-12
Remove deprecated oneline ocean component

### LINKED ISSUE #332: Remove onelineocean

kdorheim:
Remove the online ocean component it has been replaced with a different ocean component and has not been maintained. 

Closes #332 

**Type of development** 

* PATCH: minor changes (e.g., documentation or bug fixes)

**Checklist**
- [ NA] [Roxygen](https://roxygen2.r-lib.org/) where applicable but also thorough and coherent inline documentation 
- [NA] [Unit tests](https://r-pkgs.org/tests.html) when applicable
- [x] Complies with the [Hector code style guide](https://jgcri.github.io/hector/articles/manual/StyleGuide.html)
- [ ] The PR passes automated testing suite


bpbond (changes requested):
@kdorheim There's no longer a one line ocean log output, so R test needs to be changed.

bpbond:
This actually exposed a bug in the R test code (specifically `test_wrapper.R`), which putatively confirmed that a certain number of log files were written but didn't clear out the log directory before running the model. As a result it couldn't detect if the number of files was less than expected, or even if none were written at all.

PR #432, opened 2021-02-24
Clarify forcing names 

### LINKED ISSUE #378: Clarify forcing names.
It would be useful to consider re-naming a couple of the forcing variables to be more descriptive and not ambiguous. 

Suggest: H2O -> StratH2O and O3 -> TropO3.

(Since H2O changes in the troposphere is a feedback and, therefore, included in the climate sensitivity. And stratospheric and tropospheric O3 forcings are generally treated separately as they have different drivers and time paths.)

kdorheim:
Closes #378

Rename H2O and O3  per @ssmithClimate  suggestion:

>(Since H2O changes in the troposphere is a feedback and, therefore, included in the climate sensitivity. And stratospheric and tropospheric O3 forcings are generally treated separately as they have different drivers and time paths.)

It looks like someone already started working on this the output returns FH2O_strat and FO3_trop , unless @ssmithClimate were you thinking that the o3 component needs to be entirely renamed?

### DIFF
--- a/inst/include/component_data.hpp
+++ b/inst/include/component_data.hpp
@@ -216,8 +216,6 @@
 #define D_EMISSIONS_NOX           "NOX_emissions"
 #define D_EMISSIONS_CO            "CO_emissions"
 #define D_EMISSIONS_NMVOC         "NMVOC_emissions"
-//Do not redefine Ma though it is used for o3 component
-//#define D_EMISSIONS_O3 "O3_concentration"
 
 // oc component
 #define D_EMISSIONS_OC          "OC_emissions"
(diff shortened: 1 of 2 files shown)


ssmithClimate (approved):
This looks good!

bpbond:
@kdorheim This PR is targeting `master` ...shouldn't it be to `stay-positive` instead? If yes please change.

kdorheim:
@bpbond good catch that is correct, it was branched off of `stay_positive` but when I opened the PR I didn't direct it to there!! it should be correct now.

bpbond:
@kdorheim What do you mean someone already started work on this?

kdorheim:
RF_O3_TROP was already in use, before digging into it I was under the impression that that was something that I was going to have to change

kdorheim:
FYI the two CI tests are failing because of a lintr dependency (shakes fists at the sky)

kdorheim:
@bpbond is this ready to merge?

bpbond:
@kdorheim as far as I know yes

bpbond (commented):
Thanks @kdorheim

PR #434, opened 2021-03-25
Schedule CI runs once a month

kdorheim:
While working through a Hector UI issue with @stephpenn1 it was clear that unless we are actively merging branches into the master branch that the CI won't run and catch errors when dependencies change (something is currently up with the remotes package which was causing the errors).

This PR adds the ability to launch a CI test with the press of a button on the github page without opening a PR and launches the Hector github actions suite once on the 1st day of the month. 🤷‍♀️

### DIFF
--- a/.github/workflows/codecov.yaml
+++ b/.github/workflows/codecov.yaml
@@ -5,6 +5,12 @@ on:
   pull_request:
     branches:
       - master
+  schedule:
+   # * is a special character in YAML so you have to quote this string, it is set up to run at 6am 
+   # on the first day of the month, this is helpful for periods of time when PRs are not being merged.
+    - cron:  '0 6 1 * *'
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
 
 name: test-coverage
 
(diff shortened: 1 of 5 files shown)


bpbond:
```
  schedule:
     # * is a special character in YAML so you have to quote this string
     # Launch run at 6am the 1st of every month.
     - cron:  '0 6 1 * *'

   # Allows you to run this workflow manually from the Actions tab
   workflow_dispatch:
```

Whoah what black magic is this!

kdorheim:
looks like the weird travis holdover is holding this up... but @crvernon removed that test I belive

bpbond:
Should we merge?

kdorheim:
Sure! I've tagged you the reviewer so if you approve sounds good to me!

bpbond:
OK. My only 'concern' is that we might well want more than a once-a-month static fire test but we can adjust that as needed later.

bpbond:
Hahahaha whoops

kdorheim:
Agreed we might want to adjust how often it runs automatically. I was a tad worried about exceeding our free github actions CI minutes... also there should be a button that allows us to rerun whenever we want without having to push changes on the branch

PR #441, opened 2021-05-21
Ocean component breaks command-line Hector

kdorheim:
**Describe development**

 In #439 I exposed Hector ocean variables & parameters so that they could be accessible to the R wrapper. However, in doing so I broke command line Hector's ability to write out this information directly into csv files. This was not caught by the github actions tests because the test_hector.sh script did not fail even though the Hector core failed during spin up 
```
Thu May  6 15:04:24 2021:DEBUG:addVisitor: Core adding a visitor
Thu May  6 15:04:24 2021:NOTICE:main: Calling prepareToRun()
Thu May  6 15:04:24 2021:NOTICE:prepareToRun: Computing dependencies and re-ordering components...
Thu May  6 15:04:24 2021:NOTICE:prepareToRun: Preparing to run...
Thu May  6 15:04:24 2021:NOTICE:prepareToRun: Spinning up model...
* Program exception:
msg:  	Assertion failed: Date required for atmosphere-ocean carbon flux, high-latitude
func: 	getData
file: 	ocean_component.cpp
ffile:	ocean_component.cpp
line: 	378
```

This PR will 

* give the ocean component the ability to write output at individual years, required for printing results during spin up & with the csv output stream 
* update the command line github action to ensure that the hector cores run until completion 

**Type of development** 

* PATCH: minor changes (e.g., documentation or bug fixes)

**Checklist**
- [x] [Roxygen](https://roxygen2.r-lib.org/) where applicable but also thorough and coherent inline documentation 
- [x] [Unit tests](https://r-pkgs.org/tests.html) when applicable
- [x] Complies with the [Hector code style guide](https://jgcri.github.io/hector/articles/manual/StyleGuide.html)
- [ ] The PR passes automated testing suite

PR #442, opened 2021-06-07
testing out a test change

kdorheim:
Testing out a bash test script change. 

Notes this is not to be merged if this causes the test_hector.sh script to fail on the CI then this change should be added into #441


kdorheim:
As expected with this change the command line Hector CI test fails because of the error thrown in the test_hector.sh script. Will add this change into #441

bpbond:
👍

PR #443, opened 2021-06-07
Tier 1 doeclim challenges 

kdorheim:
**Describe development**

The first part of the Hector-DOEclim coupling rework 

* Expand on the documentation in the temperature component (no expected change in temp)
* Make the conversion factors consistent (no expected change in temp)
~* Change the HL & LL temperatures returned to the output visitor as relative (should only change the HL and LL temp)~ --> see #444 this will be dealt with latter 
* Remove the TEMP EQ variable, it is a duplicate of global mean temp 

**Type of development** 

This specific PR is probably a **MINOR**  change 

**Checklist**
- [x] [Roxygen](https://roxygen2.r-lib.org/) where applicable but also thorough and coherent inline documentation 
- [x] [Unit tests](https://r-pkgs.org/tests.html) when applicable
- [x] Complies with the [Hector code style guide](https://jgcri.github.io/hector/articles/manual/StyleGuide.html)
- [x] The PR passes automated testing suite


src/temperature_component.cpp
```diff
@@ -224,54 +224,66 @@ void TemperatureComponent::prepareToRun() {
         C[i] = 0.0;
     }
 
-    // DOECLIM parameters calculated from constants set in header
+    // Dependent DOECLIM model parameters, based on constants set in the header
+    // Constants & conversion factors
+    kcon = 3155.0;               // conversion factor from cm2/s to m2/yr; TODO do we want to use secs_per_Year / 10000 to be consistent?
```
kdorheim, line 229: @bpbond part of me wants to change this to `secs_per_Year / 10000` so that it is consistent with the seconds per year but this does end up changing the output, here is the MSE between the old and new output per variable. It seems like a trivial change in output to me, @bpbond what do you think? 

```
# A tibble: 5 x 2
  variable      MSE
  <chr>       <dbl>
1 Ca       1.38e- 6
2 FCO2     1.60e-10
3 Ftot     1.60e-10
4 heatflux 6.28e- 9
5 Tgav     2.73e- 9
```

inst/include/temperature_component.hpp
```diff
@@ -72,27 +72,27 @@ class TemperatureComponent : public IModelComponent {
...
-    const double bsi = 1.3;               // warming factor for marine surface air over SST (due to retreating sea ice)
-    const double cal = 0.52;               // heat capacity of land-troposphere system, W*yr/m2/K
-    const double cas = 7.80;              // heat capacity of mixed layer-troposphere system, W*yr/m2/K
-    const double flnd = 0.29;             // fractional land area
-    const double fso = 0.95;              // ocean fractional area below 60m
+    const double ak = 0.31;               // slope in climate feedback - land-sea heat exchange linear relationship (W/(m2K))
+    const double bk = 1.59;               // offset in climate feedback - land-sea heat exchange linear relationship,(W/(m2K))
+    const double csw = 0.13;              // specific heat capacity of seawater (Wyr/(m3K))
+    const double earth_area = 5100656E8;  // (m2)
+    const double secs_per_Year = 60.0 * 60.0 * 24.0 * 365.2422; //  secs. * min. * hrs. * tropical calendar days (seconds)
```
kdorheim, line 81: In the ocean component 

https://github.com/JGCRI/hector/blob/master/src/ocean_component.cpp#L203
`double time = 60*60*24*365.25;  // seconds per year`

we use a different number of seconds per year different by 11 minutes... do we care that they are different? should they be consistent?

src/temperature_component.cpp
bpbond: What's MSE here? Yes, looks trivial, and I agree consistency would be nice.

kdorheim: mean squared error

inst/include/temperature_component.hpp
bpbond: Probably we don't care enough  🤷‍♂️

src/temperature_component.cpp
bpbond: Definitely trivial!

kdorheim:
@bpbond I know that we spoke about changing the HL & LL output from being returned as absolute to an anomaly as part of this PR but looking into it, to do this we actually need the capabilities of #439. I played around with reimplementing them here but it requires multiple lines of code changes and I anticipate a nightmare of merge conflicts. What do you think about addressing that in a different issue after the stay positive branch is merged into master?

bpbond:
Oh, I think that's totally fine.

src/temperature_component.cpp
```diff
@@ -224,54 +222,66 @@ void TemperatureComponent::prepareToRun() {
         C[i] = 0.0;
     }
 
-    // DOECLIM parameters calculated from constants set in header
+    // Dependent DOECLIM model parameters, based on constants set in the header
```
bpbond, line 225:
```suggestion
    // DOECLIM model parameters, based on constants set in the header
```

inst/include/temperature_component.hpp
```diff
@@ -72,27 +72,27 @@ class TemperatureComponent : public IModelComponent {
     void setoutputs(int tstep);
 
     // Hard-coded DOECLIM parameters
-    const int dt = 1;                     // years per timestep (this is implicit in Hector)
+    double dt = 1;                     // years per timestep (this is implicit in Hector)
```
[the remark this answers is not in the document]
kdorheim: I changed it because in the `temperature_component.cpp` any time dt was called it appeared as `double(dt)` which seemed redundant I figured that if dt was created as a double then we could avoid all of those extra calls. 

Should I left all the double calls in the `temperature_component.cpp`, I was just trying to streamline the code/make it easier to read. As a new a cpp coder I am not sure if that was the right move or not so feedback welcome. 

Also Hector complies and runs with `const in dt = 1` without the all the double calls in the `temperature_component.cpp`  is that the correct move?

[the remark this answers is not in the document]
bpbond: The change to a double (so eliminating all those `double(dt)` usages) is fine; I'm just asking why not declare this as `const double dt = 1;`.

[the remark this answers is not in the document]
kdorheim: That's a great idea!

src/rcpp_constants.cpp
```diff
@@ -1316,20 +1316,13 @@ String VOLCANIC_SCALE() {
 return D_VOLCANIC_SCALE;
 }
 
-//' @describeIn temperature Global mean temperature
+//' @describeIn temperature Global mean temperature anomaly
```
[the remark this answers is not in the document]
kdorheim: I always thought that Tgav was a temperature anomaly in https://github.com/JGCRI/hector/wiki/OutputFiles 

temperature | Tgav | degC | global atmospheric temperature anomaly
-- | -- | -- | --

and then that is why the ocean component has to convert to absolute temp cause everywhere else Hector is operating in temp anomaly right?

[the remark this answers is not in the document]
bpbond: Right. It needs a preindustrial mean (15C) to add to the anomaly to arrive at the current global absolute temp.

kdorheim:
okay to merge?

bpbond:
Looks good to me! This is merging into your branch—go for it.

PR #445, opened 2021-06-18
Medium Fixes

kdorheim:
This PR is for the medium tier of fixes as part of the doeclim rework. 

Our main objective here is to make sure that the temperature component and the forcing component are using consistent RF values for CO2. As of right now, these components use two hardcoded values. 

__Forcing__
[forcing component](https://github.com/JGCRI/hector/blob/fc59802a79228ad24a87ba394ad8cfde8a2dff58/src/forcing_component.cpp#L281)
```
 // This is identical to that of MAGICC; see Meinshausen et al. (2011)        
unitval Ca = core-&gt;sendMessage( M_GETDATA, D_ATMOSPHERIC_CO2 );        
if( runToDate==baseyear )            
     C0 = Ca;       
     forcings[D_RF_CO2 ].set( 5.35 * log( Ca/C0 ), U_W_M2 );
```
The value 5.35 is being used as the radiative forcing effiency for CO2 this comes from Meinshausen et al. 2011
Screen Shot 2021-06-18 at 9 30 43 AM

__Temperature__
Whereas the [temperature component](https://github.com/JGCRI/hector/blob/fc59802a79228ad24a87ba394ad8cfde8a2dff58/inst/include/temperature_component.hpp#L82) uses  
Screen Shot 2021-06-18 at 9 46 25 AM

__So What?__
What is the difference between these two values? 

if alpha CO2 = 5.35 then Q2CO2 = 3.71 (forcing component)
vs. 
if Q2CO2 = 3.7 then alpha CO2 = 5.34 (temperature component)

It seems small, hopefully, the difference of 0.01 will have little to no impact on Hector's output. Anyways because we are going to start changing the RF efficiencies these should be consistent with one another, pulling from a single value instead of using two different hard-coded values. 

```
    const double q2co = 3.7;              // radiative forcing for atmospheric CO2 doubling (W/m2)<br class="Apple-interchange-newline">
```


kdorheim:
Here is how I have been going about this 

1) make the forcing component use a RF efficiency that is read in via the ini file 
2) have the temperature component use the RF efficiency to determine the q2co.

kdorheim:
Progress & Problems 

So far step 1 has been successful, I've been able to read in an alpha CO2 value from the ini file and it reproduces "old hector"

image

However, I am running into issues with step 2, when I am trying to have the temperature component use the RF efficiency to determine the q2co. 

image

When I've played around with debugging the temperature component in Xcode it seems like the problem is that aCO2 is being initiated but the value is not actually passed to the component :(

kdorheim:
@bpbond can you take a look at the temperature component and see if there is something obviously wrong with how I am reading the aco2 value in? I based it on how the ocean component reads in data from other components.

src/temperature_component.cpp
```diff
@@ -226,6 +228,15 @@ void TemperatureComponent::prepareToRun() {
     // Constants & conversion factors
     kcon = secs_per_Year / 10000;               // conversion factor from cm2/s to m2/yr;
     ocean_area = (1.0 - flnd) * earth_area;    // m2
+    
+
+    aCO2 = core->sendMessage( M_GETDATA, D_ACO2 );
```
kdorheim, line 233: This is how the ocean component reads in tgav 🤷‍♀️, but aCO2 ends up being 0 which causes q2co to be 0

kdorheim:
@bpbond I _think_ I might have figured out what was wrong with the aCO2 value being passed between components so please hold off on this for now

bpbond:
@kdorheim OK so I will ignore this for now :)

kdorheim:
@bpbond Alright! So passing in a parameter so that the forcing component and temperature component use consistent information for the CO2 RF is now functional! However it with this new dev we are failing the old new tests (the new development causes some changes to the model output) please see below for some diagnostic information about the output variables that are being affected. 

Looking at the outputs returned by the new Hector (using consistent CO2 RF) vs old Hector it is not obvious what is different. 
image


If we plot the percent change between old Hector and new Hector 

1. the difference between the output increased with run time 
2. the difference between the emission-driven runs is larger than the concentration driven runs, which makes sense to me 

image


Looking at the mean percent difference between the two output streams, there is less than a 0.1% difference between the two output streams. That seems trivial to me @bpbond what do you think? 

```
   variable scenario                      percent
   <chr>    <chr>                           <dbl>
 1 Ca       hector_rcp45_constrained.ini  2.20e-9
 2 Ca       hector_rcp45.ini              3.60e-3
 3 FCO2     hector_rcp45_constrained.ini  6.42e-5
 4 FCO2     hector_rcp45.ini              8.48e-3
 5 Ftot     hector_rcp45_constrained.ini -2.40e-6
 6 Ftot     hector_rcp45.ini              5.23e-3
 7 heatflux hector_rcp45_constrained.ini  3.47e-2
 8 heatflux hector_rcp45.ini              4.06e-2
 9 Tgav     hector_rcp45_constrained.ini  3.01e-2
10 Tgav     hector_rcp45.ini              3.63e-2
```

I am pretty convinced that change has a trivial effect on the Hector output streams what says you @bpbond?

bpbond:
I'm confused. The absolute-value graphs above shows Ca different by ~45 ppm, but the percentage calculation reports 2.20e-9 and 3.60e-3, which doesn't seem consistent.

kdorheim:
So on the absolute plots there are 4 colored lines, 2 for the emission driven and 2 for the concentration driven runs. There we don't want to compare the difference between emission and concentration but the difference between the runs of the same type (emission or concentration driven). But those lines are on top of each other (and poorly chosen colors) so you may have to zoom in to see the lines on top of each other which is why I also plotted the change between the runs.

bpbond:
Thanks! 👍  I understand and agree—trivial changes, let's just update the comparison data.

bpbond:
Look at all those lovely green checks!

src/temperature_component.cpp
```diff
@@ -162,7 +168,7 @@ void TemperatureComponent::setData( const string& varName,
         } else if( varName == D_DIFFUSIVITY ) {
             H_ASSERT( data.date == Core::undefinedIndex(), "date not allowed" );
             diff = data.getUnitval(U_CM2_S);
```
ssmithClimate, line 170: Variable names are a bit cryptic sometimes. Consider  perhaps diff -> ocean_heat_diff so that the code is easier to understand. (and perhaps S -> clim_Sens).

ssmithClimate (approved):
Great to have this consistency there. This will likely save some headaches in the future! 

I added a couple comments to consider.

inst/include/component_data.hpp
```diff
@@ -54,6 +54,8 @@
 #define D_RF_SO2                D_RF_PREFIX "SO2"
 #define D_RF_VOL                D_RF_PREFIX "vol" // forcing from volcanic so2
 #define D_RF_halocarbons        D_RF_PREFIX "halocarbons"
+#define D_ACO2                  "aco2"      // effective radiative forcing
```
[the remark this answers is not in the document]
kdorheim: I think it is the scalar value we use to get CO2 radiative forcing so I think you are right it is forcing efficiency. The terminology related to forcing is important but it is also something I easily get mixed up on. Thanks!

inst/input/hector_rcp26.ini
```diff
@@ -144,6 +144,7 @@ baseyear=1750		; when to start reporting; by definition, all F=0 in this year
 ; Optional radiative forcing constraint
 ; If supplied, the model will use these data, ignoring what it calculates
 ;Ftot_constrain=csv:constraints/MAGICC_RF_8.5.csv
+aco2=5.35		; radiative forcing, W/m2
```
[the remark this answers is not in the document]
kdorheim: No I don't think so, I think that this would also be forcing efficiency but it is used to calculate the forcing per CO2 doubling that is used in the temperature component.

src/temperature_component.cpp
```diff
@@ -10,6 +10,10 @@
  *
  *  Created by Ben Vega-Westhoff on 11/1/16.
  *
+ * Meinshausen, M., Raper, S. C. B., and Wigley, T. M. L.: Emulating coupled atmosphere-ocean
```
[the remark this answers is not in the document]
kdorheim: I added it because in line 239 we use an equation to get the forcing per CO2 doubling from the radiative forcing from forcing efficiency for CO2. But good point we should also add the citations for DOECLIM.

src/temperature_component.cpp
kdorheim: Agree that the parameter names can be confusing, however that might be a to-do to address post Hector v3

src/temperature_component.cpp
```diff
@@ -97,11 +107,6 @@ void TemperatureComponent::init( Core* coreptr ) {
     tgav_constrain.allowInterp( true );
     tgav_constrain.name = D_TGAV_CONSTRAIN;
 
-    // Define the doeclim parameters
```
kdorheim, line 100: @bpbond because these parameters will be read in via the ini file setting them here is seems redundant, okay to delete or is there some reason to keep these lines of code?

kdorheim:
@bpbond & @ssmithClimate I've responded to Steve's first round of feedback. Please take a look and let me know if more changes are necessary or if this PR is good to go 👍. Thanks!

src/temperature_component.cpp
bpbond: OK to delete afaik

src/temperature_component.cpp
```diff
@@ -10,6 +10,16 @@
  *
  *  Created by Ben Vega-Westhoff on 11/1/16.
  *
+ * DOECLIM is based on
```
bpbond, line 13: This is great @kdorheim — thank you

bpbond (approved):

src/temperature_component.cpp
ssmithClimate: Looks great.

PR #447, opened 2021-06-30
Optimize github actions 

kdorheim:
**Describe development**

Optimize the github actions to minimize the # of action minutes by switching from running tests on mac platform to a Linux platform. 

**Type of development** 

* PATCH: minor changes does not affect Hector code. 

**Checklist**
~- [] [Roxygen](https://roxygen2.r-lib.org/) where applicable but also thorough and coherent inline documentation~
~- [] [Unit tests](https://r-pkgs.org/tests.html) when applicable~
~- [] Complies with the [Hector code style guide](https://jgcri.github.io/hector/articles/manual/StyleGuide.html)~
- [ ] The PR passes automated testing suite

### DIFF
--- a/.github/workflows/rcmd.yml
+++ b/.github/workflows/rcmd.yml
@@ -27,12 +27,10 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          # for latest stable version of R for MacOSX
-          - {os: macOS-latest, r: 'release'}
+          # test on one version of R for MacOSX
           - {os: macOS-latest, r: '3.6'}
 
-          # for windows stable versions
-          - {os: windows-latest, r: 'release'}
+          # test on one version of R for windows
           - {os: windows-latest, r: '3.6'}
 
           # latest stable version of R for ubuntu-16.04 is 3.5; ubuntu-latest is migrating to 20.04 by the end of 2020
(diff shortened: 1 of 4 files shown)


bpbond:
Oh wow that's an obvious one to change! Thank you @kdorheim

bpbond (approved):

.github/workflows/rcmd.yml
```diff
@@ -27,12 +27,10 @@ jobs:
       fail-fast: false
       matrix:
         config:
-          # for latest stable version of R for MacOSX
-          - {os: macOS-latest, r: 'release'}
+          # test on one version of R for MacOSX
           - {os: macOS-latest, r: '3.6'}
```
bpbond, line 31: I feel like if we're going to only test on one version, it should be the 'release', not '3.6'?

kdorheim: 👌

bpbond:
@kdorheim The problem was that R's `curl` package couldn't be installed because there was no underlying curl library available (something about Ubuntu versus Mac defaults I guess). This has been fixed in 721429fc4 by installing it before proceeding with the R installation.

kdorheim:
@bpbond thanks for addressing the issue with the lintr test!

PR #449, opened 2021-07-05
Rework makefiles

bpbond:
Hector's makefiles were unclear, had lots of unneeded cruft, and didn't completely work (e.g. the `testing` target wouldn't be invalidated by changes to unit testing code). This PR heavily reworks them for clarity and concision.

### DIFF
--- a/.github/workflows/unit-testing.yaml
+++ b/.github/workflows/unit-testing.yaml
@@ -1,6 +1,6 @@
 # Build the gtest Hector and run it
 
-name: googletest Hector
+name: Hector unit tests
 
 # Controls when the action will run.
 on:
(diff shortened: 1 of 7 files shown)


.github/workflows/unit-testing.yaml
```diff
@@ -1,6 +1,6 @@
 # Build the gtest Hector and run it
 
-name: googletest Hector
+name: Hector unit tests
```
kdorheim, line 3: To clarify, these unit tests are different that the R hector test that unit tests correct?

src/testing/Makefile
```diff
@@ -1,17 +1,29 @@
...
 
+## This (SRCS) will include testing_main.cpp
 SRCS	= $(wildcard *.cpp)
-OBJS	= $(SRCS:.cpp=.o)
-DEPS	= $(SRCS:.cpp=.d)
-
 
--include $(DEPS)
+## ** TEMPORARY **
+## the following files aren't working yet; remove them
```
kdorheim, line 7: So when they are working we will remove them from the undone list so that they will actually run?

kdorheim (approved):
🙏 this is great! thank you for looking at these, Hector's makefiles were really confusing!

.github/workflows/unit-testing.yaml
kdorheim: Looking through the rest of the PR this is pretty obvious... 🤦‍♀️

src/testing/Makefile
bpbond: @kdorheim Right. I have created #450 to track this.

PR #455, opened 2021-07-17
Fix Xcode filerefs, vignette rebuilds, and test failures

bpbond:


bpbond:
Debugging notes

OK, when a Hector core is run and reset, this error occurs:
```
Error in reset(core) : 
  Error resetting to date= 0 :  msg:  	Assertion failed: tracking mismatch: ocean_c and ?
func: 	operator+
file: 	fluxpool.hpp
ffile:	../inst/include/fluxpool.hpp
```
It happens in the first timestep of the spinup of the reset core, and is being thrown in simpleNBox::stashCValues(), in this line:
```
ocean_model_c = ocean_model_c - oa_flux + ao_flux;
```
I don't quite understand why, but suspect that it's because Skylar's variable ocean_model_c isn't getting reset in simpleNBox::reset() unlike all other variables.

Ah...I think because it's after 1850, the variables are tracking, but because there's no real ocean_model_c reset it stays tracking after that...which it shouldn't. Okay.

I could try to fix it there; but this is a hack anyway, right? Better to implement tracking in the ocean model?

But...that sounds hard. Maybe a useful first step is to implement the reset functionality for her proxy variable, get that to work, and then move on later to dealing with the ocean model.

PR #457, opened 2021-07-18
Carbon tracking code upgrades

### LINKED ISSUE #467: Tracking upgrades
As I finalize #466 , some thoughts:

- [x] We'll want a tracking vignette
- [x] Remove the diff stuff in the tracking code (after checking the need for it has been addressed)
- [x] Ideally the user can use a new function `fetchtracking()` to get the tracking output data from within R
- [x] Right now the tracking output file looks like this; rethink at all?

year | pool_name | pool_value | pool_units | source_name | source_fraction
-- | -- | -- | -- | -- | --
1746 | Diff | -0.6125 | U_PGC | veg_diff | 0
1746 | Diff | 15.7918 | U_PGC | soil_diff | 0

bpbond:
* The `ocean_model_c` variable used for carbon tracking code was hard-coded to a specific value. Changed to use the real value returned from the ocean component.
* Allow setting the start-tracking date from the INI file. I put this in the core's section of the file, not that of simpleNbox. We may want to track more things in the future, so it should be globally accessible?
* Added the `trackDate` variable to the RCP4.5 INI file (not any of the others).
* ~~Included tracking start date in the core information reported by R (see below). Not that I have _not_ allowed the user to set this from within R, as you can't change `startDate` or `endDate` either from there.~~ Would welcome your thoughts here @kdorheim and @skygering 

```
> library(hector)
> core <- newcore(ini_file)
> 
> core
Hector core:	unnamed hector core
Start date:	1745
End date:	2300
Current date:	1745
Tracking date:	1850
Input file:	/Library/Frameworks/R.framework/Versions/4.1/Resources/library/hector/input/hector_rcp45.ini
```

Closes #467


bpbond:
@kdorheim I believe you have reviewed all PRs that went into this one, so am merging into `land-tracking` (v3) and closing #374 .

Cheers @skygering !

PR #458, opened 2021-07-19
Fix unit handling in sendmessage

### LINKED ISSUE #327: Invalid units in `sendmessage` are silently converted to `U_UNDEFINED`
Currently, passing bogus units to a Hector core from the R interface (via `setvar -> sendmessage`) will silently convert them to `Hector::U_UNDEFINED`:

https://github.com/JGCRI/hector/blob/master/src/rcpp_hector.cpp#L307-L309

This should probably at least throw a warning, right? Or should it be an outright error, as the commented code suggests?

dawnlwoodard:
@mbins and I added behavior to stop if invalid units passed to sendmessage in rcpp_hector.cpp when setting data, and to set to unitless if invalid or missing units passed when getting data. Also fixed invalid units in biome.R (it was using PgC instead of Pg C). One thing to look at though is that we hard coded the string for "setData" into the if statement - not sure if there's a better way to do it that C++ recognizes. 

Closes #327

### DIFF
--- a/src/RcppExports.cpp
+++ b/src/RcppExports.cpp
@@ -5,6 +5,11 @@
 
 using namespace Rcpp;
 
+#ifdef RCPP_USE_GLOBAL_ROSTREAM
+Rcpp::Rostream<true>&  Rcpp::Rcout = Rcpp::Rcpp_cout_get();
+Rcpp::Rostream<false>& Rcpp::Rcerr = Rcpp::Rcpp_cerr_get();
+#endif
+
 // GETDATA
 String GETDATA();
 RcppExport SEXP _hector_GETDATA() {
--- a/tests/testthat/test_set_get_data.R
+++ b/tests/testthat/test_set_get_data.R
@@ -60,5 +60,5 @@ for (v in emissions) {
 }
 
 test_that("Setting variable with invalid unit throws an error", {
-  expect_error(setvar(hc, year, VEG_C(), 500, "boogedyboo"))
+  expect_error(setvar(hc, year, VEG_C(), 500, "boogedyboo"), regex = "invalid unit")
 })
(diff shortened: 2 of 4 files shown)


dawnlwoodard:
@bpbond not sure about backward compatibility but seems like it should be? And yes, should be easy enough to write a test for it. I'm still new to tests but I can try adding something later today.

kdorheim:
> Awesome! Thank you for doing this. A couple quick questions:
> 
> * It looks like it's 100% backward compatible?
> * Would it be possible to write a test for this functionality?
> 
> @kdorheim Is there any reason we would want this targeted to the v3 branch instead of `master`? I don't think so?

My vote would be master!

bpbond:
Sounds good—thank you and thank you @mbins . Let me know if you need any help with the testing code step.

src/RcppExports.cpp
```diff
@@ -5,6 +5,11 @@
 
 using namespace Rcpp;
```
bpbond, line 6: This is an auto-generated file that we should probably add to `.gitignore`...

tests/testthat/test_set_get_data.R
```diff
@@ -58,3 +58,7 @@ for (v in emissions) {
     expect_equal(fetchvars(hc, year, v)[["value"]], val)
   })
 }
+
+test_that("Setting variable with invalid unit throws an error", {
+  expect_error(setvar(hc, year, VEG_C(), 500, "boogedyboo"))
```
bpbond, line 63: Let's make sure we're triggering correct error:
```r
  expect_error(setvar(hc, year, VEG_C(), 500, "boogedyboo"), regex = "invalid unit")
```

bpbond (changes requested):
Looks great—thank you @dawnlwoodard @mbins !

tests/testthat/test_set_get_data.R
dawnlwoodard: ooh thanks - that's super helpful to know about

bpbond (approved):

PR #460, opened 2021-07-20
Expose rh as time series

dawnlwoodard:
So, annoyingly more complicated than I hoped because rh wasn't already accessible as a time series, but I eventually got it doing what I wanted. It passes all tests and I can successfully get the values I want over the whole model period with fetchvars from the R interface. 

This pull request updates several functions in simpleNBox-runtime.cpp (sum_rh, rh, rh_fda, and rh_fsa) along with their definitions in simpleNBox.hpp so that they can take time as an argument, but also behave correctly when no time argument is given. The updates essentially allow the set of rh functions to parallel the behavior of the sum_npp and npp functions.  The pull request also includes some of the more typical changes needed to expose a variable to the R interface.


bpbond (approved):
Looks good @dawnlwoodard ! Thanks for doing this.

Tests need to pass of course :)

dawnlwoodard:
> Looks good @dawnlwoodard ! Thanks for doing this.
> 
> Tests need to pass of course :)

Weird. It builds fine and passes tests on my computer. I thought based on the last pull request that I didn't need to commit the RcppExports files but those are the only difference between my repository and what I pushed here. Not sure what else would be breaking everything. I'll take a look later today.

dawnlwoodard:
@bpbond or @kdorheim if you have a sec this week to consult or point me in the right direction for fixing these check issues, that'd be lovely (no rush though). I'm just not totally sure where to start since it all works fine on my machine.

bpbond:
@dawnlwoodard When in doubt, rebuild the f***ing documentation 😄

dawnlwoodard:
> @dawnlwoodard When in doubt, rebuild the f***ing documentation 😄

hahaha duly noted. Thanks for the rescue! Glad it worked and could be merged.

PR #464, opened 2021-07-21
Remove Outdated Files

kdorheim:
As part of Hector v3 we want to clean up the Hector repository so that it contains functional/maintained code. So in this PR I've identified some files that I believe we can get rid of, since we are removing code it could be considered a code-breaking change, for users that are looking for these files they will still be available with previous versions of the Hector releases (the versions that they are applicable to) & in the future we will be using the metadata repository approach for manuscript materials. 

Things removed here: 
* old/svg_log.log - commit history for Hector before we. migrated to using git for version control 
* scripts - a pile of scripts & data files used in the Hartin et al. 2015


bpbond (approved):
👍

PR #474, opened 2021-07-28
Split FFI into two separate inputs, FFI and DACCS

bpbond:
For @natanelha work, split fossil fuel industrial emissions input into two, both of which must be positive: `ffi_emissions` and `daccs_uptake` (direct air carbon capture and storage).


src/simpleNbox.cpp
```diff
@@ -94,6 +96,7 @@ void SimpleNbox::init( Core* coreptr ) {
 
     // Register the inputs we can receive from outside
     core->registerInput(D_FFI_EMISSIONS, getComponentName());
+    core->registerInput(D_DACCS_UPTAKE, getComponentName());
```
kdorheim, line 99: Should DACCS also be registered as a capability?

bpbond: I don't think so; FFI and LUC aren't registered that way. It's an input.

kdorheim: Ah I thought that everything that passed into the getdata also had to be registered as a capability. Sound good.

kdorheim (commented):
Ben I think it looks pretty good, I am curious about two things. 

1.  How to the daccs is determined, is it just assumed that there is no daccs until after emissions become net negative? Does it even matter to hector if for some period of time there is a combination of emissions & daccs going on at the same time? That might not matter for the Hector csv inputs but I suspect that might be a scenario we see coming out of GCAM. 
2. What happens when running under constrained CO2 concentrations? I presume that the residual between the constraint and atmospheric co2 is dumped into the deep ocean is that correct?

kdorheim (approved):

bpbond:
>How to the daccs is determined, is it just assumed that there is no daccs until after emissions become net negative? 

No, now that we've split the time series, you could have DACCS while you also have emissions. Right now, in the 2.6 file, that's not true, but see #475 

>What happens when running under constrained CO2 concentrations? I presume that the residual between the constraint and atmospheric co2 is dumped into the deep ocean is that correct?

Yes, that functionality is unchanged for now.

Thanks!

ssmithClimate:
In the IAMC protocol CCS is split out from positive emissions. (And we used this in the gridded CO2 emissions as well - the net negative CO2 was distributed over agriculture land.) So this is information that should be readily available from the main IAMs.

PR #477, opened 2021-07-29
Return tracking data from C++ to R 

### LINKED ISSUE #469: Returning tracking data to R
Re 
>Ideally the user can use a new function fetchtracking() to get the tracking output data from within R
in #467 -

How would this work? It's a data frame, not a single time series. Right now the tracking visitor is printing all its data to a CSV file; could it instead (or also) be printing to a `stringbuf` that is then returned and run through `read.csv()` in R?

We'd need a new overload of `sendMessage` that returns the buffer I guess?

Thoughts @pralitp @kdorheim @dawnlwoodard ?

bpbond:
This PR:
* Adds a `csvBuffer` string stream variable to the CSVFluxpoolVisitor. Tracking data is written to this, and then copied to the file output at the end of the run
* `core::mkcore()`, which is called by R to instantiate a Hector core, adds a fluxpool visitor to each created core
* The core now calls all visitors before running to let them record core pointer and tracking date
* A new `core::TrackingData()` function looks for the csv fluxpool visitor, gets its buffer, and returns that string to R

As a result now you can do
```r
core <- newcore("inst/input/hector_rcp45.ini")
run(core)
get_tracking_data(core)
```

It's a little hacky — I would like to talk with you @pralitp about the dynamic cast usage — but it works.

@natanelha

Closes #469


bpbond:
Still need to add tests, documentation, etc., so marking this as draft for now.

bpbond:
@leeyap You can push a new version of this branch with your work when ready

Issue #484, opened 2021-08-04
Changing the tracking date via R


bpbond:
OK, how do we handle tracking date changes in R?

Possibilities include:
* INI file only
* a dedicated `set_tracking_date()` function
* as a parameter to both the R `run()` and C++ `core::run()` - this would follow the `runtodate` logic
* as a parameter to the R `run()`, which then uses `setData()`
* via `setvar()`

If you have time, next week I would like to talk this through with you @kdorheim .

kdorheim:
👍 want to put something on our calendars?

PR #485, opened 2021-08-04
Carbon tracking vignette

leeyap:
Hi @bpbond,

Here is a draft of the carbon tracking vignette. Please let me know any edits/suggestions or if I missed anything. Thanks!

### DIFF
--- a/vignettes/carbon_tracking.Rmd
+++ b/vignettes/carbon_tracking.Rmd
@@ -14,9 +14,14 @@ knitr::opts_chunk$set(echo = TRUE)
 
 ## Introduction
 
-This tutorial will walk through Hector's carbon tracking feature and show how to run the tracking command and display the results.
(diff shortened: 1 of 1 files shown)


bpbond:
@leeyap Can you change the base so that this PR is going to `tracking-vars`, not `master`? Thanks.

bpbond:
Hi @leeyap -

Looks great! I have made some suggested edits for language and clarity and just pushed them (seemed easier); feel free to revert if something doesn't make sense to you.

There is one thing that will need to change — Kalyn and I are going to decide how we want to be able to set the tracking date programmatically from R (as opposed to editing the INI file; see #484 . Once that's done, we'll want to update the vignette to use that mechanism.

leeyap:
@bpbond - the changes look good, thank you for that! 

Re: changing the date in R, that sounds good. Keep me posted on that and I'll update appropriately.

bpbond:
Merging this so it's available over in #488

PR #487, opened 2021-08-10
Return full ocean tracking information to atmosphere

bpbond:
We were just using the net ocean-atmosphere flux to do tracking, which meant that nothing from the ocean was making it into the atmosphere (in tracking terms) once anthropogenic forcing took hold. This PR fixes that, and as a bonus, simplifies and clarifies the code! 

oceanfix

See https://github.com/NatanelHa/Ha-Hector-Internship/issues/6


bpbond:
Thanks @dawnlwoodard . I agree, although there's only ~1 PgC/yr going from ocean to atmosphere, and that's from the surface boxes (obviously) so will have plenty of atmosphere-origin C in it...anyway, agreed, we will want to sanity-check further.

PR #488, opened 2021-08-11
Allow getting/setting tracking date from R

### LINKED ISSUE #484: Changing the tracking date via R

bpbond:
After discussion with @kdorheim about best way to implement this (see #484  ). This involved some changes to the core so that `sendMessage()` handles things correctly.

Now you can do this:
```
> library(hector)
> core <- newcore("inst/input/hector_rcp45.ini")
> fetchvars(core, NA, TRACKING_DATE(), "")
  scenario year     variable value      units
1            NA trackingDate  9999 (unitless)
> setvar(core, NA, TRACKING_DATE(), 1850, "")
> fetchvars(core, NA, TRACKING_DATE(), "")
  scenario year     variable value      units
1            NA trackingDate  1850 (unitless)
> run(core)
Hector core:	Unnamed Hector core
Start date:	1745
End date:	2300
Current date:	2300
Input file:	/Users/d3x290/Dropbox/Documents/Work/Code/hector/inst/input/hector_rcp45.ini
> head(get_tracking_data(core))
  year pool_name pool_value pool_units  source_name source_fraction
1 1850        HL    140.600       Pg C         deep     0.002546630
2 1850        HL    140.600       Pg C intermediate     0.200533000
3 1850        HL    140.600       Pg C           HL     0.522017000
4 1850        HL    140.600       Pg C           LL     0.270039000
5 1850        HL    140.600       Pg C      atmos_c     0.004864410
6 1850        LL    784.167       Pg C         deep     0.000976032
> shutdown(core)
Hector core (INACTIVE)
```

Closes #484


bpbond:
@leeyap Can you push changes to your vignette and testing code so that they use and test this mechanism? Thanks.

vignettes/carbon_tracking.Rmd
```diff
@@ -32,19 +33,28 @@ library(hector)
...
+fetchvars(core, NA, TRACKING_DATE(), "")
+```
+
+This returns the variable `trackingDate` with a `value` of 9999. This default value does not turn carbon tracking on (because the model never reaches year 9999).
+
+Now, we will set a `trackingDate`, check that it worked, and run a core. Let's begin tracking in 1750.
+
+``` {r set.date}
+setvar(core, NA, TRACKING_DATE(), 1750, "")
+fetchvars(core, NA, TRACKING_DATE(), "")
```
bpbond, line 56: @leeyap great—thanks!

bpbond:
@leeyap Now that we can turn tracking on and off, we don't need the `trackingcore()` function in the test code; please remove it, along with the `hector_rcp45_tracking.ini` file, and use the setvars mechanism to enable tracking when needed.

bpbond:
@leeyap When you push, re comment above, please tag Kalyn so she knows to review. Thanks.

leeyap:
@bpbond will do. Hope to get it done tomorrow.

leeyap (approved):
Looks good to me. @kdorheim , could you take a look and make sure the vignette and test file are good to go? Thanks!

bpbond:
Just to clarify - Kalyn, please review the entire PR if you have time; I'd like any feedback you may have on how I implemented this. Thanks!

vignettes/carbon_tracking.Rmd
```diff
@@ -19,6 +19,7 @@ This vignette will walk through Hector's carbon tracking feature and show how to
 Hector's carbon tracking feature helps to trace the movement of carbon between various pools as the model runs. At the user-chosen start date, each pool is
 deemed to be composed of 100% carbon from itself, i.e. at that moment in time 
 the deep ocean pool is 100% composed of carbon from the deep pool. 
+
```
kdorheim, line 22: @leeyap not sure why I cannot comment directly on this line but at https://github.com/JGCRI/hector/pull/488/files#R5 the ymal header is for a regular markdown not a vignette. 

it should read something like 

```
output: rmarkdown::html_vignette
vignette: >
  %\VignetteEncoding{UTF-8}
  %\VignetteIndexEntry{Example: Carbon tracking with Hector}
  %\VignetteEngine{knitr::rmarkdown}
editor_options: 
  chunk_output_type: console
```

tests/testthat/test_tracking.R
```diff
@@ -1,25 +1,17 @@
 context("Carbon tracking")
```
kdorheim, line 1: @bpbond thoughts about adding carbon tracking to the old-new test?

R/RcppExports.R
```diff
@@ -40,7 +40,13 @@ LL_SEVERE <- function() {
     .Call('_hector_LL_SEVERE', PACKAGE = 'hector')
 }
 
-#' @describeIn emissions Black carbon emissions (\code{"Tg"})
+#' @describeIn core Start of carbon tracking (Year)
```
kdorheim, line 43: It should read 
```
#' @describeIn parameters Start of carbon tracking (Year)
```

there is no Roxygen documentation object called `core` so right now `help(TRACKING_DATE)` does nothing. It should work though after you make this change and rebuild the documentation.

kdorheim (changes requested):
I think it looks pretty good but does need a few changes.

@bpbond 2 things to consider with regards to the unit tests something to think about is if we want to add a tracking data old new test to make sure that the tracking behavior doesn't change with updates? Or is it safe to assume that so long as the total amount of carbon in the pools doesn't change then the tracking will not change either? 

Another thought, do we want to change a carbon cycle parameter such as beta & ensure that when a parameter value changes the carbon tracking output also changes?

R/aadoc.R
```diff
@@ -238,3 +238,16 @@ NULL
 #' @name temperature
 #' @family capability identifiers
 NULL
+
```
kdorheim, line 241: Please remove the `Identifiers for quantities in the core` chunk of code.

tests/testthat/test_tracking.R
bpbond: 🤔 

I don't _think_ it's necessary? Here's my logic:
* As you note below, `oldnew` already covers carbon totals; it seems unlikely that the fractions could vary without the totals varying
* Tracking behavior (specifically, `fluxpool` behavior) is already covered by C++ unit test

bpbond: On a related note:

>do we want to change a carbon cycle parameter such as beta & ensure that when a parameter value changes the carbon tracking output also changes?

It's hard for me to imagine a case when this wouldn't be true, following the logic above. But I will add a test doing this.

R/aadoc.R
bpbond: 👍

R/RcppExports.R
bpbond: 🎉 It does work now! Thank you :)

tests/testthat/test_tracking.R
kdorheim: ah seems unnecessary sounds like it is good as written! 👍

kdorheim (approved):
🥇

PR #490, opened 2021-08-12
Create test_hector.R

leeyap:
New test code to expand Hector coverage to 100%

@bpbond - let me know what you think or if it should be more comprehensive. Thanks!

### DIFF
--- a/tests/testthat/test_hector.R
+++ b/tests/testthat/test_hector.R
@@ -1,19 +1,9 @@
-context("Expansion of test code to include all files")
+context("Test code for hector.R")
 
 inputdir <- system.file("input", package = "hector")
 inifile <- file.path(inputdir, "hector_rcp45.ini")
 
-test_that("Test that invalid input variable returns error", {
-
-    # Test that if the length of var_split is >2, error is returned
-    core <- newcore(inifile)
-    expect_error(setvar(core, NA, "global.permafrost.beta", 10, ""),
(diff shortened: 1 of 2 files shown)


tests/testthat/test_hector.R
```diff
@@ -0,0 +1,70 @@
+context("Expansion of test code to include all files")
```
bpbond, line 1: This isn't really a context, or at least not a useful one. If we need to break this file into multiple files so that there's a single coherent context, let's do that please.

tests/testthat/test_hector.R
```diff
@@ -0,0 +1,70 @@
...
+
+inputdir <- system.file("input", package = "hector")
+inifile <- file.path(inputdir, "hector_rcp45.ini")
+
+test_that("Test that invalid input variable returns error", {
```
bpbond, line 6:
```r
test_that("Invalid input variable returns error", {
```

tests/testthat/test_hector.R
bpbond: Or change context to "Miscellaneous" if that's the best course?

bpbond (changes requested):

bpbond (changes requested):
See suggested `context` changes but otherwise looks good!

bpbond (approved):

PR #493, opened 2021-08-18
Carbon tracking - final work

### LINKED ISSUE #467: Tracking upgrades

bpbond:
* Spent a while investigating the big diff values for atmosphere, detritus, and soil that occur, and I'm satisfied that this is a function of a small pool with high turnover; we have the solver values to adjust to, and that seems fine. Removing all the 'diff' related temporary code.
* Added `component` as a column in the tracking data output
* Expanded and clarified some documentation points

Closes #467

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hector
 Title: The Hector Simple Climate Model
-Version: 2.5.0
+Version: 3.0.0
 Authors@R: c(person("Robert", "Link", 
             email = "robert.link@pnnl.gov", 
             role = "aut",
@@ -21,6 +21,10 @@ Authors@R: c(person("Robert", "Link",
 	        email = "pralit.patel@pnnl.gov", 
 	        role = "ctb",
 	        comment = c(ORCID = "0000-0003-3992-1061")),
+	     person("Leeya", "Pressburger", 
+	        email = "leeya.pressburger@pnnl.gov", 
+	        role = "ctb",
+	        comment = c(ORCID = "0000-0002-6850-2504")),
 	     person("Alexey", "Shiklomanov",
 	        email = "alexey.shiklomanov@nasa.gov",
 	        role = "ctb",
--- a/src/simpleNbox-runtime.cpp
+++ b/src/simpleNbox-runtime.cpp
@@ -232,7 +232,7 @@ void SimpleNbox::stashCValues( double t, const double c[] )
     fluxpool oa_flux = omodel->get_oaflux();
     fluxpool ao_flux = omodel->get_aoflux();
     
-    // Land-use change uptake from atmosphere to veg, detritus, and soil
+    // Land-use change emissions and uptake between atmosphere and veg/detritus/soil
     fluxpool luc_e_untracked = luc_emission(t, in_spinup) * yf;
     fluxpool luc_u_untracked = luc_uptake(t, in_spinup) * yf;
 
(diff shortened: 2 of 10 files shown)


bpbond:
@leeyap Could you review the changes to your vignette, as well as the `get_tracking_data()` documentation?

@kdorheim Ready for your review.

Thanks!

leeyap (approved):
Looks good to me. Thanks, @bpbond!

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: hector
 Title: The Hector Simple Climate Model
-Version: 2.5.0
+Version: 3.0.0
```
kdorheim, line 3: EPPPP!!! 🥳

src/simpleNbox-runtime.cpp
```diff
@@ -244,20 +244,16 @@ void SimpleNbox::stashCValues( double t, const double c[] )
     const fluxpool npp_total = sum_npp();
     const fluxpool rh_total = sum_rh();
 
-    // TODO: If/when we implement fire, update this calculation to include it
-    // (as a negative term).
     // BBL-TODO this is really "exchange" not a flux
```
kdorheim, line 247: When are these TODOs going to be addressed? Is there a github issue related to them?

DESCRIPTION
```diff
@@ -21,6 +21,10 @@ Authors@R: c(person("Robert", "Link",
 	        email = "pralit.patel@pnnl.gov", 
 	        role = "ctb",
 	        comment = c(ORCID = "0000-0003-3992-1061")),
+	     person("Leeya", "Pressburger", 
+	        email = "leeya.pressburger@pnnl.gov", 
+	        role = "ctb",
+	        comment = c(ORCID = "0000-0002-6850-2504")),
 	     person("Alexey", "Shiklomanov",
```
kdorheim, line 28: Add Skylar!

vignettes/carbon_tracking.Rmd
```diff
@@ -75,13 +77,16 @@ each model time point after `trackingDate`.
...
 ```
 
-The `tdata` dataset now contains information on year, pool name, amount of 
-carbon in each pool, and the fraction of each source for that pool.
+The `tdata` dataset now contains information on year, the model component,
+the name of the pool within that component, the amount of 
+carbon in the pool, and the fraction (0-1) of each source for that pool. 
+The source fractions for a given pool and year will always sum to 1.
+
+Side note: the names in the carbon tracking data (`pool_name`, `pool_value`,
```
kdorheim, line 87: Isn't 'component' also a name returned from the model output?

kdorheim (changes requested):
It is really coming together! So exciting! A few minor comments & a few questions that I occurred to me while reviewing this PR, I don't necessarily know if they even need to be addressed or if this PR is the appropriate place to do so. 

1) setting the tracking var date tells us when we start tracking the carbon, how do users know what date is used in the final comparison? When fetch_tracking is called is it querying the core's current date? Should that information be returned with the data frame? 
2) is there a way to launch the tracking from the ini file? thinking about the cpp users does it return a csv file to the output dir?

src/simpleNbox-runtime.cpp
bpbond: I don't think we have a GitHub issue but it'll be related to dealing with #397 . I will add as an issue too though.

DESCRIPTION
bpbond: Doh! Thanks

bpbond: @skygering Do you have an ORCID ID?

vignettes/carbon_tracking.Rmd
bpbond: Yes; I'm just talking about the three that overlap with the model output but differ in their names.

bpbond: Clarifying

bpbond:
@kdorheim Thanks for the review!

>How do users know what date is used in the final comparison?

The first date in the tracking output will be the same as `trackingDate`. I'll add a note to the vignette clarifying this.

>is there a way to launch the tracking from the ini file?

Yes, it's a core-level parameter there. Adding a note clarifying this.

DESCRIPTION
bpbond: Moving this to an issue

bpbond:
@kdorheim Great comments—thank you. I think this is ready to merge once checks pass.

bpbond:
...but of course will wait for your approval 😄

kdorheim (approved):

Issue #497, opened 2021-08-19
No parameters to inherit


bpbond:
When building the R package this is printed:
```
Warning: Topic 'parameters': no parameters to inherit with @inheritParams
* Topic 'TRACKING_DATE': no parameters to inherit with @inheritParams
* Topic 'PREINDUSTRIAL_CO2': no parameters to inherit with @inheritParams
* Topic 'BETA': no parameters to inherit with @inheritParams
* Topic 'Q10_RH': no parameters to inherit with @inheritParams
* Topic 'WARMINGFACTOR': no parameters to inherit with @inheritParams
* Topic 'F_NPPV': no parameters to inherit with @inheritParams
* Topic 'F_NPPD': no parameters to inherit with @inheritParams
* Topic 'F_LITTERD': no parameters to inherit with @inheritParams
* Topic 'F_LUCV': no parameters to inherit with @inheritParams
* Topic 'F_LUCD': no parameters to inherit with @inheritParams
* Topic 'ECS': no parameters to inherit with @inheritParams
* Topic 'AERO_SCALE': no parameters to inherit with @inheritParams
* Topic 'VOLCANIC_SCALE': no parameters to inherit with @inheritParams
* Topic 'DIFFUSIVITY': no parameters to inherit with @inheritParams
Warning: Topic 'methods': no parameters to inherit with @inheritParams
* Topic 'format.hcore': no parameters to inherit with @inheritParams
* Topic 'print.hcore': no parameters to inherit with @inheritParams
```

@kdorheim Do you know is this legitimate, or can we remove all the Roxygen `@inheritParams` tags and get rid of this?

kdorheim:
I think we can remove the `@inheritParams` tags.

bpbond:
@leeyap Do you mind taking this on?

leeyap:
@bpbond Sure thing. Remind me what file(s) need to be edited?

bpbond:
aadoc.R? Sorry,not at computer. You should be able to search for the occurrences above though. 

Thanks @leeyap!

bpbond:
@leeyap No that's wrong. The lines are in `hector.R` and `rcpp_constants.cpp`

bpbond:
Also, after editing you will need to rebuild the package documentation (Build -> Documentation).

leeyap:
Okay, thanks. Just to clarify, I should strike the whole line, eg `//' @inheritParams BETA` would all be deleted

bpbond:
Right!

PR #498, opened 2021-08-20
Issue edits

leeyap:
Hopefully this should fix the @inheritParams issue in #497 - removed tags from hector.R and rcpp_constants.cpp


bpbond:
@leeyap please change the PR base to whatever you branched from, hopefully `land-tracking`?

PR #502, opened 2021-08-23
Permafrost

bpbond:
Infrastructure: permafrost data structures, output, etc. Passes all tests.

Not integrated into runtime C cycle yet.


bpbond:
@dawnlwoodard points out [this new paper](https://www.nature.com/articles/s43247-022-00498-3) - we may want to update the default warning factor in the v3 permafrost?

PR #527, opened 2021-10-28
leeyabot

leeyap:
Work for `leeyabot` :robot:

----

First set of commits: initial prep to integrate script. Made changes to `comp_data` to include all years and necessary variables and modified `test_old-new` to include years previously specified in `comp_data`.

Next set of commits (once first round is approved) will be the code for `leeyabot` itself. This includes a yaml file that allows for automatic commenting and an R markdown file that actually runs the code.

----

`leeyabot` :robot: is automated to comment on a pull request to show the changes that a PR makes to the current version of the model. It does so by creating a summary table of R squared and normalized RMSE values and returns a graph of differences between model outputs from different versions of Hector.
 
Currently, it looks at RCP 4.5 (to match the values in the `comp_data` csv file) and the variables `atmos_c`, `Ftot`, and `Tgav` for the latest version of Hector up against v3.0 (again to match the csv data). It can easily look at additional variables, and I have code to allow it to run for multiple scenarios. @bpbond and I also talked about including a graph of the outputs themselves instead of the differences for different versions of Hector (I have this code as well). 

I can adjust to get data/figures for the differences between v2.5 and v3.0 for reference, if necessary.

@bpbond (and @kdorheim, if you have time) - feedback welcome and appreciated!


bpbond:
Starting to look good @leeyap ! A few requests:

- [ ] Please remove Hector output messages from the bot comment
- [ ] Round numbers in table to a sensible number of digits (if you're using `kable` I believe it has a parameter for this)
- [ ] Right now the graph has a static title, right? We probably want to change that. For the stored compdata, we can have the first line give the version number used to generate the compdata (meaning that any code that reads that file's will need to do `skip = 1`). Your Rmd can then extract that info by doing `readLines(compdatafile, n=1)` and insert it into the graph title. Make sense?
- [ ] Why do the graphs above look like they do? Shouldn't they be zero lines right now?

scripts/OutputDifferences.Rmd
```diff
@@ -0,0 +1,130 @@
...
+differences <- differences %>% select(scenario, year, variable, units, version, diff)
+```
+
+
+```{r summary-table, echo= FALSE}
+
+# Organize data
+output %>%
+    select(value) %>%
+    rename(latest = value) -> new_values
```
bpbond, line 60: FYI: you can do 
```r
select(latest = value)
```
in a single step.

bpbond:
Could you paste in a `leeyabot` output showing what happens when there's a significant divergence? For example, add 1 to all the Hector outputs before doing your comparisons. I'd just like to see an example if possible. Thanks

leeyap:
diff_plot

@bpbond - Here's a plot when the reference data was multiplied by 2.

bpbond:
@leeyap Thanks. Is the graph title auto-generated at this point, or is it static? If the latter I think we will want to change.

scripts/OutputDifferences.Rmd
```diff
@@ -0,0 +1,129 @@
...
+
+```{r summary-table, echo= FALSE}
+
+# Organize data
+output %>%
+    select(latest = value) -> new_values
+
+both_versions <- compdata %>%
+    select(year, variable, value) %>%
+    rename(v3.0.0 = value) %>%
```
kdorheim, line 63: Instead of naming v3.0.0 what about naming it something like `compdata_value` because `leeyabot` will not always be comparing data to v3.0.0 in the near future, it will be using Hector v3.5.0. Making that column name slightly more generic would help avoid confusion as Hector continues to develop.

scripts/OutputDifferences.Rmd
```diff
@@ -0,0 +1,129 @@
...
+error <- as.data.frame(apply(table, MARGIN = 2, RMSE))
+rownames(error) <- c("atmos_c", "Ftot", "Tgav")
+
+# Find mean of data to normalize
+mean3 <- by(both_versions$v3.0.0, both_versions$variable, mean)
+mean3 <- as.data.frame(c(mean3[1], mean3[2], mean3[3]))
+
+# Divide RMSE by means
+NRMSE <- as.data.frame(error[,1] / mean3[,1])
+rownames(NRMSE) <- c("atmos_c", "Ftot", "Tgav")
```
kdorheim, line 104: Instead of hard coding the strings "atmos_c", "Ftot", and "Tgav" consider using the R Hector functions `ATMOSPHERIC_C(), GLOBAL_TEMP(), RF_TOTAL()` that way if there are any changes in the strings on the cpp side of things this script won't fail.

scripts/OutputDifferences.Rmd
```diff
@@ -0,0 +1,129 @@
...
+
+``` {r graphs, echo = FALSE, fig.width = 12, fig.height = 6, message = FALSE}
+# Just the numerical differences
+diff_plot <- ggplot(differences, aes(year, diff, color = variable)) +
+    geom_hline(yintercept = 0, alpha = 0.5) +
+    geom_line() +
+    facet_grid(variable~scenario, scales = "free") +
+    scale_color_manual(labels = c("atmos_c (Pg C)", "Ftot (W/m2)", "Tgav (degC)"), values = c("gold", "aquamarine4", "slateblue4")) +
+    ggtitle("Differences in Hector outputs in current version relative to v3.0.0") +
+    labs(x = "Year", y = "Difference in value", col = "Variable") +
```
kdorheim, line 121: I think BBL might have already mentioned this but change the `ggtitle` so that the relative to version x is dynamic. 
Related to that I would make the `sacle_color_manual` settings dynamic as well. 

the code would look something like this 
```
color_scheme <- c("gold", "aquamarine4", "slateblue4") 
names(color_scheme) <- unique(differences$variable)

ggplot() + 
    blah blah plot code() + 
    scale_color_manual(values = color_scheme) + 
    blah blah the end of of plot code 
```

kdorheim (changes requested):
@leeyap I think it looks great! My one comment would be to make more of the code dynamic, which is somewhat annoying but it will help make the code robust to version changes in the future. 

* the one required change would be to make sure that v 3 is not hard coded in the script
* highly encourage that the hector variable names are not hard coded in but ultimately okay if they stay that way

bpbond:
@leeyap It looks like `outputDifferences.html` accidentally got committed. Can you remove this please? Otherwise it looks good to go.

leeyap:
@bpbond Done, thanks!

bpbond (approved):

bpbond:
@kdorheim This is holding on your approval.

kdorheim (approved):
@leeyap great work! this looks awesome

bpbond:
Great work @leeyap 👏

PR #534, opened 2021-11-22
leeyabot updates

### LINKED ISSUE #533: Possible leeyabot behavior change
Currently [leeyabot](https://github.com/JGCRI/hector/pull/527) always prints a message, the summary table, and the graph. But when all the R2 values are 1 and all the NRMSE values are 0, i.e. the model behavior hasn't changed, is there any point? 

Maybe in that circumstance the bot should just say "The current pull request's outputs do not differ from master" and stop. 🤷

leeyap:
Adjusted code so that `leeyabot` does not return a summary table or graph if output differences are zero, but rather to simply print a message. 

Closes #533 (@bpbond)


bpbond (approved):

bpbond:
@leeyap OK to merge?

leeyap:
@bpbond Okay to merge!

PR #535, opened 2021-11-22
Incorporate 'asis' chunk output for cleaner appearance

bpbond:
This may not work 😃 but thought it was worth trying.


leeyap (approved):
Nitpicky stylistic thing - can we remove the extra space at the end of the comment? Otherwise, looks good to me 👍

PR #543, opened 2022-01-05
Switch to SSPs

kdorheim:
Migrate from the RCPs to the SSPs with the input materials and update the tests and the vignettes to use the ssps instead of the rcps. 

some questions 
1. misc RF do we want it to be included in the hector default scenarios? the misc rf includes bc on snow, contrails,  solar rf. The misc RF is included in AR6 but not in CMIP. 
2. to improve the emission-driven CO2 RF we had to adjust beta but this causes lots of failures with the biome tests. Should we leave beta as is (my hunch is that this is the right decision) or do we want to update beta to be consistent with AR6?


bpbond:
Hi @kdorheim 

0. This PR should be to `land_tracking`, not `master`?
1. I would this yes?
2. Hmm, why does changing beta cause tests to fail? That's odd. I'm not sure about this one.

kdorheim:
@bpbond 
0. good catch! this is why not to open PRs last thing before bed 
1. I think it just depends if we want default Hector scenarios to be consistent with AR runs or with the scenarios from RCMIP/CMIP6. 
2. The beta to match the CO2 RF from AR6 is 0.13, which is much lower than the current default of 0.36. Because of the way that the tests were written that check to see if the temperature is warmer now fail because of the changes made to Hector's carbon cycle. I think that updating the default parameters should probably be done in jointly in a systematic manner but knowing that until it is updated concentration forced Hector and emission forced Hector will produce very different outputs from one another and that when we finally do make the changes to beta we will have to revisit the biome tests.

kdorheim:
We will not be fully consistent with AR6 until we have updated the historical emissions to use the CEDS time series (which hopefully is something we can accomplish during the hector hackathon). 

I am inclined to leave the hector default scenarios to be consistent with the RCMIP scenarios since that is the MIP that most directly relates to Hector as a simple climate model. We could provide users with an AR6 historical ini & input files so that if they choose to they can add future information if they want to do a AR6 SSP run  (could also add all the possible AR6 historical + future SSPs to hectordata) Providing users will all the different possible scenarios would be a lot of work & has the potential to bloat the Hector repository. 

This being said we could always consult with one of our more scenario-focused colleagues (Claudia, Brian, Steve, ect.)

bpbond:
Touching base with one of them—probably Steve?—sounds good. I am happy to trust your inclination, though.

kdorheim:
@bpbond the `command-line.yaml` aka test_hector.sh script is going to fail on ssp370 at year 2268 and ssp485 will fail at year 2185 because the earth pool goes negative. We could limit the run until 2100 in the ini file to avoid this problem. Or could we increase the size of the earth c pool?

bpbond:
Hi @kdorheim I think artificially limiting the run time is not the best approach. I'm sorry, feeling slow this morning: _why_ does it go negative, when it didn't before with e.g. the high emissions of rcp85? 

Anyway, it seems like increasing the size of `earth_c` will be the way to go. It should be effectively infinite, after all.

kdorheim:
Re increasing the size of the earth pool - that would be a good plan but see next answer

Re why did we not run into the issue with rcp85 - that is a good question and I did some digging. Well, it looks like the emissions tables only had data out until 2100, however, the ssp585 and ssp370 were extrapolating the emissions to run out until 2300 was must have been causing the earth c pool to go negative. Expanding the input tables so that they continue to 2300 prevents this error from being thrown. Should Hector throw an error if the run to end of data is larger than the last date in the input files?

ssmithClimate:
In the absence of a specific instruction (for example setting a flag to extend data as constant if not available), if you're trying to run a model and haven't provided data throwing an error seems appropriate.

bpbond:
Agreed and yes on the error. We should allow extrapolation _within_ emissions time series, but not past their ends.

kdorheim:
> Agreed and yes on the error. We should allow extrapolation _within_ emissions time series, but not past their ends.

Would this be a change to the interpolation function? Or would we want to add some sort of check like https://github.com/JGCRI/hector/blob/a75702ba5a3ab9684a4eae1e49ef221172b6a494/src/core.cpp#L429 here? that will throw an error? Lastly is this something that we would want to address now or open it as a post-Hector v3 issue?

bpbond:
@kdorheim Simpler than that. Time series have settings on whether they're allowed to interpolate or not; should be a one-line change. I can assist.

kdorheim:
@bpbond that would be awesome! 

Then I think that the outstanding issue with the PR is the issue of the default run, I need to touch base with @ssmithClimate about it.

bpbond:
@kdorheim what I think you want is something like

```c++
ffiEmissions.allowPartialInterp( true );  // don't allow interpolation beyond the end of the time series
```
in the simpleNBox `init()` function.

kdorheim:
@bpbond & @crvernon any clue why the ubuntu R-CMD runs aren't launching?

bpbond:
>Can't find any online and idle self-hosted runner in the current repository, account/organization that matches the required labels: 'ubuntu-16.04'
>Waiting for a runner to pick up this job...

That's a new one to me. 🤷 

Googling:

>I have fixed this issue by adding the runner name that I added on my yml file at "runs-on" to the idle runner labels!

???

bpbond (approved):

PR #550, opened 2022-01-21
Replace old ubuntu os requests for RCMD tests

pralitp:
Switch the OS to `ubuntu-20.04` which currently has the "latest" R at 4.1 and still supports testing 3.6.

Note:  I originally considered using the `ubuntu-latest` tag instead of specifying a version to ensure we always test on the latest but there is a line inherited form the upstream config which I wasn't sure how it should be updated in such a case:
```
Rscript -e 'writeLines(remotes::system_requirements("ubuntu", "20.04"))'
```

### DIFF
--- a/.github/workflows/rcmd.yml
+++ b/.github/workflows/rcmd.yml
@@ -37,9 +37,9 @@ jobs:
           # test on one version of R for windows
           - {os: windows-latest, r: 'release'}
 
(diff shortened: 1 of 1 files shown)


bpbond:
Oh great thank you @pralitp !

bpbond (approved):
@kdorheim OK to merge? Thank you Pralit!

PR #551, opened 2022-01-21
Cleanup tracking data output

### LINKED ISSUE #483: Dynamic cast usage
As mentioned in #477 the core rifles through the visitor list to the csvfluxpool one, using `dynamic_cast` to check for a non-null pointer. This is tacky and should probably be reimplemented using virtual functions in the `aVisitor` class.

pralitp:
Cleanup collecting tracking data output by adding an `outputTrackingData` method to `AVistor` which only `CSVTrackingVisitor` will implement.  Never the less it is a cleaner solution than having to dynamic_cast.  This closes #483

### DIFF
--- a/inst/include/avisitor.hpp
+++ b/inst/include/avisitor.hpp
@@ -58,6 +58,12 @@ public:
     virtual void reset( const double reset_date ) {}
 
     //------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


bpbond (approved):
Thanks @pralitp !

PR #552, opened 2022-01-21
Auto-reset cores

### LINKED ISSUE #546: If cores auto-reset they should notify user?
When you run a core; change a parameter; and re-run (without resetting) it seems like the core auto-resets because it's dirty? I think so? But in that case it really should print a `message` telling the user it's done so.

leeyap:
See notes in #546 

This PR fixes the issue where a Hector core would auto-reset without telling the user. Now, if a dirty core is auto-reset before a run, a warning message prints. 

Also updated all relevant tests to reset before running a core in order to avoid triggering the warning.

Closes #546


bpbond:
Thanks @leeyap ! note conflict to fix 👷‍♀️

bpbond (approved):

PR #558, opened 2022-01-28
Enable tgav constraint and fix test_hector.sh bugs

### LINKED ISSUE #557: Tgav constraint
@kdorheim asks
>Do we know if the Hector tgav constraint actually works?

I am opening this issue mostly for the pleasure of closing it.

bpbond:
The tgav constraint was doubly broken: first, it had been disabled during the DOECLIM integration and never re-implemented; and second, the tgav constraint test in `test_hector.sh` wasn't working correctly.

This PR:
*  Updates `test_hector.sh` to run all INI files found in `./inst/input`, instead of using hard-coded names. We also now use `xargs` to do this, instead of `find exec`, which means that if Hector errors the script will error correctly too
* Fixes the `sed` calls so that they allow for optional whitespace after the semicolon
* Adds a check after each `sed`, confirming that the INI file was changed
* Re-implement the temperature constraint. We use the user-supplied Tgav value to back-calculate the `temp_landair` and `temp_sst` normally calculated by DOECLIM
* Adds a package test for containing temperature through R
* Minor things: fix a missing file in the Xcode project; fix the README in `./output/`

tgavs

Closes #557


kdorheim:
@bpbond is this ready for review or wait until the tgav constraint code is re-implemented?

bpbond:
@kdorheim Please wait until I do that.

bpbond:
@kdorheim Ready for your review

src/temperature_component.cpp
```diff
@@ -431,51 +426,58 @@ void TemperatureComponent::run( const double runToDate ) {
...
+    } else {  // Handle the initial conditions
         temp_landair[0] = 0.0;
         temp_sst[0] = 0.0;
     }
     temp[tstep] = flnd * temp_landair[tstep] + (1.0 - flnd) * bsi * temp_sst[tstep];
 
+    // If the user has supplied temperature data, use that instead
+    if( tgav_constrain.size() && runToDate <= tgav_constrain.lastdate() ) {
+        H_LOG( logger, Logger::WARNING ) << "** Overwriting temperatures with user-supplied value" << std::endl;
+        temp[tstep] = tgav_constrain.get( runToDate );
```
kdorheim, line 464: Add a comment here describing what is going on in `temp_landair ` and `temp_sst`.

kdorheim (approved):
Looks great Ben! Would it be useful to have the capability for R hector users to implement the constraint through in R or should it remain a feature to implement via the ini file?

bpbond:
Wait why didn't pushing 314b628 trigger tests? 😕

bpbond:
@kdorheim Could you remove your approval on this PR? I don't understand why tests aren't running.

kdorheim (changes requested):
does this trigger the tests?

bpbond:
@kdorheim Thanks! Not sure what happened, but yes tests did rerun. Just pushed a commit I hope will fix the test failure and then ready to merge from my pov.

bpbond:
Hurray! Back to you @kdorheim

kdorheim (approved):
Beautiful! Thanks for adding the R functions!

PR #548, opened 2022-01-14
Ensure OUTPUT directory exists

pralitp:
Refactor the logger util method `chk_logdir` into h_util instead and rename it to `ensure_dir_exists` so that we can use it to ensure the `OUTPUT_DIRECTORY` exists as well.  

Note, this really only pertains to command line hector, no from R.

This closes #115

PR #560, opened 2022-02-04
Update project.pbxproj

bpbond:
Update Xcode file to reflect #548

### DIFF
--- a/project_files/Xcode/hector.xcodeproj/project.pbxproj
+++ b/project_files/Xcode/hector.xcodeproj/project.pbxproj
@@ -7,8 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		270FA7802791BF2F00E6A50B /* nh3_component.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 270FA77F2791BF2F00E6A50B /* nh3_component.cpp */; };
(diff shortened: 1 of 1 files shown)


project_files/Xcode/hector.xcodeproj/project.pbxproj
```diff
@@ -7,8 +7,6 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		270FA7802791BF2F00E6A50B /* nh3_component.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 270FA77F2791BF2F00E6A50B /* nh3_component.cpp */; };
-		270FA7812791BF2F00E6A50B /* nh3_component.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 270FA77F2791BF2F00E6A50B /* nh3_component.cpp */; };
```
kdorheim, line 11: Is there a reason why the nh3_componet is being removed? Will this cause issues?

kdorheim: well good news is that I was able to check out this branch and build & run via Xcode

kdorheim (approved):

project_files/Xcode/hector.xcodeproj/project.pbxproj
bpbond: OK that's good. I see nh3_component.cpp so not sure why the deletion above.

PR #561, opened 2022-02-04
Replace custom code to echo to screen with boost.iostreams

pralitp:
Replace the custom std::streambuffer in Logger (which was used to provide the ability to echo logging output to both screen and file) with boost::iostreams

### DIFF
--- a/inst/include/logger.hpp
+++ b/inst/include/logger.hpp
@@ -15,7 +15,7 @@
  */
 
 #include <iostream>
-#include <fstream>
+#include <boost/iostreams/filtering_stream.hpp>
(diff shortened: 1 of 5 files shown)


bpbond (approved):
>to both screen and file

And I assume boost::iostreams can do this?

Looks good @pralitp — thanks.

pralitp:
> And I assume boost::iostreams can do this?

Yup!

Oops, I didn't realize we still have C++ unit tests.  I'll need to update those wrt these changes.  Will do so now.

bpbond (approved):

PR #565, opened 2022-02-16
Leeyabot tweak

### LINKED ISSUE #564: Leeyabot tweaks
Looking at the leeyabot post in #562 
68747470733a2f2f61737365742e636d6c2e6465762f383036343135393238343436323539393165396333383963383337306539353234653962623235333f636d6c3d706e67

First, the y-axis is in scientific notation, which isn't very handy. Should we change?

Second, the 1x3 layout isn't a super efficient use of space. I wonder if we should use the R Hector 'big 4' (defaults returned by `getvars`) and make a 2x2 grid instead.

Thoughts @leeyap @kdorheim ?

leeyap:
Minor changes to `leeyabot`:

- y axis is no longer in scientific notation
- Variables of interest are now the four Hector defaults
- Adjusted plot slightly to accommodate the four variables

Note that when using `facet_wrap` instead of `facet_grid` to get a 2x2 grid, I removed the `scenario` argument from the call because the graph looked cluttered with two headings. If we ever want to look at more scenarios, we may need to switch back to `facet_grid`. There might also be a more clever way that I'm not thinking of.

Closes #564


bpbond:
Thank you @leeyap . What do you think—is this useful? An improvement, or not? @kdorheim

bpbond:
To be clear, directing the question to you both.

What about plotting the runs themselves—2 lines per facet, not just the difference? I'm not trying to make extra work for anyone ( 👀 @leeyap ) just wondering about alternatives.

kdorheim:
Including the big 4 is helpful. While the difference between Hector runs is a useful diagnostic it is hard to conceptualize what a 0.1 change means plotting the two lines against one another could be useful. Summarizing the percent difference min, max, mean could potentially be helpful 🤷‍♀️

leeyap:
I think when we originally were working on `leeyabot`, we had been plotting the two lines (comparison and update) on the same graph, but because some of the changes were so small, it just looked like one line. 

I'm not sure why the summary table with statistics isn't showing up here 🤔 but theoretically, it would be easy enough to add additional metrics.

leeyap:
@bpbond @kdorheim 

Well (in case you couldn't tell by the 7000 commits) I can't seem to figure out why the summary table values are different here than they are on my machine. At least it's showing up in the comment now!

Also, in better news, here is a graph showing the actual values of variables and how they differ. I multiplied the comparison data values by two to get the dramatic difference there. 

Let me know if there are any changes you'd like me to make. Thanks!

leeyap:
...another potential issue:

I was looking at the comparison dataset against a Hector output run, and the values for `FCO2` are different. Not sure if that's okay, but I thought there shouldn't be a difference between `land-tracking` and `comp_data`.

leeyap:
...another potential issue:

I was going through the code in more detail and I noticed that some values for `FCO2` differ between the latest version (which I think is just v3, I'm branched from `land-tracking`) and the comparison data (also v3). All other variables are the same. Not sure if this is okay, just seemed weird to me.

bpbond:
It is definitely weird.

leeyap:
Yes, it should be!

leeyap:
Closing this mess of a PR - see #573 for the same changes with less clutter!

PR #573, opened 2022-03-01
Trying to fix leeyabot again

leeyap:
Same changes as #565:
"y axis is no longer in scientific notation
Variables of interest are now the four Hector defaults
Adjusted plot slightly to accommodate the four variables"

Also made some general cleanup comments and code fixes.

Fixed the code that was making the summary table not appear in comments.

Hopefully this avoids the weirdness of #565 where my machine wasn't in sync with GitHub.


bpbond (approved):
@leeyap 👏 👏  way to go

PR #576, opened 2022-03-11
Tiny leeyabot fix

leeyap:
I didn't realize, but when I submitted my PR last week to update `leeyabot` I had commented out the `emo` package (the script wouldn't knit locally if I didn't). I forgot to change it back before #573 was merged, sorry about that 🤦‍♀️


bpbond:
Poor little `leeyabot`!

🤖

PR #581, opened 2022-03-15
Ensure consistency between atmos_c and Ca

### LINKED ISSUE #459: atmos_c and Ca equal but offset by a year
Bug discovered by @NatanelHa (screenshot of results output and code to reproduce are below): atmospheric carbon (atmos_c) and atmospheric CO2 (Ca) outputs in Hector are equal (after unit conversion) but offset by 1 year. So, for example, the year 2000 value of atmos_c matches the year 2001 value of Ca/2.13. atmos_c seems to be the one correctly aligned with the fluxes - since the way the code is set up, the net flux to the atmosphere in year 2100 corresponds to the difference between the atmospheric C in 2100 from 2099. But seems like we should take a look to see which is correct and then make sure that the fluxes are aligned with that one. 

As far as I can tell this is because Ca is set in slowparameval, which happens first and uses the previous year, whereas everything else gets set in stashCValues, which happens after the solver runs and the year is updated to the new year (based on the run method in carbon_cycle_solver.cpp). 

So if everything else is correct and the output is the only issue, we could potentially just change this line "Ca_ts.set(t, Ca);" in record_state in simpleNBox.cpp to "Ca_ts.set(t-1, Ca);" but seems worth thinking through if there are other consequences of how it's currently working and making sure that other things aren't also offset in a way they shouldn't be.

As a side note (should this be a separate issue?), the fluxes don't exactly match the difference in the atmosphere values from year to year. Seems like it would be nice to have these cleanly match up. Is this a solver-related thing? 

Other side node: it doesn't seem intuitive that the ATMOSPHERIC_C variable that is output for users doesn't include methane. I get why it isn't in that variable for the calculations, but from a user-standpoint it seems logical that it would be in an output variable for atmospheric carbon. 

**R code to reproduce issue**

```r
library(tidyverse)
vars <- c(ATMOSPHERIC_C(), ATMOSPHERIC_CO2(), FFI_EMISSIONS(), OCEAN_CFLUX(), LAND_CFLUX())
ini_file = system.file("input/hector_rcp45.ini", package="hector")
core <- hector::newcore(ini_file)
invisible(hector::run(core))
output <- hector::fetchvars(core, 2000:2100, vars=vars)

output %>% pivot_wider(id_cols=c(scenario, year,variable,value), names_from = variable, values_from=value) %>% 
  mutate(atm_flux=rowSums(cbind(ffi_emissions, -atm_ocean_flux, -atm_land_flux)),
         Ca_pgc=Ca*2.13, Ca_pgc_diff=c(NA,diff(Ca_pgc)), atmos_c_diff=c(NA,diff(atmos_c))) %>% 
  select(-c(scenario,ffi_emissions,atm_ocean_flux,atm_land_flux, Ca)) %>% 
  relocate(year,atmos_c,Ca_pgc,atmos_c_diff,Ca_pgc_diff,atm_flux) ->
  atm_output

head(atm_output,10)
tail(atm_output,10)
```
image

bpbond:
As @NatanelHa discovered and @dawnlwoodard documented in #459 , the `atmos_c` and `Ca` model outputs are off by a year. This is not good, but the fix is simple: make sure the latter gets updated when the former takes on a new value at the end of `simpleNbox::stashCValues()`.

Following Dawn's demonstration of the problem, new output looks like this:
```
> output %>% select(year, variable, value) %>% pivot_wider(names_from = "variable",values_from = "value") %>% mutate(Ca_from_atmos_c = atmos_c/ 2.13)
# A tibble: 101 × 4
    year atmos_c    Ca Ca_from_atmos_c
   <dbl>   <dbl> <dbl>           <dbl>
 1  2000    751.  352.            352.
 2  2001    754.  354.            354.
 3  2002    756.  355.            355.
 4  2003    760.  357.            357.
 5  2004    763.  358.            358.
 6  2005    767.  360.            360.
 7  2006    771.  362.            362.
 8  2007    775.  364.            364.
 9  2008    779.  366.            366.
10  2009    782.  367.            367.
```

There's no effect on the model behavior:

comp

Closes #459


bpbond:
Update: with 7872e86 I have just removed `Ca` and `Ca_ts` entirely as independent variables. This simplifies the code and eliminates any possibility of an inconsistency; `Ca` (now the function `Ca(time)`) is merely `atmos_c` expressed in different units.

comp

kdorheim (approved):
All of these changes look good to me! The only reason to wait on merging this PR is if we want to do handle the confusing naming of atmos C here in this PR

bpbond:
I'd vote to handle that separately, as part of a thorough go-through of all the names. If we're going to break things, we might as well do it comprehensively!

kdorheim:
Sounds like a plan to me! Merge whenever you are ready 👍 Thanks @bpbond, @dawnlwoodard, @NatanelHa for finding, diagnosing, and resolving this issue!

PR #583, opened 2022-03-16
Fix memory leak and use of uninitialized variable

### LINKED ISSUE #582: Memory leak / other memory issues
**Describe the bug**
When running a memory profiler on GCAM (still using hector 2.5.0) a came across a few issues in Hector.  At least one of which I can verify is still valid (apologies for the screen shots!  There was more information in them than if I tried to just get the errors from logs):

core_ozone

There is a memory leak in `Core::init` because `OzoneComponent` is listed twice.  Although the pattern used itself is prone to this sort of thing since we never check if a component of `temp-getComponentName()` has already has been added or not.  We should probably create a little function to add model components which will check that and warn + delete when it does happen.

I'm less sure about these.  There is an "invalid memory copy" in temperature component / the call to `erf` on this line.  Again I'm not sure why:
temp_comp_erf

Finally there is a usage of uninitialized value in ocean_csys:
oceancsys

**To Reproduce**
These were generated using Intel Inspector using the memory issue detector level 2 (again using hector 2.5.0 via GCAM):
```
inspxe-cl -collect mi2 -result-dir ~/intel/inspxe/projects/gcam5.5_mem -- ./gcam.exe -C configuration_ref.xml
```

**Expected behavior**
No memory leak (granted the leak is so small I doubt it would ever amount to anything significant).  I'm not really sure what consequences there are (if any) for the other two.

**Actual behavior**
No memory errors detected.

**Runtime and build environment (please complete the following information):**
 - OS: Linux
 - Boost: 1.76.0
 - R: NA

**Additional context**

bpbond:
This PR fixes two problems identified by @pralitp in #582 
* We had two `new OzoneComponent()` calls in `core.cpp`; one has been removed
* `ocean_csys::convertToDIC` was using an uninitialized variable as it was being called for the intermediate and deep boxes, which don't have active chemistry (and thus their `volumeOfBox` was never set). This has been changed
* Also a couple of minor tweaks elsewhere

There is no performance change:

comp

Closes #582

### DIFF
--- a/src/core.cpp
+++ b/src/core.cpp
@@ -111,9 +111,6 @@ void Core::init() {
     temp = new N2OComponent();
     modelComponents[ temp->getComponentName() ] = temp;
 
-    temp = new OzoneComponent();
-    modelComponents[ temp->getComponentName() ] = temp;
-
     temp = new ForcingComponent();
     modelComponents[ temp->getComponentName() ] = temp;
     temp = new slrComponent();
--- a/src/ocean_component.cpp
+++ b/src/ocean_component.cpp
@@ -248,7 +248,6 @@ void OceanComponent::prepareToRun() {
     const double I_preind_C = I_vol_frac * preind_C_ID;
     const double D_preind_C = D_vol_frac * preind_C_ID;
 
-
     // Set up our ocean box model.
     H_LOG( logger, Logger::DEBUG ) << "Setting up ocean box model" << std::endl;
     surfaceHL.initbox( HL_preind_C, "HL" );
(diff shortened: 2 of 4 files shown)


bpbond:
There are no more memory leaks after this:

Screen Shot 2022-03-16 at 2 32 40 AM

kdorheim:
^^ just curious was the memory leak figure made using Xcode? There were no changes in Hector output 👍 but did run time change at all? Do we need to check for leaky memory periodically?

kdorheim:
Could you humor me and plot a more specific ocean variable? such as DIC?

kdorheim (commented):
May be overly cautious here but could we plot ocean carbon cycle variable here?

bpbond:
Ah great idea.

bpbond:
dic

bpbond:
@kdorheim No hurry, but reminder — waiting on your review.

PR #584, opened 2022-03-18
Install libcurl

bpbond:
The `test-coverage` check is failing in land-tracking, because covr needs curl which needs libcurl. Bringing over the step we have in `leeyabot`.


bpbond:
This is a tiny change, merging without @kdorheim approval or waiting for the slow windows-latest run to finish.

PR #587, opened 2022-03-20
Net biome production constraint (v3-fluxpools-tracking)

### LINKED ISSUE #496: Re-implement NBP using fluxpools
Per @kdorheim [comment](https://github.com/JGCRI/hector/pull/493#discussion_r692448908), need to reimplement #397 in the v3 branch. As part of this get rid of that TODO in the code.

bpbond:
**Background.** This PR supplants the now-closed #397 : it adds a net biome production, the land-atmosphere exchange, constraint to Hector. This capability is needed for GCIMS SFA work. 

Doing so is considerably trickier than the other constraints implemented to date, because NBP is the sum of net primary production (NPP) and heterotrophic respiration (RH), and if we're imposing a constraint on it these need to be adjusted. Further, this needs to happen _during_ the solver, not after it, so that the land fluxes stay roughly consistent. It's not simply a matter of substituting in a new value.

**Method.** After NPP and RH have been computed in `calc_derivs`, we compare NPP-RH-LUC to the constraint value, if there is one; any difference is eliminated by increasing NPP by `diff/2` and subtracting the same value from RH. The NPP and RH sub-fluxes are then changed as well. After the solver finishes, in `stashCValues` we basically do same thing happens again: NPP and RH and their sub-fluxes are adjusted so that the constraint is honored, and so the carbon-tracking mechanism operates correctly.

This is demonstrated by imposing the following constraint:
```
; Sample NBP constraint data for testing
Date,NBP_constrain
2001,0.1
2002,0.2
2003,0.3
2004,-0.4
2005,-0.5
2006,-0.6
2007,0.7
2008,0.8
2009,0.9
2010,1.0
```
Results are shown here (dashed lines indicate the constraint period):
nbp_constraint_demonstration

1. The model hits the constraint exactly
2. It does so raising RH...
3. ...and lowering NPP
4. Because of this, atmospheric CO2 rises rapidly during the constraint period, as we've basically shut down the land C sink
5. As a result, when the model exits the constraint period, NPP shoots up very high (because the atmosphere is saturated with CO2) 
6. In a similar vein, the suppressed NPP means, once outside of the constraint, that RH is low 
7. ...and thus the land becomes a large sink as it draws down the excess atmospheric CO2

The overall run looks like this:
overall

In contrast to #397 the numerics here are solid (I hope) so the model will hit numeric constraints exactly.

**Compatibility.** This is not fully backward-compatible with existing Hector inputs and outputs, because I standardized the terminology to "NBP", getting rid of `LAND_CFLUX`, `landamos_flux`, etc. This work also adds a new setting `nbp_constrain` to the INI files and R interface, and fixes a variety of minor issues I found while doing this work.

Passes all local R package tests, lintr, and C++ unit tests. 
Closes #496


src/simpleNbox-runtime.cpp
```diff
@@ -240,32 +245,72 @@ void SimpleNbox::stashCValues( double t, const double c[] )
     fluxpool luc_fas_flux = atmos_c.flux_from_fluxpool(luc_u_untracked * ( 1 - f_lucv - f_lucd ));
 
     // Calculate net primary production and heterotrophic respiration
-    const fluxpool npp_total = sum_npp();
-    const fluxpool rh_total = sum_rh();
+    fluxpool npp_total = sum_npp();
+    fluxpool rh_total = sum_rh();
+    //BBL cout << "rh_total = " << rh_total << endl;
```
kdorheim, line 250: Does the c out need to be here?

src/simpleNbox-runtime.cpp
```diff
@@ -240,32 +245,72 @@ void SimpleNbox::stashCValues( double t, const double c[] )
...
+    // If there an NBP constraint? If yes, at this point adjust npp_total, rh_total,
+    // and the newveg/newdet/newsoil variables
+    double rh_nbp_constraint_adjust = 1.0;
+    const int rounded_t = round(t);
+    if(!core->inSpinup() && NBP_constrain.size() && NBP_constrain.exists(rounded_t) ) {
+        const unitval nbp_constrained = NBP_constrain.get(rounded_t);
+        const unitval diff = nbp_constrained - unitval(alf, U_PGC_YR);
+        
+        // Adjust fluxes
+        //BBL cout << "npp_total = " << npp_total << " rh_total = " << rh_total << endl;
```
kdorheim, line 279: Is this cout necessary? Should it be a debugging message or something?

src/simpleNbox-runtime.cpp
```diff
@@ -240,32 +245,72 @@ void SimpleNbox::stashCValues( double t, const double c[] )
...
+    }
+    
+    nbp.set( alf, U_PGC_YR );
+    nbp_ts.set( t, nbp );
+
+    // Apportion NPP and RH among the biomes
+    // This is done by NPP and RH; biomes with higher values get more of any C change
     for( auto it = biome_list.begin(); it != biome_list.end(); it++ ) {
         std::string biome = *it;
+        //BBL cout << "--- " << biome << endl;
```
kdorheim, line 311: same comment as a bove

src/simpleNbox-runtime.cpp
```diff
@@ -617,6 +671,35 @@ int SimpleNbox::calcderivs( double t, const double c[], double dcdt[] ) const
...
+          npp_fad = npp_fad * npp_ratio;
+          npp_fas = npp_fas * npp_ratio;
+
+          // Do same thing for RH
+          fluxpool rh_current_old = rh_current;
+          rh_current = rh_current - diff / 2.0;
+          const double rh_ratio = rh_current / rh_current_old;
+          rh_fda_current = rh_fda_current * rh_ratio;
+          rh_fsa_current = rh_fsa_current * rh_ratio;
+//const double new_nbp = npp_current.value(U_PGC_YR) - rh_current.value(U_PGC_YR) - luc_emission_current.value(U_PGC_YR) + luc_uptake_current.value(U_PGC_YR);
```
kdorheim, line 698: Why is this code commented out? Is it useful to leave in for debugging? If so please include a comment explaining why it is commented out and when you may want to turn it on.

tests/testthat/test_concforced.R
```diff
@@ -257,9 +257,38 @@ test_that("Tgav constraint works", {
     hc <- ssp245()
```
kdorheim, line 257: should we rename this file to test constraints? instead of test concentration forced?

kdorheim (changes requested):
Looks good! Just a few minor questions about some code that was commented out and a question about potentially changing a test that file name.

R/messages.R
```diff
@@ -53,6 +53,11 @@ fetchvars <- function(core, dates, vars = NULL, scenario = NULL) {
   valid <- dates >= strt & dates <= end
   dates <- dates[valid]
 
+  if (length(dates) == 0) {
```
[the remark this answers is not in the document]
bpbond: Ah, that what it's doing but the naming is confusing. Will change.

src/simpleNbox-runtime.cpp
bpbond: Removed

tests/testthat/test_concforced.R
bpbond: I wondered that too—thanks. Will do.

kdorheim (approved):
👏 big step towards v3

bpbond:
@dawnlwoodard No need to review, but FYI

PR #594, opened 2022-03-24
BBL tweaks to "A tour of the Hector R interface" page

bpbond:
Minor


leeyap (approved):
Agreed, looks good to me.

Issue #586, opened 2022-03-20
DOECLIM variables & time series


bpbond:
These guys:
```cpp
    std::vector<double> temp;
    std::vector<double> temp_landair;
    std::vector<double> temp_sst;
    std::vector<double> heatflux_mixed;
    std::vector<double> heatflux_interior;
    std::vector<double> heat_mixed;
    std::vector<double> heat_interior;
    std::vector<double> forcing;
```

_really_ should use the Hector time series class.

On a very related note, the whole `TemperatureComponent::getData` structure is bizarre, with a big if/else block depending on whether a date is supplied or not. This should be restructured to be like simpleNbox.

bpbond:
Let's shelve it for now.

bpbond:
The getdata section was the big one, in that it had bad behavior.

kdorheim:
Could this be a v3.5 milestone?

bpbond:
Yes!

PR #595, opened 2022-03-24
Reorganize getdata section of temperature_component.cpp 

kdorheim:
This PR address part of #586 re organizes how the output variables are returned


inst/include/temperature_component.hpp
```diff
@@ -103,7 +103,7 @@ class TemperatureComponent : public IModelComponent {
     double taudif;           // interior ocean heat uptake time scale, yr
     double tauksl;           // sea-land heat exchange time scale, yr
     double taukls;           // land-sea heat exchange time scale, yr
-    unitval q2co2;           // radiative forcing for atmospheric CO2 doubling (W/m2)
```
bpbond, line 106: ~Why removal of units?~

src/temperature_component.cpp
```diff
@@ -180,7 +180,7 @@ void TemperatureComponent::setData( const string& varName,
             volscl = data.getUnitval(U_UNITLESS);
         } else if(varName == D_QCO2) {
             H_ASSERT( data.date == Core::undefinedIndex(), "date not allowed" );
-            q2co2 = data.getUnitval(U_UNDEFINED);
```
bpbond, line 183: ...because it's unitless I guess :)

bpbond (approved):
I didn't follow logic all the way through but seems right 👍  thanks!

inst/include/temperature_component.hpp
kdorheim: Oh no it has units and I need to add them!

kdorheim: Removed it from a unitval because the math it is used in does not like the unitval, beforehand we had some code that basically extracted the unitval information and reassigned it as a const double which seemed uncessairy.

kdorheim:
@bpbond okay to merge?

bpbond:
Yes!

Issue #177, opened 2017-01-25
startDate non-inclusive?


rgieseke:
Is `startDate` non-inclusive?

It seems output streams start at 1746 only
```
ag 1745 sample_outputstream_rcp26.csv.csv
sample_outputstream_rcp26.csv
38650:1781,rcp26,0,temperature,Tgaveq,-0.0174588,degC
45669:1832,rcp26,0,ocean,OmegaAr_HL,2.17454,(unitless)
47483:1845,rcp26,0,ocean,pH_LL,8.17458,pH
47622:1846,rcp26,0,ocean,pH_LL,8.17452,pH
55317:1901,rcp26,0,forcing,FO3,0.1745,W/m2
72830:2020,rcp26,0,CH4,CH4,1745.82,ppbv CH4
90921:2146,rcp26,0,ocean,pH_HL,8.11745,pH
103956:2237,rcp26,0,simpleNbox,soil_c,1745.67,Pg C
104099:2238,rcp26,0,simpleNbox,soil_c,1745.06,Pg C
```
even though the ini files list the `startDate` as 1745:
```
startDate=1745
```

bpbond:
This is misleading, I agree. The outputstream code writes the _end date_ of the current timestep; e.g., when we're running 1 January 1745 - 31 December 1745, what actually gets written is (as you note) "1746".

What should be written is either "1745" (the year) or "1745.5" (the mathematical midpoint).

bpbond:
I think this is significant enough to mark it as a "bug" and put it in our next major release milestone. Thoughts @cahartin and others?

rgieseke:
Thanks for the explanation! 
So currently the temperature value for 1746 is the *average* value for 1745, correct?

bpbond:
@rgieseke right.

rgieseke:
Does this affect all output variables or are they any that only have yearly resolution? (Happy to add a note the [Output Files](https://github.com/JGCRI/hector/wiki/OutputFiles) wiki page).

bpbond:
This should affect everything.

rgieseke:
Added a note there based on your comment.

bpbond:
Thanks @rgieseke

rplzzz:
[...]

PR #596, opened 2022-03-24
BBL tweaks to outputstream page - minor

bpbond:


bpbond:
@kdorheim Should we address/fix/close https://github.com/JGCRI/hector/issues/177 ?

bpbond:
Is this page linked to from anywhere? 😕

kdorheim:
> @kdorheim Should we address/fix/close #177 ?

Would fixing it be a big-time commitment? If not then I think we probably should

kdorheim:
> Is this page linked to from anywhere? 😕

ohh hmm good point we could probably add a link to in in the build hector article where we discuss running command line, Xcode & VS hector. It is currently listed on the manual index and would be findable via the search bar once that is set up.

bpbond:
How would I (or you) add that link?

leeyap (approved):
Nothing to add!

bpbond:
OK I have assigned #177 to myself to take a hard look at.

kdorheim:
> How would I (or you) add that link?

it is going to be something like [text related to article](ArticleName.html)

PR #604, opened 2022-04-03
Model output includes startDate

### LINKED ISSUE #177: startDate non-inclusive?

bpbond:
Changes model output so that the csvOutputStream as well as `hector::fetchvars` data now start with 1745, or whatever `startDate` is. 

Closes #177

### DIFF
--- a/src/core.cpp
+++ b/src/core.cpp
@@ -395,9 +395,9 @@ bool Core::run_spinup()
         for( NameComponentIterator it = modelComponents.begin(); it != modelComponents.end(); ++it )
             spunup = spunup && ( *it ).second->run_spinup( step );
         // Let visitors attempt to collect data if necessary
-        for( VisitorIterator visitorIt = modelVisitors.begin(); visitorIt != modelVisitors.end(); ++visitorIt ) {
-            if( ( *visitorIt )->shouldVisit( in_spinup, step ) ) {
-                accept( *visitorIt );
+        for( auto visitorIt : modelVisitors ) {
+            if( visitorIt->shouldVisit( in_spinup, step ) ) {
+                accept( visitorIt );
             }
         } // for
     } // while
@@ -467,9 +467,9 @@ void Core::run(double runtodate) {
     // component hasn't had a chance to initialize its active chemistry yet, and without that
     // calc_revelle() will throw an error. Kind of hacky; but the only alternative I see is
     // to create a whole new done_with_spinup() signal to the components--a lot of work.
-    for( VisitorIterator visitorIt = modelVisitors.begin(); visitorIt != modelVisitors.end(); ++visitorIt ) {
-        if( ( *visitorIt )->shouldVisit( true, lastDate ) ) {
-            accept( *visitorIt );
+    for( auto visitorIt : modelVisitors ) {
+        if( visitorIt->shouldVisit( true, lastDate ) ) {
+            accept( visitorIt );
         }
     }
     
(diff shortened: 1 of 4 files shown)


tests/testthat/test_wrapper.R
```diff
@@ -40,6 +40,10 @@ test_that("Basic hcore functionality works", {
   expect_silent(run(hc, 2100))
   expect_silent(run(hc))
 
+  # All years should appear in the output
+  out <- fetchvars(hc, dates = 1745:2300)
+  expect_equivalent(sort(unique(out$year)), 1745:2300)
```
kdorheim, line 45: Why are there two expect_silent(run) in a row after each other? Is the second run(hc) call just to run the date out to the year 2300 for all years should appear in the output? test?

src/ocean_component.cpp
```diff
@@ -299,6 +299,11 @@ void OceanComponent::prepareToRun() {
     surfaceLL.mychemistry.As            = ocean_area * part_low; // surface area m2
     surfaceLL.mychemistry.U             = 6.7; // average wind speed m/s Hartin et al. 2016
 
+    // Initialize surface flux tracking variables
+    annualflux_sum.set( 0.0, U_PGC );
+    annualflux_sumHL.set( 0.0, U_PGC );
+    annualflux_sumLL.set( 0.0, U_PGC );
```
kdorheim, line 305: The annualflux_sum is also set in the run component around line 360 does it really need to be set twice? Or could we remove that code block?

kdorheim (commented):
Some questions

tests/testthat/test_wrapper.R
bpbond: This is older code, not something I added. The model is at 2100 at that point, and those lines test that it will run to 2100 (no change) and then continue to 2300, all without error.

src/ocean_component.cpp
bpbond: It needs to be zeroed her as well, so that the initial output visitor pass — which now occurs _before the component has ever run_ — can correctly report model state.

src/core.cpp
```diff
@@ -460,8 +459,24 @@ void Core::run(double runtodate) {
...
     // 6. Run all model dates.
     H_LOG( glog, Logger::NOTICE) << "Running..." << endl;
+    
+    
+    // Let visitors record initial model state
+    // Note that just for this initial output, in calling shouldVisit below, we're telling
+    // visitors that spinup is 'true' (which it's not). This is necessary because the ocean
+    // component hasn't had a chance to initialize its active chemistry yet, and without that
+    // calc_revelle() will throw an error. Kind of hacky; but the only alternative I see is
+    // to create a whole new done_with_spinup() signal to the components--a lot of work.
```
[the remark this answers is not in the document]
bpbond: It's a bit hacky, yes; allows the visitor to collect model state even though the ocean chemistry has not yet been turned on. That last bit is the issue; it occurs the first time the ocean is `run()`. Now, however, the visitor needs to output things and, because we're no longer in spinup, it tries to collect info on chemistry (and fails). So instead, we tell the visitor, one last time, that we're still in spinup.

kdorheim (approved):

PR #605, opened 2022-04-03
Require dates for more model variables

### LINKED ISSUE #591: Zeroes getting returned from fetchvars
Uh oh, I have a suspicion that my #496 may have broken `fetchvars` — although if that's the case don't know why tests didn't flag the issue! @leeyap and @BennettForville found
```r
> library(hector)
> RCP45 <- system.file("input/hector_ssp245.ini", package = "hector")
> core45 <- newcore(RCP45)
> reference <- fetchvars(core45, NA)
> head(reference)
             scenario year variable  value    units
1 Unnamed Hector core   NA       Ca 277.15 ppmv CO2
2 Unnamed Hector core   NA     Ftot   0.00     W/m2
3 Unnamed Hector core   NA     FCO2   0.00     W/m2
4 Unnamed Hector core   NA     Tgav   0.00     degC
```
which doesn't look good.

bpbond:
#591 documents that, if you call `fetchvars(hc, dates = NA)` you get back the end-of run Ftot, Ca, FCO2, and Tgav data with "NA" for year.

Fixing this was slightly tricky. Previously, all these variables could be queried from their respective model components _without_ supplying a date, and the current value internal value would be returned. This was handy but opened the door to the R user querying without a date as well. Fixing this required:
* Change Ca to require a date, which means supplying the date in the two internal queries for it: in `csvOutputStreamVisitor` (straightforward) and `OceanComponent` (less so). We now ensure that the parameter C0 is written to the `atmos_c_ts` variable, and that it's allowed to interpolate, so that the ocean can get values correctly.
* Change all forcing variables to require a date. This involved reworking `forcingComponent::getData`, which was a mess.
* `fetchvars(hc, NA)` now returns an error, and it was potentially unclear, so I added a special check for this.

Closes #591

### DIFF
--- a/tests/testthat/test_messages.R
+++ b/tests/testthat/test_messages.R
@@ -10,5 +10,12 @@ test_that("Invalid input variable returns error", {
     expect_error(setvar(core, NA, "global.permafrost.beta", 10, ""),
                  regexp = "Invalid input variable: '")
     shutdown(core)
+})
 
+test_that("Default fetchvar without dates returns error", {
+
+    # Test that fetchvars errors if neither dates nor vars is supplied
+    core <- newcore(inifile)
+    expect_error(fetchvars(core, NA), regexp = "all require dates")
+    shutdown(core)
 })
--- a/R/messages.R
+++ b/R/messages.R
@@ -41,6 +41,9 @@ fetchvars <- function(core, dates, vars = NULL, scenario = NULL) {
         f()
       })
     )
+    if (anyNA(dates)) {
+        stop("The default vars (", paste(vars, collapse = ", "), ") all require dates")
+    }
   }
 
   if (is.null(scenario)) {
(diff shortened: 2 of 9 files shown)


tests/testthat/test_messages.R
```diff
@@ -10,5 +10,12 @@ test_that("Invalid input variable returns error", {
     expect_error(setvar(core, NA, "global.permafrost.beta", 10, ""),
                  regexp = "Invalid input variable: '")
     shutdown(core)
+})
+
+test_that("Default fetchvar without dates returns error", {
```
kdorheim, line 15: Could we also test a nondefault variable like a halocarbon RF or some other random time series variable? The way that the test is set up and described gives the impression that this error will only occur for the default fetchvar variables.

kdorheim (changes requested):

tests/testthat/test_messages.R
bpbond: Well, that _is_ what the error will catch: default variables (`vars` parameter is NA) and no dates (NA).

kdorheim: Haha okay true. So only three variables are defined as default variables 
`default_fetchvars <- c(ATMOSPHERIC_CO2, RF_TOTAL, RF_CO2, GLOBAL_TEMP)`

I guess my question is why are we treating these variables differently than other Hector variables? Or was the behavior that problematic, the return values when date == NA only any issue for these three variables?

bpbond: My thought was that if someone does
```r
fetchvars(hc, years = NA, vars = GLOBAL_TEMP())
```
they'll get an error but then can look at the documentation and understand that global temperature requires dates. But if they do
```r
fetchvars(hc)
```
they might be confused because they haven't requested (they don't think) any variables.

kdorheim (approved):

PR #606, opened 2022-04-03
Convert most loops to C++11 auto

bpbond:
The [C++11 `auto`-style for loop](https://en.cppreference.com/w/cpp/language/range-for) is simpler and clearer. Let's take advantage of this language feature.

I don't think there's any reason not to, but adding @pralitp for perspective.

### DIFF
--- a/inst/include/fluxpool.hpp
+++ b/inst/include/fluxpool.hpp
@@ -340,8 +340,8 @@ ostream& operator<<(ostream &out, fluxpool &rhs ) {
     if(rhs.tracking) {
         out << endl;
         std::vector<std::string> sources = rhs.get_sources();
-        for (int i = 0; i < sources.size(); i++) {
-            out << "\t" << sources[i] << ": " << rhs.get_fraction(sources[i]) << endl;
+        for ( auto src : sources ) {
+            out << "\t" << src << ": " << rhs.get_fraction( src ) << endl;
         }
     }
     return out;
--- a/inst/include/fluxpool.hpp
+++ b/inst/include/fluxpool.hpp
@@ -103,9 +103,9 @@ fluxpool::fluxpool(unitval v, unordered_map<string, double> pool_map, bool track
 
      // check pool_map data
     double frac = 0.0;
-    for ( auto itr : ctmap ) {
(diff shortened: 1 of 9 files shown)


kdorheim (approved):
Sounds good to me but I am not a C++ expert @pralitp would a more helpful reviewer here. 

My one question is that can this be used for all C++ for loops? If not why not? And if so why not make this change to all the C++ for loops?

bpbond:
@kdorheim We don't have many other kinds of for loops, I don't think...most of ours are iterating over collections of objects, which is what this `auto` syntax is so nice for.

kdorheim:
So I guess the question is why only change some of the for loops to auto loops?

bpbond:
Oh did I miss some? Was just searching for `begin()` calls :)

kdorheim:
Not sure if these for loops need to be updated but 


forcing_componet.cpp -  line 437
```

        // Halocarbons can be disabled individually via the input file, so we run through all possible ones
         for (unsigned hc=0; hc<halos.size(); ++hc) {
            if( core->checkCapability( halos[hc] ) ) {
                // Forcing values are actually computed by the halocarbon itself
                forcings[ halos[hc] ] = core->sendMessage( M_GETDATA, halos[hc], message_data( runToDate ) );
                }
        }
```

INIReader.cpp - line 64

```
    // Convert to lower case to make lookups case-insensitive
    for (unsigned i = 0; i < key.length(); i++)
        key[i] = tolower(key[i]);
```

spline_forsythe.cpp -  there are for loops in lines 89, 110, 118, 126

temperature.cpp - there are fore loops starting on lines 346 and 471

simpleNox.hpp - there are for loops in lines 243, 260, 285

bpbond:
I'll take a look. Thanks.

bpbond:
@kdorheim I added a couple other loops, thanks.

Many, however, aren't suitable for re-doing using the `auto` form: either something in the loop needs an index value, or (especially in the dependency finder code) they're returning objects, or other subtleties.

kdorheim:
Awesome if it's not applicable then its not, glad that we are being consistent with the for loops that can use that format.

PR #610, opened 2022-05-09
Updating output names 1: non halocarbon concentrations 

kdorheim:
This is a part of the overhaul of the output variable names. 

There are a few questions that I have run into looking at names/strings for the ch4, n2o, and atmospheric co2 stuff. 
1. the halo carbons constraints follow the pattern `D_CONSTRAINT_CF4` whereas it is `D_CO2_CONSTRAIN`. Do we want to be consistent with our use of constrain vs constraint 
2. we discussed switching to atmos_co2 what about `CO2_conc` to be consistent with all of the other ghg concentration naming patterns?


inst/include/component_data.hpp
```diff
@@ -326,14 +326,13 @@
...
 #define D_TIMESTEPS             "ocean_timesteps"
 #define D_REVELLE_HL            "Revelle_HL"
 #define D_REVELLE_LL            "Revelle_LL"
 
 // SimpleNbox component
 #define D_NBP                   "NBP"
-#define D_ATMOSPHERIC_CO2       "Ca"
+#define D_ATMOSPHERIC_CO2       "CO2_conc"
 #define D_PREINDUSTRIAL_CO2     "C0"
 #define D_ATMOSPHERIC_C         "atmos_c"
```
kdorheim, line 337: @bpbond should this be atmos_co2? Since it is atmos CO2 in units of C?

bpbond: It's not in units of C; it's parts per million (vol).

>we discussed switching to atmos_co2 what about CO2_conc to be consistent with all of the other ghg concentration naming patterns?

Hmm, both have advantages. I think `CO2_conc` makes the most sense?

inst/include/component_data.hpp
```diff
@@ -250,7 +250,7 @@
 #define D_HCDELTA_CH3Br            D_HCDELTA_PREFIX CH3Br_COMPONENT_BASE
 
 // methane component
-#define D_ATMOSPHERIC_CH4       "CH4"
+#define D_ATMOSPHERIC_CH4       "CH4_conc"
```
bpbond, line 253: Given these, yeah definitely `CO2_conc`

inst/include/component_data.hpp
```diff
@@ -326,14 +326,13 @@
 #define D_CO3_LL				"CO3_LL"
 #define D_CO3_HL				"CO3_HL"
 #define D_CO3				    "CO3"
-#define D_ATM_OCEAN_CONSTRAIN   "atm_ocean_constrain"
```
bpbond, line 329: 👍

bpbond (approved):
Are vignettes checked for name changes?

inst/include/component_data.hpp
kdorheim: I think CO2_conc makes the most sense. But I was wondering if D_ATMOSPHERIC_C should go from atmos_c to atmos_co2, because that is less misleading since D_ATMOSPHERIC_C does not include carbon from non co2 sources?

kdorheim:
> Are vignettes checked for name changes?

ohhh great question, none of the R functions changed in this PR which is where I would expect to see changes causing vignettes to fail. All vignettes build for this PR. I have not checked the contents to see if there have been any changes to formatting which could happen if a custom color scale or labels were being used.

inst/include/component_data.hpp
bpbond: `atmos_co2` seems highly misleading to me; the variable is tracking carbon, not CO2. 

`atmos_co2-c`?

kdorheim:
I updated the CH4, N2O concentrations to be consistent with how we are treating the CO2 concentrations vs atmospheric. Did you want to take a look at those changes or okay to merge? @bpbond

bpbond:
@kdorheim OK to merge but one question first:

`CONCENTRATIONS_CO2()` is (i) plural (which is a little weird? Nothing else is) and (ii) really long.

Any interest in making that `CONC_CO2()`? Maybe that's too short 😆

kdorheim:
Can totally change to `CONC_`  CONCENTRATIONS_ is pretty long, bleh but figured it was consistent with the set of `EMISSIONS_` functions

PR #612, opened 2022-05-10
Updating output names 3: ocean variables

kdorheim:
Updating the names of the ocean output variables. 

There are two outstanding issues 

* Trying to fetch HL.sst and LL.sst within R throws an error. I think it has to do with the `.` but it is not clear as to why this is an issue for this variable and not for the other ones 
* What do we want to name `HL_DO_Cflux`?


bpbond:
>Trying to fetch HL.sst and LL.sst within R throws an error. I think it has to do with the . but it is not clear as to why this is an issue for this variable and not for the other ones

That's...odd. What's the error specifically?

>What do we want to name HL_DO_Cflux?

🤷  we need an ocean person to suggest good names! 

What was the issue with `HL.downwelling`?

kdorheim:
Hi Ben, here is a reproducible example of the problem that I have been running into, do you have time to take a look at it? 

```
# Build and run Hector from Xcode, there should be no problem.
# Take a look at the output stream you will see HL.sst and LL.sst returned
# as output variables no problem.


devtools::load_all()
ini <- here::here("inst", "input", "hector_ssp126.ini")
core <- newcore(ini)
run(core)
# Awesome! the hector core ran wihtout problem

# this will throw a problem
SST_HL()
fetchvars(core, 1900:1920, vars = SST_HL())

# so does this!
SST_LL()
fetchvars(core, 1900:1920, vars = SST_LL())


# But other HL and LL variables are fine
SST_LL()
fetchvars(core, 1900:1920, vars = c(PCO2_HL(), OCEAN_C_HL(), PH_HL()))

# whyyyy????
```


R code attached as a github supported txt file
[hlsst_problem.txt](https://github.com/JGCRI/hector/files/8680193/hlsst_problem.txt)

bpbond:
@kdorheim This is kind of my fault, what with my "let's make this consistent with how simpleNbox does biomes!"

## What happens

1. We call `fetchvars()` in R
2. It dispatches (in `messages.R:67` to `core::sendMessage()`
3. Now the crucial bit. In `core.cpp:707`, the 'datum' part of the message, e.g. "HL.sst", is split apart if there are any periods (`SNBOX_PARSECHAR`):
```c++
    boost::split( datum_split, datum, boost::is_any_of( SNBOX_PARSECHAR ) );
```
4. The `datum_capability` is then set to whatever comes after the split, if there is one, since we're expecting something like "boreal.soil_c" (and "soil_c" is the "capability" we need to find)
5. The core searches (in line 728) for this capability and passes on the message to the appropriate component 
6. With the new `SST_HL()` = "HL.sst", what happens? The string is split, and the core searches for capability "sst" but can't find it, and so errors

## How to fix

Two options:
1. Rename these variables so they don't use a "." in their names
2. Have the ocean component register the capability "sst", and then when the core sends us a "getData" message, follows the model of `simpleNbox::getData()` to extract the "HL" and return correct value

Lmk, as the kids say. Happy to help if useful.

bpbond (approved):

bpbond:
...note conflicts obviously.

PR #613, opened 2022-05-10
Updating output names 4: temperature componet

kdorheim:
The unit tests are failing here related to the problem with the ocean component. 

One outstanding issue is would we want to change the use of tgav to tas within the C++ code?


bpbond:
>unit tests are failing

😕 

>tgav to tas within the C++ code?

I would think so?

kdorheim:
@bpbond okay to merge?

PR #615, opened 2022-05-20
Final v3 name changes

kdorheim:
Final output name changes & documeantion


kdorheim:
@bpbond the one question I have is that for the output markdown I wrote a script that scrapes a cpp file to determine the R function names & the strings they return for the output article it does need some cleanup and documentation but do you think this is something that we would want to add to the package? Thoughts?

[R_outputs.txt](https://github.com/JGCRI/hector/files/8741192/R_outputs.txt)

bpbond:
Sorry @kdorheim will look

bpbond:
`output.Rmd` right?

kdorheim:
yes please take a look at those rmd documents but should we keep the script I wrote to generate those? see [R_outputs.txt](https://github.com/JGCRI/hector/files/8741192/R_outputs.txt)

bpbond:
That's slick @kdorheim - nice job. Yes, I would keep that script, if only as documentation for the future. Add a couple comments to it, and stick it into `misc/` or somewhere?

kdorheim:
@bpbond okay to merge?

bpbond (approved):

PR #619, opened 2022-06-16
Add CH4 params to R interface; fix units string bug in unitval

bpbond:
The methane component parameters `LIFETIME_SOIL()` and `LIFETIME_STRAT()` were exposed in the R interface but not connected to anything—using them in `fetchvars` or `setvar` would produce an error.

Needed for @nataliecrowe060 work; see https://github.com/nataliecrowe060/Methane-NCrowe/issues/4 

This PR also removes two temp files that were accidentally committed during the output-renaming work.

A number of the GitHub Actions yaml files were still set to run for the `land_tracking` branch, not `v3_dev`; this has been fixed.

### DIFF
--- a/.github/workflows/command-line.yaml
+++ b/.github/workflows/command-line.yaml
@@ -12,8 +12,7 @@ on:
   pull_request:
     branches:
       - master
-      - land_tracking
-      - tracking-vars
+      - v3_dev
 
   schedule:
     # * is a special character in YAML so you have to quote this string
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -3,7 +3,7 @@ on:
     branches:
       - main
       - master
-      - land_tracking
+      - v3_dev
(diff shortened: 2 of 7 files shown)


bpbond:
Why is the lintr check failing? First, I didn't touch any of those files; second, it doesn't fail locally. Confusing.

bpbond:
Added a new C++ unit test.

Basically, in adding a new variable to Hector's R interface I realized that [unitsval::unitName](https://github.com/JGCRI/hector/blob/master/src/unitval.cpp) had a mistake: it skipped one of the unitval units, right near the end of the list. This meant that any attempt to get unit names after that mistake would fail. Fixed that and then realized we should have a test to prevent this from ever happening above—that's the dea9659 commit  above.

kdorheim (approved):
@bpbond looks good thanks for deleting those files! And good catch on the unitvals bug

PR #622, opened 2022-06-18
Add h_util.cpp back to hector-tests target

bpbond:
Somehow `h_util.cpp` got removed from the `hector-tests` target, breaking the testing code build. This PR adds it back.


kdorheim (approved):
👍

PR #623, opened 2022-06-18
Silence boost-related clang warnings

bpbond:
Boost's `formatter.hpp` has a documentation error that generates clang warnings when building Hector:

Screen Shot 2022-06-18 at 2 30 25 PM

These warnings are a PITA and make it hard to find Hector-related issues.

This PR surrounds several Boost includes with clang pragmas, silencing the warnings.

### DIFF
--- a/inst/include/logger.hpp
+++ b/inst/include/logger.hpp
@@ -15,7 +15,10 @@
  */
 
 #include <iostream>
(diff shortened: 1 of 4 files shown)


kdorheim (approved):
Thanks for doing this!

PR #624, opened 2022-07-05
Update the parameter values used in the ini

kdorheim:
In this PR update the parameter values in the ini files, add carbon cycle parameter units to the units table and adjust testing. 
While I have made several changes to the test_biome.R, but a number of them are still failing. I am not as familiar with how the biomes work as @bpbond or @dawnlwoodard so input would be important.


tests/testthat/test_biome.R
```diff
@@ -1,14 +1,23 @@
...
-  newcore(system.file("input", "hector_ssp245.ini",
-    package = "hector"
-  ),
-  name = "test core",
-  suppresslogging = TRUE
-  )
+    core <- newcore(system.file("input", "hector_ssp245.ini", package = "hector"),
+            name = "test core", suppresslogging = TRUE)
+
+    # names <- c(NPP_FLUX0(), BETA(), Q10_RH(), ECS(), DIFFUSIVITY(), AERO_SCALE())
```
kdorheim, line 7: @bpbond something that I find somewhat curious/alarming is that if we initiate the hector core with the old hector parameter values more of the tests pass. But wouldn't we expect the sum(biomes) = global regardless of the parameters used?

bpbond: Ugh, that doesn't sound good. I can look at this in more detail later.

bpbond (approved):
Looks good @kdorheim . Maybe change the PR message to clarify _why_ you're changing these parameters?

kdorheim:
FYI this is relevant to @leeyap  https://github.com/JGCRI/trackingC/issues/48#issuecomment-1172482226

bpbond:
@kdorheim I'm taking a hard look at this. I think we have a model problem, but it's going to take me a little bit to track down.

kdorheim:
@bpbond okay to merge?

bpbond:
Look at all those lovely green checks! OK to merge.

PR #625, opened 2022-07-05
Add the search bar

kdorheim:
I am working with a team at algolia on getting the doc search bar set up, hopefully, this works.


kdorheim:
Although I don't think it will be online until this is merged into master and active on the gh-page

PR #632, opened 2022-07-24
Make package style (lintr) check part of regular testing, not separate GitHub Action

bpbond:
I'm really tired of forgetting to check `lintr` and then having things error on GitHub Actions!


kdorheim:
Agreed! This is very annoying and this seems like a great idea!

kdorheim (approved):
Thanks for doing this, this is a great idea! I only have one question about the package version

bpbond:
The minimum version requirement for `lintr`? It's because of a bug when running `expect_no_lints()` in older versions — see https://github.com/r-lib/lintr/issues/421

PR #634, opened 2022-07-26
Apply consistent format across the Hector C++ codebase

bpbond:
This PR supersedes #418 .

This branch applies [ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) (version 14.0.6, installed via `brew install clang-format`) to all of the C++ source and header files. I didn't specify a particular format and so believe the default LLVM format was used.
```
hector % clang-format -i src/*.cpp
hector % clang-format -i src/ini.c
hector % clang-format -i inst/include/*.h
hector % clang-format -i inst/include/*.hpp
```

This resulted in build errors because `clang-format` rearranges includes alphabetically, and so both `avisitor.hpp` and `h_util.hpp` were tweaked to include `<iostream>`.

Result: totally consistent code style throughout!  🎉


kdorheim:
Whoot whoot! That looks great! I guess my one question is, how do we keep enforcing this in future developments? Is it something we run or github runs?

bpbond:
Good question. If we wanted to do this automatically, I think we'd use a pre-commit hook that runs `clang-format`.

We could also add a GitHub Action test: run clang-format on modified files, and if anything changes, fail.

kdorheim:
Hmm I don't have a preference so long as we choose one option that makes this a sustainable or a must do

bpbond:
@kdorheim Following our conversation yesterday, in 3c93200dae3014b3ba6bb9693a36217d1a44e717 I have added information to `vignettes/articles/StyleGuide.Rmd`on
* the C++ style used; and
* how to run `clang-format`

I don't know how/whether to rebuild the help pages though; is this documented anywhere?

kdorheim:
> I don't know how/whether to rebuild the help pages though; is this documented anywhere? 

What do you mean? Like build the style guide? It is built as part of the pkgdown documentation

kdorheim (approved):

bpbond:
Oh, I don't have to do anything?!? 🎉  Thanks :)

kdorheim (approved):

PR #637, opened 2022-09-08
Update NEWS.md

bpbond:
Not fully complete (need links for submitted Dorheim and Pressburger manuscripts), but here's a draft of the v3 `NEWS.md` file.

FYI @stephpenn1


kdorheim (approved):
Thanks for doing this!

PR #645, opened 2022-10-29
Remove f_lucv and f_lucd

bpbond:
This PR removes `f_lucv` and `f_lucd` entirely from the Hector model.

divergence


bpbond:
Superseded by #648

PR #648, opened 2022-11-01
Remove f_luc* variables

bpbond:
## Merge only after #647 

If we are going to shift to proportional LUC losses, as in #647 , then we either keep the `f_luc*` approach as an option (but see note about complexity [here](https://github.com/JGCRI/hector/pull/639#issuecomment-1295813020)) or get rid of it. That is, we can either keep or remove the ability of users to custom-specify how LUC removes/adds carbon to the terrestrial land surface.

This PR does the latter, by cherry-picking the changes from #645 into a new branch . 

This has no effect on the output relative to #647 :

diff


bpbond:
Updated comment above for clarity.

NAMESPACE
```diff
@@ -17,7 +17,6 @@ export(CH3BR_CONSTRAIN)
 export(CH3CCL3_CONSTRAIN)
 export(CH3CL_CONSTRAIN)
 export(CH4_CONSTRAIN)
-export(CO2_CONSTRAIN)
```
kdorheim, line 20: I flagged this as an issue with #639 we are losing a lot of R wrapper functions with this commit.

kdorheim (changes requested):
It looks good other than what I think may have been an accidental change in the arc/rcpp_constatnts file that causes lots of the function to be removed from the R wrapper

NAMESPACE
bpbond: I don't understand what's happening here.

bpbond:
@kdorheim Thank you for a36399dfc! Something has gotten borked in my roxygen2 setup, not sure what but it was throwing up all kinds of errors when I asked it to rebuild.

kdorheim (approved):

bpbond:
Merged in latest `v3_dev` changes and confirming that this PR still does not change output:

diff

diff_percent

Updated comparison data in 18022cd1 and re-enabled old-new test.

Issue #450, opened 2021-07-06
Deal with remaining nonfunctional testing files


bpbond:
As they are fixed, remove from the Makefile exclusion list. See https://github.com/JGCRI/hector/pull/449#discussion_r664498533

kdorheim:
@bpbond okay to merge or still on the todo list?

bpbond:
Leave open please

PR #652, opened 2022-11-03
Bring test_core.cpp up to date and re-enable it

bpbond:
See #450


kdorheim (approved):
Is this unit-testing part of the gtest?

bpbond:
@kdorheim Right, exactly. Thank you.

PR #653, opened 2022-11-03
Bring test_csv_file_reader.cpp up to date and re-enable it

bpbond:
See #450

PR #654, opened 2022-11-03
Bring test_dependency_finder.cpp up to date and re-enable it

bpbond:
See #450 

@pralitp Re-enabling this Hector test file was triggering the following error from inside the googletest framework itself:

Screen Shot 2022-11-03 at 7 02 25 AM

...that stemmed from lines in `test_dependency_finder.cpp` like this:

Screen Shot 2022-11-03 at 7 07 53 AM

As you can see from the diff, I removed the ` << "Ordering: " << ordering << endl` end of the lines, after which things build and tests run fine. But I don't quite understand why.  Any insight? Are these lines now giving false positives, or was this something cosmetic that was fine to remove? Thanks.

### DIFF
--- a/project_files/Xcode/hector.xcodeproj/project.pbxproj
+++ b/project_files/Xcode/hector.xcodeproj/project.pbxproj
@@ -82,6 +82,7 @@
 		27B2C270261511C8005DE26D /* spline_forsythe.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 27B2C24D261511C8005DE26D /* spline_forsythe.cpp */; };
 		27B2C271261511C8005DE26D /* main.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 27B2C24E261511C8005DE26D /* main.cpp */; };
 		27F2613726B55617004BDD47 /* csv_tracking_visitor.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 27962904268F237200333AA3 /* csv_tracking_visitor.cpp */; };
+		354045FC2913D50A00CA297D /* test_dependency_finder.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 273FC38F26A3570B00D5303E /* test_dependency_finder.cpp */; };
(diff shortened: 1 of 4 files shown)


bpbond:
@kdorheim #450 can be closed after merging #652 #653 #654

bpbond:
Thanks Pralit!

>Did we used to provide one somewhere and it got moved around in such a way that it is now out of scope

Not that I know of.

bpbond:
P.S. I guess the only thing that's being lost is the custom failure message, so not urgent.

pralitp:
I just reverted removing the custom messages and it built and ran the tests just fine.  However, I used the "HEAD" `googletest` as they recommended which might make a difference (or compiler: I used GCC 10.2)

@bpbond were you getting this locally or when enabled via GitHub Actions?  If the later I can try again in Docker to try to get closer to what GH actions is doing.

bpbond:
It ran fine for me locally; only on GA was there a problem. 

Thanks!

kdorheim (commented):
This is outside of my C++ knowledge, I think Pralit's feedback is more important here

kdorheim (approved):

bpbond:
@pralitp OK, it runs fine on GA. I will revert those changes, update my local googletest, and push. OK to merge after that?

Thanks!

pralitp (approved):

pralitp:
Great!  One minor fix I was going to suggest but can obviously go elsewhere is a doc fix:
```
diff --git a/vignettes/articles/BuildHector.Rmd b/vignettes/articles/BuildHector.Rmd
index 03c9690..69e7bea 100644
--- a/vignettes/articles/BuildHector.Rmd
+++ b/vignettes/articles/BuildHector.Rmd
@@ -164,7 +164,7 @@ To run build and run the Hector unit tests, type `make testing`
 and then

 ```
-./src/Testing/hector-unit-tests
+./src/unit-testing/hector-unit-tests
 ```

 ## Xcode (Mac OS X)
```

bpbond:
Skipped tests since just a single doc file modified in c3f28fd3

Issue #547, opened 2022-01-11
Variable names in R


leeyap:
There doesn't seem to be a clear way to figure out what a Hector variable is in R (that I know of). Both in terms of science, but also in terms of communicating between R and the core (e.g. can call `BETA()` to get `"beta"` but not the other way around.

We could potentially create a "dictionary" in the manual of variable names and brief description/how to access them.

bpbond:
Maybe linked with #362

bpbond:
@kdorheim Does the existence of `data/inputstable.rda` mean that this can be closed?

kdorheim:
hmm I think the inputstable.rda is missing the function name element of what Leeya was seeking but we sort of have that already with the output table article...

kdorheim:
Added in #656

PR #656, opened 2022-11-07
Expose variables & add the lookup function capability 

kdorheim:
Documentation changes & added a new capability, this PR does not change Hector output 

This PR is somewhat miscellaneous but
* Updates the RF_ALBEDO name to be consistent with the output rename that was done in an early PR 
* Exposes LUC_UPTAKE so that users can set these emissions using the R interface 
* Cleans up and completes the units-table, note that the units table is not a comprehensive table of all hector output/input units but only the input variables that are commonly changed/set by users
* Add capability to generate a table of Hector functions & their string names as desired in #547


data-raw/lookup-data.R
```diff
@@ -0,0 +1,37 @@
...
+    assertthat::assert_that(is.function(x))
+    x()
+}
+
+# Read ini the contents of the rcpp file, this is written with the working
+# directory set to the root directory of the Hector repository/package.
+lines <- readLines("./src/rcpp_constants.cpp")
+
+# Specific functions to exclude
+# TODO is there a way to make this not hard coded?
```
bpbond, line 22: Hmm not that I can see.

data-raw/lookup-data.R
```diff
@@ -0,0 +1,37 @@
...
+conditions <- (grepl("^String ", x = lines) & !grepl("^ ", x = lines))
+string_names <- lines[conditions]
+string_names <- gsub(pattern = "\\(String", replacement = "()", x = string_names)
+
+# Parse out the function names, evaluate the function then store results in
+# a table.
+fxn_names <- unlist(lapply(string_names, function(s) {
+    unlist(strsplit(s, split = " "))[2]}))
+fxn_vals <- unlist(lapply(fxn_names, hector_eval_func))
+fxntable <- data.frame(fxn = fxn_names, string = fxn_vals)
```
bpbond, line 35: This script is quite clever 👏

bpbond (commented):
Overall it looks good—nice job @kdorheim .

bpbond:
>Something I am not sure about is the naming of the function and the associated .R scripts, I settled on getfxn to be consistent with getunits but also think that fxn.R script is kinda vague so I am open to other naming suggestions 🤷‍♀️.

Hmm, I don't have an immediate better suggestion. It works.

> In data-raw there is a TODO because there is some hard coding that I don't love, here are a few potential courses of action

I see the TODO and responded to it. How does that relate to your courses-of-action list? They seem to be concerned with units/dates (I'm a little unclear on what problem exactly).

kdorheim:
I guess it would be nice to be able to the fxns table to generate the information in the output table see the vignette/articles/output.hml but that would be a bit more work I may open as a future development issue. Unless you and/or @leeyap have other concerns, then I think this PR can be merged

bpbond:
That would be my vote. A table with all the names and strings is hugely helpful, even if then the users need to go check about the units, etc.

`getfxn` does need to be documented somewhere. Is a logical place in a vignette we could add it?

bpbond (approved):

leeyap:
@kdorheim this is super helpful! I think it makes sense to return the function as a string, what do you mean by "return the actual function?"

bpbond:
Added @leeyap aș a formal reviewer, as I agree it'd be very useful to get her feedback.

leeyap (approved):

leeyap:
I agree that the table is very very useful, even without units since that's easy enough to check in R.
Otherwise, no comments from me, thanks @kdorheim!

kdorheim:
We could add a vignette, as of now though it is documented with rxoygen so help(getfxn) will generate documentation and it will automatically be added to the online pkgdown documentation.  I've added an example code to that documentation. Do you still think we need a vignette?

bpbond:
Not for now anyway, no. Good to go from my pov.

Issue #633, opened 2022-07-25
Add to Hector publications


bpbond:
https://www.sciencedirect.com/science/article/pii/S2590332222003268

Issue #563, opened 2022-02-14
Add hector_cmip citation


kdorheim:
Related to [DOECLIM integration](https://github.com/JGCRI/hector/pull/562#) #562, we will want to mint a version of hector_cmip6 on zenodo & include the doi citation where appropriate. Right now I think that is limited to the ocean_componet.cpp file.

kdorheim:
Places to add the citation 
* MEAN_TOS_TEMP in the oceanbox.hpp 
* Initial HL & LL ocean temp in ocean_componet.cpp

kdorheim:
Addressed in #658

Issue #539, opened 2021-12-15
Beta parameterization for v3


bpbond:
https://www.nature.com/articles/s41586-021-04096-9

bpbond:
@leeyap As part of your work could you quantify the difference between our default beta and one we can calculate from this paper?

leeyap:
![beta_diff](https://user-images.githubusercontent.com/87151968/150814846-f789716a-4644-4c78-9c79-84ace9fd6e79.jpg)
Were you thinking something like this?

bpbond:
Oh nice, yes! Could you include a line for our current default in red?

leeyap:
Like this?
![beta_diff](https://user-images.githubusercontent.com/87151968/150823222-4b5f9300-caaf-4135-8e10-b0302a050ad9.jpg)

bpbond:
@leeyap What's the β median you're using for your analysis?

leeyap:
I've been using the mean and s.d. of 0.54 ± 0.03

Issue #495, opened 2021-08-19
Add a citation file?


bpbond:
https://github.blog/2021-08-19-enhanced-support-citations-github/

ssmithClimate:
And we should make sure we are properly citing data and papers we use in hector.

kdorheim:
Closed with #658

PR #658, opened 2022-11-08
Addressing assorted issues on the way to v3

kdorheim:
This PR is a bit of a Hodge podge addressing remaining v3 issues 

* #633 add publication featuring Hector [documentation change]
* #563 add citation to leeya's cmip6 data [documentation change]
* #340 & #539 update default hector parameters [changes output behavior]
* #495 add citation file hat tip to @crvernon 

Due to the changes in the default parameters, we will see changes in Hector output behavior. I after we get Leeya bot results I will update the comp data just wanted to include a figure with the comparison with observations. 

image


kdorheim:
@leeyap and @bpbond any idea why the leeybot figures are not being pushed to the PR?

leeyap:
Hmm, weird. Looking at the check, it seems the graphs were posted here? https://github.com/JGCRI/hector/commit/19f09b37e133ff1fa5c140fea20104debf9f6193#commitcomment-89269587

Not sure why they aren't showing up on the PR, let me take a closer look...

leeyap:
Did the cml package update? What happens in https://github.com/JGCRI/hector/pull/658/commits/87f206f74a68af81d1f3923012161a11d0ec9b13?

kdorheim:
yea I got error messages stating that the previous commands had been deprecated and update to the commands that are now being used 🤷‍♀️ but may be that wasn't the correct thing to do

leeyap:
Hmm I see, yes, that should be right! 

Maybe try installing the specific version of cml?
image

leeyap:
@kdorheim Idea: I wonder if they changed the way you specify push, PR, etc. I glanced at the documentation and the example they show has this:
image
Which is different from how we have ours set up

kdorheim:
Alright so for future selves and those following along there had been an update to [cml](https://github.com/iterative/cml) which was causing the issues with leeya bot.

vignettes/articles/Hector-Wider-World.Rmd
```diff
@@ -1,7 +1,7 @@
 ---
 title: "Hector in the Wider World"
 author: "Leeya Pressburger"
-date: "7/30/2021"
+date: "11/7/2022"
```
bpbond, line 4: Change to auto-generated date?

bpbond (approved):
Looks good to me—thanks @kdorheim ! See one suggestion below.

vignettes/articles/Hector-Wider-World.Rmd
kdorheim: conflicted because the auto-update would update every time the HTML is generated not necessarily when the actual last update to the document content. I would probably opt to drop the date entirely over the auto-generate date.... thoughts?

bpbond: Oh, good point. I retract the suggestion — it's more important to document when _the content_ was updated, as you say. OK, then one other suggestion: change to YYYY-MM-DD?

kdorheim: sounds good to me!

kdorheim:
@bpbond "Command Line Hector / ubuntu (pull_request)" should we merge into dev v3 branch or address this failing test first?

bpbond:
Weird, I wonder what caused that. Up to you, but I'd vote fix here rather than start erroring `v3_dev`. I can take a look at it if useful—lmk.

bpbond:
@kdorheim You made this change:

Screen Shot 2022-11-10 at 8 05 14 PM

But when the `test_hector.sh` script tries to turn on a CO2 constraint (in line 57 of that script), it's looking for 
>sed 's/;[[:space:]]*CO2_constrain=csv:tables\/ssp245_emiss-constraints_rf.csv/

So it can't find the correct line, and by design dies in line 58.

bpbond:
Also it looks like the data files for `tas_constrain`, `RF_tot_constrain`, and `NBP_constrain` were removed, but their columns never made it into the various "emiss-constraints_rf.csv" files. So in those cases, the script can't find the data and dies. I am pushing the script naming fix but will let you handle this second issue!

bpbond:
@kdorheim I realized maybe this ☝️ may be unclear or grumpy-sounding (which I didn't intend!). Summarizing:
* The `sed` commands in the shell script are looking for a particular file pattern, which has now changed. I have updated them in d4ad069b
* But, some of them look for individual constraint test files: "tas_historical.csv" (line 80) and "nbp_output.csv" (line 94) for example. These no longer exist—they were removed in 37802731). They're explicitly _test_ files so it doesn't seem right to incorporate them into the "rcmip_ssp245_emiss-constraints_rf.csv" file...anyway, I wasn't sure how you wanted to organize this, so didn't fix.

kdorheim:
Sigh... yes deleting the constraint files was a mistake I will add them back into the appropriate directories. As for the file names ugh annoying I am trying to sort through some potential input problems & then I think it would be helpful to have a quick chat to about naming conventions & default Hector scenarios 😬

bpbond:
Re quick chat, before our Wednesday meeting? If so my calendar is up to date. Thanks

kdorheim:
@bpbond i don't understand the ` bash ./test_hector.sh ./src/hector` is passing on my local machine :(

bpbond:
@kdorheim It's actually not passing—but it's unfortunately easy to miss. On GA:



Nice and clear! On my local machine:



At first glance, no visible error message...**BUT normally this script ends with an "All done" message.** That doesn't appear, and at second glance, it started to test the RF constraint parameter, but then died via the `exit 1` in line 88, because it tried to `sed` a change but couldn't (i.e., it couldn't find the line).

THIS ISN'T OBVIOUS AT ALL. 🤦  Sorry! I have pushed 5f69d5ba so now this happens:

Screen Shot 2022-11-17 at 8 40 43 PM

...hopefully making it nice and obvious that an error occurred, and why.

PR #669, opened 2023-01-04
Add warning when preindustrial CO2, CH4, and N2O are changed from default values 

kdorheim:
As part of the update to using the AR6 RF equations, preindustrial CO2, CH4, and N2O should be treated as fixed parameters since in this PR I added log-level warnings for when these values are changed from their respective default values. 

See here 

``` 
> library(magrittr)
> here::here("inst", "input", "hector_ssp119.ini") -> ini
> core <- newcore(ini, suppresslogging = FALSE, loglevel = LL_WARNING())
> run(core)
```

Screen Shot 2023-01-04 at 9 18 49 AM

Compared with 

```
> setvar(core, dates = NA, var = PREINDUSTRIAL_N2O(), values = 100, unit = getunits(PREINDUSTRIAL_N2O()))
> reset(core)
Hector core:	Unnamed Hector core
Start date:	1745
End date:	2300
Current date:	1745
Input file:	/Users/dorh012/Documents/Hector-Versions/v3/hector/inst/input/hector_ssp119.ini
> run(core)
```
now returns 

Screen Shot 2023-01-04 at 9 20 02 AM


kdorheim:
Is this enough warning? we also have warnings in the ini files, what do you think @bpbond

src/ch4_component.cpp
```diff
@@ -98,6 +98,11 @@ void CH4Component::setData(const string &varName, const message_data &data) {
   try {
     if (varName == D_PREINDUSTRIAL_CH4) {
       H_ASSERT(data.date == Core::undefinedIndex(), "date not allowed");
+        if (data.getUnitval(U_PPBV_CH4).value(U_PPBV_CH4) != 731.41) {
+            H_LOG(logger, Logger::WARNING) << "Changing " << varName <<
+                " from default value, this is not recomended and may cause" <<
```
bpbond, line 103: Just to be my grammatically picky self: better use a semicolon here and below, not a comma. 😄

bpbond (changes requested):
Looks good to me and yes, I think that's enough warning.

Question: do we want to test in R that a warning is generated?

kdorheim:
re the testing in R... hmmm I don't know we don't really do any other log testing in R so other than "do the logs exist" so I am inclined to say no what do you think?

bpbond:
Makes sense. Thanks.

PR #672, opened 2023-01-21
R CMD CHECK cleanup

bpbond:
Trying to reduce the number of NOTEs and WARNINGs that are generated.

On my system we're down to

## NOTE -  installed size
 
>    installed size is  7.0Mb
     sub-directories of 1Mb or more:
       input   4.0Mb
       libs    2.1Mb

This is okay. One thing I noticed is that all the emissions files have values to 14 decimal places! FYI rounding everything to four decimal places cuts the file sizes by ~50%.

## NOTE - GNU make is a SystemRequirements

This is okay.

## NOTE - Files which contain pragma(s)

>   Files which contain pragma(s) suppressing diagnostics:
   ‘inst/include/logger.hpp’ ‘src/carbon-cycle-solver.cpp’
    ‘src/core.cpp’ ‘src/csv_outputstream_visitor.cpp’
    ‘src/csv_table_reader.cpp’ ‘src/csv_tracking_visitor.cpp’
    ‘src/forcing_component.cpp’ ‘src/ini_to_core_reader.cpp’
    ‘src/ocean_csys.cpp’ ‘src/simpleNbox-runtime.cpp’
    ‘src/simpleNbox.cpp’ ‘src/temperature_component.cpp’
    ‘src/unitval.cpp’

Not sure if this is okay or not. All of these pragmas relate to warnings coming from Boost 1.81.0_1.


bpbond:
By coincidence, on the R package development listserv this morning, posted by the `Rcpp` maintainer: 

>[R-pkg-devel] replacements of sprintf in compiled code
>
>Sometimes this sprintf compiler 'noise' is shrapnel from other packages we
include as headers, and harder to fix.  Brian Ripley kindly alerted us to one
remaining instance in the Rcpp headers which we fixed in October; the updated
package is currently awaiting its fate during its bi-annual update cycle but
is held in the dark and deep bowels of CRAN awaiting inspection (nine days
and counting now) as is package BH which in its annual update got a number of
headers refreshed from its upstream, plus an extra round of polish from
me. It too is available as an update in its git repo, and otherwise hiding in
those dark CRAN bowels.  I keep telling myself I should remain hopeful that
maybe one day we get to take advantage of it. Hopefully soon.

bpbond:
https://twitter.com/eddelbuettel/status/1617194358650064896

🎉

kdorheim:
Another option to decrease the inst size would be to limit the number of constraints included in the spp files... we could include the wmghgs and direct users to hectordata for the other ones.... _shrugh_

bpbond:
👍  I would say that's a problem for another (post-v3) day.

kdorheim (approved):
Ben I think this looks good thanks! added the variable to the units table we discussed last year

Issue #476, opened 2021-07-29
Final cleanup before v3 merge (running list)


bpbond:
- [x] Clean up GitHub Actions yaml files to remove development branch settings
- [x] Ensure all v3 milestone issues are linked, addressed, or reassigned to a new milestone
- [x] Search and resolve any `KRD-UPDATE` and `BBL-TODO` strings
- [x] Remove the Hector wiki as it is out dated & being replaced with the pkgdown documentation 
- [x] Reformat entire C++ code base to a common style (cf. outdated #418 )
- [x] Assertion of copyright (see Chris's email and google doc)
- [x] Authorship reworking
- [x] Update old Hartin graph on repository main page (and make auto-generated?)
- [x] Prune unneeded branches

Final Steps 
- [x] Once more through `clang-format`
- [x] Make release
- [x] doi on zenodo 
- [x] update NEWS
- [x] Update R package version number and any version numbers in C++
- [x] Inform rgieske and the pyhector folks when we have stable v3 beta

kdorheim:
@bpbond when you are referring to the GitHub Actions yaml files to remove the development branch settings are you talking about the `github_actions` branch listed in a few of the yaml files?

bpbond:
Right

bpbond:
Perhaps obvious, but I'm putting those in because do want testing on PRs against some of the dev branches, but don't want those as permanent settings.

kdorheim:
true!

kdorheim:
Do we need to do an alpha release before the final v3 release?

bpbond:
Not necessarily I don't think.

PR #674, opened 2023-01-21
Clean up TODOs

bpbond:
Removed duplicative or unneeded TODOs; left important (imo) ones

Checking related box in #476


kdorheim (commented):
I think this looks pretty good but what does [src/RcppExports-a74cb2a7.o.tmp](https://github.com/JGCRI/hector/pull/674/files#diff-4603b65448ed71965db52d416475abe40d7c49123df5d6b41baca2ecb050b3c3) do? It kind of looks like a temporary file that might have been committed by accident but want to check

bpbond:
🤦  doh my bad, will remove. Thanks for catching!

PR #677, opened 2023-01-25
Old rcmd.yml -> R-CMD-check.yaml (newest usethis)

bpbond:
* Change from old and semi-broken `rcmd.yml` to newest yaml generated by `usethis::use_github_action_check_standard()`
* Tweak branches in other yaml files


bpbond:
@kdorheim I didn't change `pkgdown.yaml` and am confused as to why it's failing

kdorheim:
Hmmm okay that is concerning and it looks like the error being thrown may be related to the problem we were seeing yesterday with the failing unit tests, for some reason some the relative path within the ini files to the csv tables are failing..... 
which is strange because it was working before the permafrost work was merged in. Which makes me wonder was there anything we changed in that branch that affected the way the core was read in? Or could there have been a change in the virtual environment we are running the tests/deployment on that led to this error?

bpbond:
>which is strange because it was working before the permafrost work was merged in. 

Well, it worked after too — look at the #676 checks

kdorheim:
Good point okay well I just relaunched the last running pkdown and it failed 

looking at the job setup I might have found the difference.... 

WORKING PKGDOWN JOB
```
Current runner version: '2.300.2'
Operating System
  Ubuntu
  [2](https://github.com/JGCRI/hector/actions/runs/3990041677/jobs/6843202699#step:1:2)2.0[4](https://github.com/JGCRI/hector/actions/runs/3990041677/jobs/6843202699#step:1:4).1
  LTS
```

FAILED PKGDOWN JOB 
```
Current runner version: '2.301.1'
Operating System
  Ubuntu
  [2](https://github.com/JGCRI/hector/actions/runs/4007696409/jobs/6880890122#step:1:2)2.0[4](https://github.com/JGCRI/hector/actions/runs/4007696409/jobs/6880890122#step:1:4).1
  LTS
```

could it be related to the change in the runner version?

.github/workflows/leeyabot.yml
```diff
@@ -1,9 +1,7 @@
 name: leeyabot
 on:
   pull_request:
-    branches:
-      - master
-      - v3_dev
+    branches: [main, master, v3_dev]
```
kdorheim, line 4: I think we can remove the v3_dev from here right? also we don't have a main so can delete that as well I think

.github/workflows/R-CMD-check.yaml
```diff
@@ -0,0 +1,52 @@
+# Workflow derived from https://github.com/r-lib/actions/tree/v2/examples
+# Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
+on:
+  push:
+    branches: [main, master, v3_dev]
+  pull_request:
+    branches: [main, master, v3_dev]
```
kdorheim, line 7: Can't we delete the v3_dev and main branch from here??

kdorheim: also can we add  ` workflow_dispatch:` so we can manually launch the test when needed it comes in handy sometimes and can we add back the block of code which triggers a scheduled run? that way if the main branch is left untouched for awhile it will still be run at least once a month as a precaution to check to see if changes in dependencies break hector?
```
  schedule:
   # * is a special character in YAML so you have to quote this string, it is set up to run at 6am
   # on the first day of the month, this is helpful for periods of time when PRs are not being merged.
    - cron:  '0 6 1 * *'
```

kdorheim (changes requested):

README.Rmd
```diff
@@ -17,8 +17,7 @@ library(ggplot2)
 
 # hector
 
-[![DOI](https://zenodo.org/badge/22892935.svg)](https://zenodo.org/badge/latestdoi/22892935) [![C++ unit tests](https://github.com/JGCRI/hector/actions/workflows/unit-testing.yaml/badge.svg)](https://github.com/JGCRI/hector/actions/workflows/unit-testing.yaml) [![Command Line Hector](https://github.com/JGCRI/hector/actions/workflows/command-line.yaml/badge.svg)](https://github.com/JGCRI/hector/actions/workflows/command-line.yaml) [![R-CMD](https://github.com/JGCRI/hector/actions/workflows/rcmd.yml/badge.svg)](https://github.com/JGCRI/hector/actions/workflows/rcmd.yml) [![codecov](https://codecov.io/gh/JGCRI/hector/branch/master/graph/badge.svg?token=EGM0lXDxRv)](https://codecov.io/gh/JGCRI/hector)
```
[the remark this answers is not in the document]
bpbond: Just a different name for the R CMD CHECK job — used to be "rcmd.yml" and now is "R-CMD-CHECK.yml"

bpbond:
>could it be related to the change in the runner version?

🤷  possible I guess? Not sure how we'd test this though.

kdorheim (approved):

PR #696, opened 2023-02-07
Correct the cff file

kdorheim:
@bpbond so there were errors in the cff file, which caused problems on the zenodo end of things. This PR is a documentation change that updates the cff with the zenodo release info however now those files are somewhat out of sink, but would enable the automatic zenodo release in the future so if we were to do a v3.0.1 think everything would be correct. 

Do we want to wait fro the v3.1.0 release or should we do a v 3.0.1 release?


bpbond (approved):
Thanks for handling this @kdorheim . I'd vote delete the release and re-do, I guess, rather than a 3.0.1.

Issue #698, opened 2023-03-01
Add global mean surface temperature as a Hector output 


kdorheim:
Right now Hector does not report global mean surface temperature as an output. It is possible to calculate this value using the sst and land air temperature. However, inorder to compare with historical values and IPCC reports, I have found myself calculating gmst time and time again. This makes me wonder if this should be an output reported by Hector instead of placing the burden on the user.

bpbond:
You're a pretty good test case, and agreed, if you're repeatedly calculating this makes sense.

PR #701, opened 2023-05-17
Enable global mean surface temperature as Hector output

kdorheim:
This PR addresses #698 by adding global mean surface temperature as a Hector output. Originally it was thought that this output variable would be redundant because it is something users can back-calculate from land tas and sst. RCMIP and IPCC regularly use this variable. Back-calculating it from output became annoying/cumbersome and the need for this output variable became apparent. 

I don't think we want to merge directly


bpbond (approved):
Looks great @kdorheim

Issue #657, opened 2022-11-07
Automate the input_params and output tables


kdorheim:
This idea mainly relates to the `vignettes/articles/output.Rmd` file where the table in the article has to be made manually with the `scripts/R_output.R`. However, with #656 we should be able to make this table using the `fxntable` that was created for the `getfxn` capabilities it would require that we have a script that generates an up-to-date `data-raw/input_params.csv`. 

Possible steps 
1. Add a script that produces `data-raw/input_params.csv` perhaps scraping from the information from an ini file this would replace the `data-raw/dave-input-params.R`
2. Modify the lookup-data.R scrip so that the units column has NAs were not applicable 
3. `vignettes/articles/output.Rmd` script should use the fxntable subset to not include inputs or functions associated with NA units. Then make the pretty R output table 
4. delete `scripts/R_outputs.R`

bpbond:
I assume this is not for v3? Added milestone.

PR #711, opened 2023-07-20
fix : convert trailing comments to single sline comments

gaelforget:
Seems safer since fewer parsers will support comments trailing section lines than single line comments.

With at least one parser,  [IniFile.jl](https://github.com/JuliaIO/IniFile.jl#readme), these comments lead to an incorrect read -- separating out the comments as single lines fixed it.


bpbond:
Thanks @gaelforget . What's the use case for this? I.e., why would another parser be reading Hector's INI files?

kdorheim:
@gaelforget, thanks for your interest in contributing to Hector! I guess I am curious as to why this change was only made for the `HFC4310_halocarbon`?

gaelforget:
> 

Other comments were single line comments or after variable definitions & did not seem to cause any issue -- unlike these few comments placed after section definitions (`[HFC4310_halocarbon ]`).

kdorheim:
This does not affect our miscellaneous scripts that are not tested as part of the CI (this is a todo #657) good to merge!

PR #714, opened 2023-08-29
Make the package versions consistent

kdorheim:
As it was pointed out in the GMD editor's feedback, there are some inconsistent package versioning numbers. I am not quite sure how it happened but with this PR all the version numbers should be consistent and then we can do a release with the correct number/delete the release with the incorrect versioning number? @bpbond what do you think?


bpbond (approved):
Looks good @kdorheim and I like your plan 👍

Issue #697, opened 2023-02-07
Change default branch from master to main 


kdorheim:
Upgrade from the antiquated master terminology by renaming the default branch as main. To do with will need to update the github action ymal files that specify the master branch from master to main.

PR #715, opened 2023-09-12
switch to the main branch

kdorheim:
This is to address #697 to change from the default branch from being master to main. To do this will need to update the github action files and documentation.


kdorheim:
@bpbond and @crvernon this should address the code changes fro #697 , I think. I think the outstanding TODOs would be to change the default branch on github and then the default branch on code cov. Do we do that now? Or wait until this is merged into master and then update? Does it need a new release associated with that change? What do you think?

.github/workflows/R-CMD-check.yaml
```diff
@@ -3,7 +3,7 @@ name: R-CMD-check
 on:
   push:
     branches:
-      - master
+      - main
```
bpbond, line 6: Or do we want to just add `- main` for now, do the branch switch, ensure everything working, and then finally delete the `- main` line?

bpbond (approved):
This seems fine; thanks @kdorheim

bpbond:
@kdorheim I'm not sure of the best sequence here, but I guess I would merge this into master and then update the branch name. Don't see the need for a new release since there's no code change.

kdorheim:
> @kdorheim I'm not sure of the best sequence here, but I guess I would merge this into master and then update the branch name. Don't see the need for a new release since there's no code change.

Yea I am also at a loss here 🤷‍♀️

PR #721, opened 2023-12-29
Enable permafrost by default in Hector

### LINKED ISSUE #675: Finish permafrost evaluation and integration
#655

bpbond:
Following #720 , this PR changes the model's INI files so that permafrost carbon is nonzero—i.e., enabled—by default. The parameterization follows that of [Woodard et al. 2021](https://gmd.copernicus.org/articles/14/4751/2021/gmd-14-4751-2021.pdf).

Adding permafrost produces a warmer model overall:

pf-ssp245

Specifically, for SSP2.45 we see +100-150 ppm CO2 and +0.5 °C with permafrost enabled. The permafrost pool loses about 250 Pg C, and the thawed pool peaks about ~130 Pg C. These results are consistent with Figure 3 in Woodard et al. (which used a much older version of the model and different parameterization).

As expected, there's almost no difference between running a single global biome and splitting into two biomes. This changes if we impose a 2x warming in the permafrost biome, as done by Woodard et al., but it seems cleaner and clearer for the _default_ model to use a single global biome.

Note: the early 1990s squiggle in the `permafrost_c` pool is due to the large Mt. Pinatubo cooling effect, which causes significant (re)freezing of soil carbon. This is expected. 

Closes #675


bpbond:
For a full context, here are diagnostics for two extreme SSPs:

pf-ssp119

pf-ssp585

bpbond:
Closing down for the long weekend. @kdorheim I have resolved most issues but a few R tests are still failing:

Screenshot 2023-12-29 at 5 42 14 PM


I will deal with those in 2024! 🥳

bpbond:
@kdorheim There is something subtle causing the model's climate to diverge, ever so slightly, when the we compare global versus biome-specific runs in `test-biome.R`. Ugh, it is being a complete PITA to track down. I will update when I know more.

bpbond:
Here's what is happening when we hit the test in `test-biome.R`. This runs the model with a single global biome (A), splits the biome into two, resets the model, runs again (B), and then compares A and B:

1-no_rounding

We're failing the test because the default tolerance is 1.5e-8 — very stringent. Above I've plotted with a tolerance of 1e-4 and you can see that by the end of the 21st century the model's climate has diverged 0.05 ppm CO2 and ~0.0003 degC between the two runs.

PR #733, opened 2024-02-23
Hector V3.2.0

kdorheim:
* Correct aerosol forcing coefficients based on Zelinka et al. (2023)
* Enable permafrost module and recalibrate model's default parameterization


kdorheim:
Each commit here represents a work that has undergone peer-reviewed PR.

bpbond (approved):

PR #739, opened 2024-04-08
Fix a minor numeric inconsistency in the hector vignette

mgondan:
Beta = 0.36 seems incorrect, isn't it 0.53?

````
beta <- fetchvars(core, NA, BETA())
beta
##              scenario year variable value      units
## 1 Unnamed Hector core   NA     beta  0.53 (unitless)
````

(...)
The current value is set to 0.36 (note that it is a unitless quantity, hence the (unitless) unit).

### DIFF
--- a/vignettes/hector.Rmd
+++ b/vignettes/hector.Rmd
@@ -113,11 +113,11 @@ beta
 The result of `fetchvars()` is always a `data.frame` with the same columns, even when returning a parameter value.
 Note also the use of `NA` in the second argument (`dates`).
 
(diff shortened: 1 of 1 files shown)


bpbond:
Good catch. Thanks for the contribution @mgondan !

mgondan:
The failure of the check seems to be unrelated (?)

kdorheim (approved):

kdorheim:
@mgondan thanks for catching that! the failed check is not a related issue will merge shortly

Issue #735, opened 2024-03-04
Values for 1745/1746


rgieseke:
There have been a few issues about this already (e.g. #354, #556, #177) 

The documentation currently states:

> An important note about time: The outputstream code currently writes the end date of the current timestep in the outputstream.csv file; e.g., when we’re running 1 January 1745 - 31 December 1745, what gets written is “1746”. This is not intuitive (most people think of “1746” denoting January 1-December 31) but numerically correct: the date is 1746.0, i.e. midnight on December 31 1745. A consequence is that carbon fluxes logged at this timepoint are effectively for the calendar year 1745 (1745-01-01 to 1745-12-31).

https://jgcri.github.io/hector/articles/outputstream.html?#output-stream-categories

As discussed in the linked issues, for NBP (formerly 'atm_land_flux') in simpleNbox it's not possible to retrieve a value for 1746 using RHector.

```R
packageVersion("hector")

‘3.2.0’

core <- newcore('hector_ssp126.ini')

fetchvars(core, 1745:2100, vars = c(NBP()))

Error: sendmessage: msg:  	Interpolation requested but not allowed (?) date: 1745

func: 	get
file: 	tseries.hpp
ffile:	../inst/include/tseries.hpp

line: 	332
```

With

```R
fetchvars(core, 1746:2100, vars = c(NBP()))
```

I get (as expected and documented):

```
fetchvars(core, 1746:2100, vars = c(NBP()))
               scenario year variable         value   units
1   Unnamed Hector core 1746      NBP -0.0802025172 Pg C/yr
2   Unnamed Hector core 1747      NBP -0.0762544017 Pg C/yr
```

However, in the command line outputstream I do get a value for 1745:

```
# Output from hector version 3.2.0 on Mon Feb 26 15:22:42 2024
year,run_name,spinup,component,variable,value,units
1745,ssp119,1,simpleNbox,NBP,0.0009932,Pg C/yr
1746,ssp119,0,simpleNbox,NBP,-0.0802,Pg C/yr
1747,ssp119,0,simpleNbox,NBP,-0.07625,Pg C/yr
```

For the Pyhector v3 update (https://github.com/openclimatedata/pyhector/pull/63) I need to find a way to handle NBP not reporting for 1746, which in my current setup requests all years for all variables and errors with the same error as RHector above
.
I can surely work around this somehow, but was wondering why the command line version reports a value for 1745.
Also, is NBP the only value for which the start year (1745) is not being reported?

[...]

Issue #734, opened 2024-02-28
Outputstream example files


rgieseke:
The [output directory readme](https://github.com/JGCRI/hector/blob/1ef0b4df8563a31580ffa94eb11b113e3f317b86/output/README.md)  mentions a sample file, which does not exist anymore I think.

> Included are example outputs from Hector driven with ssp245.

It appears sample files were moved during the R package restructuring:

https://github.com/JGCRI/hector/commit/dc3822305d26ecffa98f255bc91e1244b172ca7a#diff-ce56a62b01e03d9a0275e38b713c20bc2b09bdb2594056278a7150285def8a03

And then eventually deleted here:

https://github.com/JGCRI/hector/commit/d70a715d1f26eb49aee7d9a977df17b57d6cd0df

Could be useful to have an example file there to compare against a locally compiled version.

Issue #727, opened 2024-02-12
Documentation Issues


kdorheim:
Variable string names that need to be updated

- FCH4

Issue #718, opened 2023-11-08
Paper to add to Hector publications list


bpbond:
https://www.nature.com/articles/s41558-023-01803-4

via @cahartin - thank you!

bpbond:
Also via @cahartin - https://www.epa.gov/controlling-air-pollution-oil-and-natural-gas-operations/epas-final-rule-oil-and-natural-gas

PR #741, opened 2024-04-25
Addressing a handful of open github issues

kdorheim:
Commits in the PR are minor with respect to Hector's model behavior. These changes mostly impact documentation and such. 

The following github issues are addressed here 

- #718 (adding manuscripts to documentation pages)
- #727 (fixing the CH RF string to be consistent with the other RF types)
- #734 (add output streams)
- #740 (add links to Matilda! although this will probably need to be updated when Brown et al. is published in final form)
- #735 (add ref period to the plots)

Remaining to dos 
- @kdorheim wants to spell-check things first


.gitignore
```diff
@@ -64,7 +64,7 @@ libs/
 *.suo
 
 # Output
-output/*.csv
+output/tracking*.csv
```
bpbond, line 67: The problem with this as way to address #734 is that running an unmodified model will produce modified (from git's point of view) files, because of the timestamp at the top of the outputs...which I don't think is the behavior we want. It seems better to create an `example_outputs/` directory, perhaps as a subdirectory of `output/`.

README.Rmd
```diff
@@ -58,7 +58,8 @@ hector_tas_results$scenario <- ifelse(hector_tas_results$year <= 2016, "historic
 ggplot(hector_tas_results) +
     geom_line(aes(year, value, color = scenario), linewidth = 1) +
     theme_bw(base_size = 15) +
-    labs(color = NULL, x = NULL, y = expression("Temperature Anomaly ("~degree~"C)")) +
+    theme(axis.title = element_text(size=12))+
```
bpbond, line 61: Super picky, but please `size = 12`

bpbond (changes requested):
Looks good overall! I don't think we want to track the output files, though.

.gitignore
kdorheim: Hmmm that is a good point, do we want to add that or does it make sense to include the output streams as part of the release materials?

rgieseke:
Thanks for adding the temperature reference period, @kdorheim!
It's issue #736 though, the link in the initial post mentions #735 which was about 'Values for 1745/1746'.

kdorheim:
@rgieseke you are so right! Thanks for catching that!

bpbond (approved):

PR #745, opened 2024-07-01
Alpha and Volscl impact RF output

### LINKED ISSUE #743: RF scalar effect not reported in respective output
Adjusting RF using the volscl and alpha does not affect the RF_vol or RF aerosol results returned as output, it does have the intended effect the global temp. 

## Changing volscl 
image

## Changing alpha 

image

This seems incorrect; I would think that we would want the RF values to reflect what the temperature component is seeing... It is because we adjust the total forcing in the temperature component [here](https://github.com/JGCRI/hector/blob/1ef0b4df8563a31580ffa94eb11b113e3f317b86/src/temperature_component.cpp#L469) instead of in the forcing component.

kdorheim:
Should resolve #743


ptrscll:
**Description:** 
This pull request updates the output of Hector to correctly report changes in aerosol and volcanic radiative forcing caused by changes in alpha and volscl, resolving #743. This PR should not change the default run output.


**Previous Behavior:** 
As described in #743, changing volscl or alpha does not change any of the volcanic/aerosol radiative forcing values (or the total radiative forcing) output by Hector, despite different radiative forcing values being correctly used in temperature calculations. This PR seeks to resolve this issue.

The graphs below illustrate this issue for runs with different values of alpha/volscl. Note how temperature correctly changes but radiative forcing values remain the same.

old_vary_alpha
old_vary_vol



**New Behavior:**
	Within Hector, volscl and alpha have been switched to inputs to the forcing component rather than inputs to the temperature component. With this change, these parameters now correctly affect the output for their respective radiative forcing values. Note that we make the assumption that alpha uniformly impacts each of the five aerosols’ radiative forcings. 

Additionally, there are new unit tests that check to confirm that these two parameters have the expected impacts on their respective radiative forcings.

The graphs below show the new model behavior when varying alpha and volscl.
new_vary_alpha
image


**Comparison:**

Comparing the data frames from the default run between the dev and alpha_volcl versions of Hector, there was no change in any of the output values for total RF, ocean heat flux, global temperature, or CO2 concentrations. This is reflected in the following plot, which shows that the default runs produce the same results for both Hector versions.

image




When we adjust alpha or volscl, the new runs now reflect those changes in RF values but have identical temperature values. This is expected, since this PR is not meant to change the actual effect of alpha/volscl, just where it’s being applied. Graphs showing this trend are given below:

image

image

inst/include/component_data.hpp
```diff
@@ -65,6 +65,8 @@
 #define D_RHO_OC "rho_oc"            // BC radiative efficiency
 #define D_RHO_SO2 "rho_so2"          // SO2 radiative efficiency
 #define D_RHO_NH3 "rho_nh3"          // NH3 radiative efficiency
+#define D_AERO_SCALE "alpha"
+#define D_VOLCANIC_SCALE "volscl"
```
bpbond, line 69: These are both scalars; should they have a consistent naming structure? E.g. `aero_scalar` and `vol_scalar`?

project_files/Xcode/hector.xcodeproj/xcshareddata/xcschemes/ssp245.xcscheme
```diff
@@ -53,7 +53,7 @@
       </BuildableProductRunnable>
       <CommandLineArguments>
          <CommandLineArgument
-            argument = "./inst/input/hector_ssp245.ini"
+            argument = "/Users/dorh012/Documents/Hector-WD/hector/inst/input/hector_ssp245.ini"
```
bpbond, line 56: This absolute path on @kdorheim 's machine should be changed to a relative one.

bpbond (changes requested):
Nicely documented and an extremely sensible change! It looks great overall, though a few things to consider below.

project_files/Xcode/hector.xcodeproj/xcshareddata/xcschemes/ssp245.xcscheme
kdorheim: AH! That shouldn't have been committed! Sloppy on my part!

kdorheim:
I guess one question for @bpbond cause this is a code-breaking change it would probably be part of a V4 release... does that mean that it should go into a different dev branch? if we are thinking about doing to another V3 release? But that might be a bigger question about how we want to manage Hector releases and stuff

ptrscll:
I just pushed two commits to rename `alpha` to `aero_scalar` and `volscl` to `vol_scalar` as @bpbond suggested.

bpbond (approved):

PR #757, opened 2024-11-06
Feature/h2 ch4 lifetime

kdorheim:


kdorheim:
Unfortunately this PR has turned into a bit of a monster of H2 dev things and will probably end up being deleted. I have branched off of it to isolate the the dev better and will be submitting those are PRs instead.

kdorheim:
#758  and #775 are better implementations of this work.

PR #758, opened 2024-11-26
H2 effects on CH4 lifetime

kdorheim:
This PR is a WIP documentation is coming

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: hector
 Title: The Hector Simple Climate Model
-Version: 3.2.0
+Version: 3.3.0.9999999
 Authors@R: c(person("Kalyn", "Dorheim", 
 	        email = "kalyn.dorheim@pnnl.gov",
 	        role = c("aut", "cre"),
--- a/inst/include/oh_component.hpp
+++ b/inst/include/oh_component.hpp
@@ -71,6 +71,8 @@ private:
   double CNMVOC; // coefficent for NMVOC
   double CNOX;   // coefficent for NOX
   double CCH4;   // coefficent for CH4
+  double CH2;   // coefficent for CH4
+
 
   // logger
   Logger logger;
--- a/inst/include/oh_component.hpp
+++ b/inst/include/oh_component.hpp
@@ -71,7 +71,7 @@ private:
   double CNMVOC; // coefficent for NMVOC
   double CNOX;   // coefficent for NOX
   double CCH4;   // coefficent for CH4
-  double CH2;   // coefficent for CH4
+  unitval CH2;   // coefficent for CH4
 
 
   // logger
--- a/inst/input/hector_ssp119.ini
+++ b/inst/input/hector_ssp119.ini
@@ -138,6 +138,7 @@ CNOX=0.0042			; coefficent for NOX
 CCO=-0.000105		; coefficent for CO
 CNMVOC=-0.000315	; coefficent for NMVOC (non methane VOC)
 CCH4=-0.32			; coefficent for CH4
+CH2=-0.00044625 ; coefficent for H2
 
 ;------------------------------------------------------------------------
 [ozone]
--- a/src/ch4_component.cpp
+++ b/src/ch4_component.cpp
@@ -165,7 +165,7 @@ void CH4Component::run(const double runToDate) {
         << std::endl;
 
 // Permafrost thaw produces CH4 emissions
-#define PG_C_TO_TG_CH4 (1000.0 * 16.04 / 12.01)
+    const double PG_C_TO_TG_CH4 (1000.0 * 16.04 / 12.01);
     const double rh_ch4 =
         core->sendMessage(M_GETDATA, D_RH_CH4).value(U_PGC_YR) * PG_C_TO_TG_CH4;
 
--- a/src/oh_component.cpp
+++ b/src/oh_component.cpp
@@ -57,12 +57,15 @@ void OHComponent::init(Core *coreptr) {
 
   // Inform core what data we can provide
   core->registerCapability(D_LIFETIME_OH, getComponentName());
+  core->registerCapability(D_COEFFICENT_H2, getComponentName());
   // Register inputs accepted.  Note that more than one component
   // can accept an input
   core->registerInput(D_EMISSIONS_CO, getComponentName());
   core->registerInput(D_EMISSIONS_NMVOC, getComponentName());
   core->registerInput(D_EMISSIONS_NOX, getComponentName());
   core->registerInput(D_EMISSIONS_H2, getComponentName());
+  core->registerInput(D_COEFFICENT_H2, getComponentName());
+
 
 }
 
@@ -178,8 +181,8 @@ void OHComponent::run(const double runToDate) {
          CH2 *
           ((1.0 * +current_h2) -
            H2_emissions.get(H2_emissions.firstdate()).value(U_TG_H2));
-      
-      
+
+
     toh = a + b + c + d + e;
     H_LOG(logger, Logger::DEBUG)
         << "Year " << runToDate << " toh = " << toh << std::endl;
@@ -212,6 +215,9 @@ unitval OHComponent::getData(const std::string &varName, const double date) {
     H_ASSERT(date != Core::undefinedIndex(),
              "Date required for NMVOC emissions");
     returnval = NMVOC_emissions.get(date);
+  } else if (varName == D_COEFFICENT_H2) {
+      H_ASSERT(date == Core::undefinedIndex(), "Date not allowed for ECS");
+     returnval = CH2;
   } else if (varName == D_EMISSIONS_H2) {
       H_ASSERT(date != Core::undefinedIndex(), "Date required for H2 emissions");
       returnval = H2_emissions.get(date);
--- a/src/oh_component.cpp
+++ b/src/oh_component.cpp
@@ -149,6 +149,8 @@ void OHComponent::run(const double runToDate) {
   unitval current_nox = NOX_emissions.get(runToDate);
   unitval current_co = CO_emissions.get(runToDate);
   unitval current_nmvoc = NMVOC_emissions.get(runToDate);
+  unitval current_h2 = H2_emissions.get(runToDate);
+
 
   // get this from CH4 component, this is last year's value
   const double previous_ch4 =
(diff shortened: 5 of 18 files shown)


kdorheim:
Some specific questions for @ssmithClimate & @bpbond 

* `COEFF_H2()` - do we like this function name for the H2 coefficient to be added to the OH lifetime equation? Do we think it is worth having it exposed as an R function? Or should it be something users would need to change in the ini file? Should we expose the other coefficient R bindings as well? 
* Right now in the oh_component.cpp all of the coefficients in what corresponds to equations (1-5) are undefined unitvals, is that okay? Or do we want to define them as something?

DESCRIPTION
kdorheim, line 1: @bpbond just a heads up I did increase this for my own sanity purposes related to the gcam-hector coupling...

src/ch4_component.cpp
```diff
@@ -165,7 +165,7 @@ void CH4Component::run(const double runToDate) {
         << std::endl;
 
 // Permafrost thaw produces CH4 emissions
-#define PG_C_TO_TG_CH4 (1000.0 * 16.04 / 12.01)
+    const double PG_C_TO_TG_CH4 (1000.0 * 16.04 / 12.01);
```
kdorheim, line 168: @bpbond is this the correct usage of the const double?

src/oh_component.cpp
```diff
@@ -169,7 +177,13 @@ void OHComponent::run(const double runToDate) {
...
         ((1.0 * +current_nmvoc) -
          NMVOC_emissions.get(NMVOC_emissions.firstdate()).value(U_TG_NMVOC));
-    toh = a + b + c + d;
+    const double e =
+         CH2 *
+          ((1.0 * +current_h2) -
+           H2_emissions.get(H2_emissions.firstdate()).value(U_TG_H2));
+
+
+    toh = a + b + c + d + e;
```
kdorheim, line 186: @bpbond do we have strong feelings on the use of e as a variable name?

src/ch4_component.cpp
kdorheim: @kdorheim revert back to #define

src/oh_component.cpp
kdorheim: switch to f, we would like to be consistent with the notation above

bpbond:
> * `COEFF_H2()` - do we like this function name for the H2 coefficient to be added to the OH lifetime equation? Do we think it is worth having it exposed as an R function? Or should it be something users would need to change in the ini file? Should we expose the other coefficient R bindings as well?

Unless we reasonably expect that users will want to run sensitivity tests or otherwise manipulate these parameters, I'd vote to start with NOT exposing them via R or the INI file.

Name: `COEFF_H2` is kind of generic; can we be more specific with this name?

> * Right now in the oh_component.cpp all of the coefficients in what corresponds to equations (1-5) are undefined unitvals, is that okay? Or do we want to define them as something?

When you say "undefined", you mean they have no units? That's OK, but if you know the associated units, we might as well include them for clarity.

DESCRIPTION
```diff
@@ -1,6 +1,6 @@
 Package: hector
 Title: The Hector Simple Climate Model
-Version: 3.2.0
+Version: 3.3.0.9999999
```
bpbond, line 3: ?

inst/include/oh_component.hpp
```diff
@@ -71,6 +71,8 @@ class OHComponent : public IModelComponent {
   double CNMVOC; // coefficent for NMVOC
   double CNOX;   // coefficent for NOX
   double CCH4;   // coefficent for CH4
+  unitval CH2;   // coefficent for CH4
```
bpbond, line 74: Do you mind correcting the spelling of "coefficient" here and in previous lines?

inst/input/hector_ssp119.ini
```diff
@@ -138,6 +138,7 @@ CNOX=0.0042			; coefficent for NOX
 CCO=-0.000105		; coefficent for CO
 CNMVOC=-0.000315	; coefficent for NMVOC (non methane VOC)
 CCH4=-0.32			; coefficent for CH4
+CH2=-0.00044625 ; coefficent for H2
```
bpbond, line 141: Same spelling note; also, is there a reference/source for this value? If so please note it.

src/oh_component.cpp
```diff
@@ -57,12 +57,15 @@ void OHComponent::init(Core *coreptr) {
 
   // Inform core what data we can provide
   core->registerCapability(D_LIFETIME_OH, getComponentName());
+  core->registerCapability(D_COEFFICENT_H2, getComponentName());
```
bpbond, line 60: D_COEFFICIENT_H2

tests/testthat/test-h2.R
```diff
@@ -25,9 +25,9 @@ test_that("tau ", {
 
 test_that("H2 emissions ", {
 
-    # Check to make sure that can fetch and set the H2 emissions although
-    # at this point changing the emissions will have no impact on
-    # [ch4] dynamics but in the future it should...
+    # Check to make sure that can fetch and set the H2 emissions and
+    # that the effects of adding some H2 emissions change thigns
+    # now that we have the CH4 indrect effects implement.
```
bpbond, line 30:
```c++
# that the effects of adding some H2 emissions change things
# now that we have the CH4 indirect effects implemented
```

bpbond (changes requested):
Overall looks good! THANK YOU @kdorheim

inst/input/hector_ssp119.ini
ssmithClimate: Was there some derivation needed for this, or was this coefficient in the right format from the original source? (And how similar were the coefficients for other emissions to the ones we are using now?)

src/oh_component.cpp
ssmithClimate: Single letter variables seem overly cryptic to me. Perhaps something like this: 
toh =  toh_ch4 + toh_nox ... + toch_h2
(or capitalize if that is your style).

Also, suggest putting a comment as such above:

// toh is a sum of coefficients in an exponential representation of OH lifetime
double toh = 0.0;

[Just so that readers are clear without reading the next page of code what this is.]

ssmithClimate:
@kdorheim  - just a couple comments, this is great to see!

src/oh_component.cpp
bpbond: +1 to the point about single-letter variables.

inst/include/oh_component.hpp
kdorheim: Other questions: double vs unitval? Do we care? When should we use one over the other?

kdorheim: > If it's used strictly within the component, and pretty isolated, then sometimes unitval is overkill -- feel free to use a double imho. If it's coming from user (INI or R) then having explicit units attached can be important. Does that help?

so all of the coefficients are read in via ini file/R.... I would assume that we would want to be consistent with out treatment of them but also don't think that any of these coefficients have units that are easy to interpret

bpbond: If it's used strictly within the component, and pretty isolated, then sometimes unitval is overkill -- feel free to use a double imho. If it's coming from user (INI or R) then having explicit units attached can be important. Does that help?

inst/input/hector_ssp119.ini
kdorheim: Okay there was a derivation that was needed for this parameter, which will be described in detail in the manuscript associated with this work although that is a TBD, so I have included a TODO-H2 tag which is linked to the H2 project as a reminder to address this issue before merging into main or anything. 

@ssmithClimate Could you elaborate on this a bit more? (And how similar were the coefficients for other emissions to the ones we are using now?) I am not sure if I understand what you are asking here. It is also unclear where the original coefficients came from @ssmithClimate do you know the source of those?

kdorheim:
> > * `COEFF_H2()` - do we like this function name for the H2 coefficient to be added to the OH lifetime equation? Do we think it is worth having it exposed as an R function? Or should it be something users would need to change in the ini file? Should we expose the other coefficient R bindings as well?
> 
> Unless we reasonably expect that users will want to run sensitivity tests or otherwise manipulate these parameters, I'd vote to start with NOT exposing them via R or the INI file.
> 
> Name: `COEFF_H2` is kind of generic; can we be more specific with this name?
> 
> > * Right now in the oh_component.cpp all of the coefficients in what corresponds to equations (1-5) are undefined unitvals, is that okay? Or do we want to define them as something?
> 
> When you say "undefined", you mean they have no units? That's OK, but if you know the associated units, we might as well include them for clarity.

Okay I can remove the coeff from the R bindings but all of the coefficients have been defined in the ini file I think since V1

bpbond:
>Okay I can remove the coeff from the R bindings but all of the coefficients have been defined in the ini file I think since V1

I'm confused; this may be easier to talk through in person. Anyway, `H2_COEFF` is fine if you want to keep it; we have plenty of other parameters whose function isn't immediately obvious from their name.

inst/input/hector_ssp119.ini
ssmithClimate: Sorry for the delay. The coefficients in hector now ultimately derive from this paper:
Wigley, T.M.L., Steven J. Smith, and M.J. Prather (2002) Radiative Forcing due to Reactive Gas Emissions. Journal of Climate 15(18), pp. 2690–2696

Although the hector paper quotes the source as :
Tanaka, K., Kriegler, E., Bruckner, T., Hooss, C., Knorr, W., and Raddatz, T.: Aggregated Carbon Cycle, Atmospheric Chemistry, and Climate Model (ACC2) – description of the forward and inverse models, 1–188, Max Planck Institute for Meteorology, Hamburg, Germany, 188, 2007.
(which is a more indirect reference)

So my question was - is the overall chemistry representation of the source we're drawing the H2 coefficients from is generally consistent with the chemistry representation used now in Hector. If not it's something to flag (e.g., means we ultimately might need to update more than H2) since the two things we're merging together are not entirely consistent with each other. (Given uncertainty its likely fine, but we would want to flag this at least.)

So the way I thought to evaluate that was to look if the relative magnitude of any common coefficients for the non-H2 components of the equations that were used in the H2 work we're using now and the original equation we use now in Hector were similar. For example is the ratio of the impact of NOx to impact of CH4 or NOx to CO in the original hector (e.g. Wigley et al, Tanaka et al) are similar to what's in the H2 papers we are citing (to the extent they contain these other effects).

For example, in hector we have something like this for the CH4 OH lifetime (Hartin et al. 2015):

ln(OH)t = −0.32(ln[CH4 ]t − ln[CH4 ]t 0 ) + 0.0042(E(NOx)t ) − (E(NOx)t0) − 0.000105(E(CO)t ) − (E(CO)t0) − 0.00315(E(VOC)t ) − (E(VOC)t0)

So the impact of NOx vs CO is: 0.0042 / 0.000105 = 40 times (on an emissions basis in this case). Or we could look at the relative impact of NOx to CH4 in this equation if that's what's common between what's in hector now and what is represented in the equations from the paper's we are drawing from for the new H2 components.

To do this we need to have both sets of equations cast into an equivalent form.

ssmithClimate: @kdorheim comments here!

Issue #760, opened 2024-12-03
broken link on repo


cahartin:
The code of conduct link and contributors link is broken. the files are in the repo but looks like the path is incorrect.

`The main branch of hector does not contain the path articles/ContributionsGuide.html.`

bpbond:
Ah, as linked from `README.md`. Thank you @cahartin !

PR #761, opened 2024-12-04
Fix broken links #760

bpbond:


man/figures/README-tas_plot-1.png
kdorheim, line 1: Is this the same figure or is there a change is there a change?

bpbond:
The figure was regenerated but should be identical...

Screenshot 2024-12-04 at 12 20 12 PM

README.Rmd
```diff
@@ -86,8 +86,8 @@ interface to Hector
 ## Contributing to Hector
 
 The Hector team welcomes and values community contributions, but please see 
-our [Contribution Guide](articles/ContributionsGuide.html) and note by  contributing 
-to this project, you agree to abide to our [Contributor Code of Conduct](CODE_OF_CONDUCT.html)
+our [Contribution Guide](vignettes/articles/ContributionsGuide.Rmd) and note by  contributing
```
kdorheim, line 89: Hmm so the reason why this was `articles/ContributionsGuide.html` was so that the linking worked on the homepage of the manual. This change seems to break that link though... what about switching to `https://jgcri.github.io/hector/articles/ContributionsGuide.html`?

kdorheim (changes requested):
Hi Ben, thanks for catching this broken links please see comments below.

bpbond:
Thanks @kdorheim ! And to be clear, credit to @cahartin on catching the problem.

kdorheim (approved):

PR #764, opened 2024-12-20
Fix permafrost refreeze typo 

### LINKED ISSUE #763: question about permafrost refreeze over biomes
Hi,

First of all, I thank you for sharing your work with the open-source climate modeling community!

I am currently working on implementing (/trying to) the model in Python, and these [lines] in _simpleNbox-runtime.cpp_ (https://github.com/JGCRI/hector/blob/a8247808f58df0e620bd61dfc91096b827f6d84b/src/simpleNbox-runtime.cpp#L857-L866) caught my attention - in particular the line 865.

My question : I was wondering why, during the sum over biomes, the iterative computation of refreezed soil pool (pf_refreeze_soil) relies on the permafrost thaw pool (pf_refreeze_tp) of current biome instead of the refreezed soil pool of previous biome ?

First I thought this was linked to issue #723 but I was not able to reproduce the mentioned error (comparison with 1 biome vs biome split) after reactivating the permafrost refreeze from soil feature, I guess I am doing something wrong. 
In the other hand, I was thinking that maybe the refreezed tp is somehow part of the refreezed soil pool? but in this case, why refreezed soil and refreezed tp are added during permafrost derivatives computation a few lines after?

My apologies if this is a trivial one, but I didn't manage to find the answer !

Regards
Pierre-Jean

bpbond:
Should this be against `main` or `dev`?

Closes #763

### DIFF
--- a/src/simpleNbox-runtime.cpp
+++ b/src/simpleNbox-runtime.cpp
@@ -862,7 +862,7 @@ int SimpleNbox::calcderivs(double t, const double c[], double dcdt[]) const {
       pf_refreeze_tp =
           pf_refreeze_tp + fluxpool(biome_pf_refreeze_tp, U_PGC_YR);
       pf_refreeze_soil =
-          pf_refreeze_tp + fluxpool(biome_pf_refreeze_soil, U_PGC_YR);
+        pf_refreeze_soil + fluxpool(biome_pf_refreeze_soil, U_PGC_YR);
     }
   }
 


kdorheim:
Okay my vote would be to merge into dev

kdorheim:
For Record Keeping Purposes - this PR caused some behavior changes in Hector causing the old-new tests to fail. 


This table indicates the $R^2$ and normalized root mean squared error values for each variable. An $R^2$ of 1 and NRMSE of 0 indicate that the new values are identical to the values being used in the old-new tests 

The current pull request's outputs differ from 3.1.1 (d931a004) as follows:


|  Variable                | R squared| NRMSE|
|:-----------------|---------:|-----:|
|CO2_concentration |     1.000| 0.001|
|global_tas        |     1.000| 0.196|
|RF_tot            |     1.000| 0.321|


Comparing this PR with the previous release of Hector 


comp_plot_release





diff_plot_release

The differences are more pronounced at the end of the run, which makes sense since this bug fix affects permafrost dynamics, which are more pronounced with more warming aka latter on into a future scenario run. 

Note - these figures were created with the leeyabot reboot

Issue #585, opened 2022-03-18
Consolidate GA files?


bpbond:
The GitHub Actions `leeyabot` and `test-coverage` have very similar setups, both needing R, libcurl, and a few packages. Maybe consolidate them into a single yaml file? That would save on GA resources.

kdorheim:
It also looks like there might be a way to cache builds/environments that way we would only need to build C++ once then pass to google tests and also build R Hector once the pass to leeyabot/code coverage testing

kdorheim:
On that note we should figure out which GA need to be run every PR/ semi-regularly to catch dependency changes are they're any GA that could be manually triggered (does leeya bot need to run every time or could it be manually triggered when the PR is ready for review)?

PR #770, opened 2025-02-26
Github Actions

kdorheim:
I have streamlined the GA files (#585) as much as possible but have mostly changed how frequently they are triggered. 

* main - monthly, any push to and pr opened to 
* all other branches will have the option to be triggered manually 
* leeyabot now compares against the old new testing data and also minted release results - this work was cherry picked from @ptrscll #747


.github/workflows/leeyabot.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+      - name: Install System Dependencies (including Pandoc)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y pandoc
+          sudo apt-get install libcurl4-openssl-dev
+          sudo apt-get install -y libjs-mathjax
+
+      - name: Cache R Packages
+        uses: actions/cache@v3
```
bpbond, line 27: It might be simpler to use `r-lib/actions/setup-r-dependencies` instead of all this stuff below? See here for an example: https://github.com/COMPASS-DOE/sensor-data-pipeline/blob/main/.github/workflows/check-workflow.yaml

bpbond (approved):
Nice work -- thanks for tackling this @kdorheim ! See one comment below.

PR #775, opened 2025-03-19
Indirect Effects of H2: stratospheric water vapor

kdorheim:
This PR intends to address the issue [#6](https://github.com/JGCRI/Hector_H2_Project/issues/6). It adds the indirect effects of $H_2$ emissions on stratospheric water vapor. 

(S11) from Dorheim et al. 2024 for our previous $RF_{strat_{H_2O}}$ equation, we had 
Screenshot 2025-03-19 at 3 43 48 PM

which only accounts stratospheric $H_2O$ vapor created from methane oxidation with OH radicals. However, as part of the indirect climate effects of $H_2$, we also have $H_2$ emissions contributing to $RF_{strat_{H_2O}}$. The new implementation of $RF_{strat_{H_2O}}$ is now 

Screenshot 2025-03-19 at 3 44 42 PM

Where the coefficient of $0.00019$ W m −2 per Tg yr− 1 is the multi-model mean from Sand et al. 2023. 

Since $H_2$ emissions are set to a default value of 0 at the moment, there is no real change for any of the default Hector runs. However, we can do some emission impulse runs to take a look at how this model development affects results. 

If `CH2` (the coefficient for H2 emission effects on the CH4 OH sink) is set to 0, we have isolated the effects of the $H_2$ emissions impulse from impacting only RF stratospheric water vapor, which is illustrated in the panel below. 

unnamed-chunk-4-1

Now if we calculate the $GWP_{100}$ we get a value of 2.345 . Looking at Sand et al. Supplementary Table 6 the GWP100 of hydrogen due to strat H2O ranges from 0.9 to 3.2 with a model mean of 2.1, which is not too far off what we are getting from Hector now! 

unnamed-chunk-9-1

****
Materials used to make PR figures - [PR-775.zip](https://github.com/user-attachments/files/19350307/PR-775.zip)


kdorheim:
Questions for @ssmithClimate and @patrickrorourke, would we want to be able to consider uncertainty as being this efficient? If so then we might want to make this a free parameter that can adjusted via the ini file but it would not change results.

ssmithClimate:
Looks good!

Certainly having parameters be user-adjustable is desirable if possible.

bpbond (approved):
Sorry I've been so out of the loop this week.

This looks good, and thanks for the clear and comprehensive documentation @kdorheim .

>W m-2 per Tg yr-1

This makes my head hurt. A watt is a joule per second so doesn't the time aspect completely cancel out?!? Anyway, whatevs, if your source paper uses that it's fine.

ssmithClimate:
Yes @bpbond W/m^2 is the standard unit used in the field - apologies to your head :)

bpbond:
W/m2, sure, totally standard. But W/m2/Tg/yr?!? 🤕

ssmithClimate:
Think of it this way, is it more useful to describe an Oreo as cookie with cream filling (even though the ingredients overlap) or by the unique set of ingredients :)

kdorheim:
Now users can adjust the H2 RF coefficient for Start H2O Vapor via the ini file/or R. Note that I did not add any of the R helper functions but have included an example of how to adjust this parameter value in the PR demo. 

```
hc <- newcore(ini_file, name = "default rho H2O H2 coeff.")
rho1 <- fetchvars(hc, NA, "rho_h2o_h2")
```
will return 

|scenario                  | year|variable   |   value|units   |
|:-------------------------|----:|:----------|-------:|:-------|
|default rho H2O H2 coeff. |   NA|rho_h2o_h2 | 0.00019|W/m2/Tg |

Here are some results from perturbation runs using different coefficient values. 
unnamed-chunk-10-1

@ssmithClimate does this address your comments? 


[PR-775.zip](https://github.com/user-attachments/files/19592047/PR-775.zip)

PR #777, opened 2025-03-25
Warning fixes when compiling with g++

bje-:
Two changes to eliminate warnings from `g++`:

* Use references for exception objects per the C++ Core Guidelines:
   https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Re-exception-ref
* Use `vector<Core *>::size_type` as index type for `getcore` and `delcore` methods to avoid signed/unsigned comparisons.


pralitp (approved):
Thanks, always nice to eliminate warnings.  This all looks good with me.

bpbond (approved):
Thanks @pralitp and thank you @bje-  -- I learned a few things. Over to you @kdorheim

PR #780, opened 2025-04-22
 Indirect effects of H2 on trop O3

kdorheim:
This should address https://github.com/orgs/JGCRI/projects/7?pane=issue&itemId=72654984&issue=JGCRI%7CHector_H2_Project%7C4, implement the indirect effects of H2 emissions on trop O3 burden.

### DIFF
--- a/inst/include/unitval.hpp
+++ b/inst/include/unitval.hpp
@@ -82,6 +82,7 @@ enum unit_types {
   U_TG_NMVOC,
   U_TG_H2,
   U_DU_O3,
+  U_DU_O3_TG,
   U_GG_S, // SO2 emissions in Gg-S/yr
 
   U_TG_PPBV, // Conversion for CH4 and N2O emission to concentrations
--- a/src/o3_component.cpp
+++ b/src/o3_component.cpp
@@ -62,6 +62,11 @@ void OzoneComponent::init(Core *coreptr) {
   core->registerInput(D_EMISSIONS_CO, getComponentName());
   core->registerInput(D_EMISSIONS_NMVOC, getComponentName());
   core->registerInput(D_EMISSIONS_NOX, getComponentName());
+  core->registerInput(D_EMISSIONS_H2, getComponentName());
+  // Parameters read in from the ini
+  core->registerInput(D_COEFF_O3_H2, getComponentName());
+  core->registerInput(D_PREINDUSTRIAL_O3, getComponentName());
+
 }
 
 //------------------------------------------------------------------------------
@@ -94,6 +99,9 @@ void OzoneComponent::setData(const string &varName, const message_data &data) {
     if (varName == D_PREINDUSTRIAL_O3) {
       H_ASSERT(data.date == Core::undefinedIndex(), "date not allowed");
       PO3 = data.getUnitval(U_DU_O3);
+    } else if (varName == D_COEFF_O3_H2) {
+      H_ASSERT(data.date == Core::undefinedIndex(), "date not allowed");
+      CO3_H2 = data.getUnitval(U_DU_O3_TG);
     } else if (varName == D_EMISSIONS_NOX) {
       H_ASSERT(data.date != Core::undefinedIndex(), "date required");
       NOX_emissions.set(data.date, data.getUnitval(U_TG_N));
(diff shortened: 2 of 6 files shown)


src/o3_component.cpp
```diff
@@ -62,6 +62,11 @@ void OzoneComponent::init(Core *coreptr) {
   core->registerInput(D_EMISSIONS_CO, getComponentName());
```
kdorheim, line 62: @bpbond before I wrap up the changes for this PR I do have a question, looking at the [src/o3_component.cpp](https://github.com/JGCRI/hector/blob/a61d64af3b9cc6d0d63cd661ee0164067a8f0165/src/o3_component.cpp#L60) 

```
  // Register the emissions we accept (note that OH component also
  // accepts these.  That's ok
  core->registerInput(D_EMISSIONS_CO, getComponentName());
  core->registerInput(D_EMISSIONS_NMVOC, getComponentName());
  core->registerInput(D_EMISSIONS_NOX, getComponentName());
  core->registerInput(D_EMISSIONS_H2, getComponentName());
```
which makes me think that we can also read in the H2 emissions a second time here? Is that how we want to do this? Or should these emissions be passed as a dependency from the OH component?

bpbond: @kdorheim No, I agree, it seems cleanest to have it as a direct input here too. 

That surprises me a little! I didn't realize (or forgot) the model could route inputs to multiple components.

inst/include/unitval.hpp
```diff
@@ -82,6 +82,7 @@ enum unit_types {
   U_TG_NMVOC,
   U_TG_H2,
   U_DU_O3,
+  U_DU_O3_TG,
```
bpbond, line 85: What's the unit here, exactly? Might be worth an explanatory comment.

src/o3_component.cpp
```diff
@@ -131,12 +142,13 @@ void OzoneComponent::run(const double runToDate) {
...
   unitval current_co = CO_emissions.get(runToDate);
   unitval current_nmvoc = NMVOC_emissions.get(runToDate);
+  unitval current_h2 = H2_emissions.get(runToDate);
   unitval current_ch4 = core->sendMessage(M_GETDATA, D_CH4_CONC, runToDate);
 
   O3.set(runToDate,
          unitval((5 * log(current_ch4)) + (0.125 * current_nox) +
-                     (0.0011 * current_co) + (0.0033 * current_nmvoc),
-                 U_DU_O3));
+                     (0.0011 * current_co) + (0.0033 * current_nmvoc) +
```
bpbond, line 150: Is there documentation on the source of these coefficients?

bpbond (approved):
Sorry for the delay!

src/o3_component.cpp
kdorheim: are you talking about the 0.0011, 0.0033, 0.125, and 5?

inst/include/unitval.hpp
kdorheim: Is it O3 Dobson Units per Tg emissions (in this case it is H2 emissions)

src/o3_component.cpp
bpbond: Sorry, yes

src/o3_component.cpp
kdorheim: So leave it as is?

bpbond: That's my vote, yes

src/o3_component.cpp
kdorheim: Gah! sorry I forgot to finish my reply! So those coefficients came from our implementation of Tanaka 2007 from Wigley 2002. But long story is that there is an implementation problem (similar to what happened with the methane lifetime wrt OH equation) so these coefficients might need to be changed



Tanaka, K., Kriegler, E., Bruckner, T., Hooss, G., Knorr, W., & Raddatz, T. (2007). Aggregated carbon cycle, atmospheric chemistry, and climate model (ACC2). Reports on Earth Systems Science, 40. https://www.semanticscholar.org/paper/Aggregated-Carbon-cycle%2C-atmospheric-chemistry-and-Tanaka-Kriegler/78a873db5b1d1410a68165f1fdf1873beba4e6db

Wigley, T. M. L., Smith, S. J., & Prather, M. J. (2002). Radiative Forcing Due to Reactive Gas Emissions. Journal of Climate, 15, 2690–2696. https://doi.org/10.1175/1520-0442(2002)015&#60;2690:RFDTRG&#62;2.0.CO;2

bpbond: Thanks for the extra detail @kdorheim

Issue #742, opened 2024-05-31
Unexpected Behavior from volscl


ptrscll:
**Describe the bug**
I was practicing changing parameters in Hector and found that when I decreased the volcanic scaling factor from 1.0 to 0.5, temperatures and CO2 concentrations decreased in the future and across most of the 1900s.

**To Reproduce**
In the R interface, I tried running the ssp245 scenario and comparing the results using the default parameters with the results after changing only `VOLCANIC_SCALE()` from 1.0 to 0.5. Graphs below were produced using a date range of 1745-2100 and depicting changes in `RF_TOTAL()`, `GLOBAL_TAS()`, `VOLCANIC_SO2()`, and `CONCENTRATIONS_CO2()`. 

**Expected behavior**
I expected temperatures and CO2 concentrations to remain the same across both scenarios in the future since, to my understanding, Hector doesn't model any volcanic eruptions after the present-day. I also expected temperatures and CO2 concentrations to be higher and radiative forcing during eruptions to be less negative for the volscl=0.5 scenario in the past

**Actual behavior**
Temperatures and CO2 concentrations behaved mostly as expected before ~1850 but were lower for the volscl=0.5 scenario for most times afterwards. Additionally, radiative forcing during volcanic eruptions appears to have been more negative post-1850 for the volscl=0.5 scenario than volscl=1.0 scenario. Graphs showing these variables and volcanic SO2 are shown below:

![volscl_comparison_plot](https://github.com/JGCRI/hector/assets/103226047/d74a00a6-9544-4018-a178-aafe344bc434)

**Runtime and build environment (please complete the following information):**
 - OS: Windows

`SessionInfo()` output:
R version 4.4.0 (2024-04-24 ucrt)
Platform: x86_64-w64-mingw32/x64
Running under: Windows 10 x64 (build 19045)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.utf8  LC_CTYPE=English_United States.utf8    LC_MONETARY=English_United States.utf8
[4] LC_NUMERIC=C                           LC_TIME=English_United States.utf8    

time zone: America/New_York
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] ggplot2_3.5.1 hector_3.2.0 

loaded via a namespace (and not attached):
[...]

PR #781, opened 2025-05-29
Modify the volcanic RF inputs  

kdorheim:
This addresses #742 and makes Hector's treatment/use of volcanic forcing consistent with other similar modeling groups.


kdorheim:
The only change made to the volcanic input time series is from 1745-1750, the reference period is 0. 

SV


This prevents the cooling from the volcanic forcing in the future period, leading to a pretty constant* increase in total RF and global temperature. 


* pretty constant across scenarios, there are some small differences, though because the carbon cycle feedback/interactions
* 
RF_vol


global_tas

**** 

PR materials - [PR_781.zip](https://github.com/user-attachments/files/20513548/PR_781.zip)

kdorheim:
@bpbond the other todos that might be related to this are a fix for the historical NOx emissions and a default parameter update, do we want to include these changes here or should they be separate PRs?

PR #782, opened 2025-05-30
Time varying natural CH4 emissions

kdorheim:
Currently, natural CH4 emissions are held constant throughout the simulation run. This pull request introduces a new capability that will enable these emissions to be represented as a time series. Although there is no immediate change in the behavior (natural CH4 emissions remain constant), this development lays the groundwork for potential future adjustments where natural CH4 emissions can vary over time.

The following figures demonstrate this new capability. The time-varying natural CH4 emissions time series was randomly generated.

The anthropogenic CH4 emissions (left) do not change, whereas the natural emissions (right) are different between the two hector runs. 

ch4_emissions

The time-varying natural emissions affect CH4 concentrations and global temperature. 
CH4_concentration

global_tas

**** 

[PR_782.zip](https://github.com/user-attachments/files/20527926/PR_782.zip)

### DIFF
--- a/inst/include/ch4_component.hpp
+++ b/inst/include/ch4_component.hpp
@@ -59,9 +59,9 @@ private:
   tseries<unitval> CH4_emissions;
   tseries<unitval> CH4;           // CH4 concentrations, ppbv CH4
   tseries<unitval> CH4_constrain; // CH4 concentration constraint, ppbv CH4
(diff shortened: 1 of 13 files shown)


bpbond (approved):
Looks great @kdorheim -- thank you. I like this kind of PR that doesn't change anything immediately but, as you say, lays the foundation for cool work in the future.

PR #784, opened 2025-07-02
Easy get all hector variables

kdorheim:


bpbond:
Ah, good points, thanks.

PR #786, opened 2025-07-29
Update Methane Sink Parameters 

kdorheim:
Since the Hector CH4 and OH components were implemented in 2015, there have been some advancements in the scientific consensus of the methane lifetime. Which Hector V3.2 underestimates compared to benchmarks from AR5, AR6, and Stevenson et al 2020. 

image

In this PR we update the following. 

* Mean CH4 soil sink life time $\tau_{soil}$
* Mean CH4 stratosphere sink life time $\tau_{soil}$
* Initial CH4 OH sink lifetime $\tau_{OH_0}$
* Natural CH4 emissions 
* The NMVOC, CO, NOx coefficients for the  CH4 OH lifetime sink equation

So now the CH4 lifetime with respect to the OH sink and total atmospheric CH4 lifetime is more consistent with other published values.  

image

image

Which is great! Looking at CH4 concentrations, though, you might ask why the old implementation is more aligned with the CMIP6 [CH4] concentrations after 1850? 

image

This just has to do with the way that we optimized Hector's natural CH4 emissions to these concentrations. In contrast, with the new implementation, the natural CH4 emissions are held constant at the inferred preindustrial value. 

@ssmithClimate  and @bpbond  there are two questions that I would like to address before I think this PR is ready for review. 

1. Should we use a constant natural CH4 emission? Or would we be interested in making use of our new time-varying natural CH4 emissions capability from #782? 
2. Where would be a good place to document how the new NMVOC, CO, NOx coefficients were derived? In this Pull Request? I'm looking for a clear reference point in the ini files that we can cite in the future.


bpbond (approved):
Nice work @kdorheim 

Re documenting things clearly, perhaps (1) add DOIs to the INI file comments and (2) add your rendered Overleaf document to the repo, e.g. in a `docs/` folder or `inst/extdata/` or something like that?

Issue #795, opened 2025-11-03
rh_ch4 bug(s)


kdorheim:
Unexpected rh_ch4 behavior/issues raised by Raines Lucas

1. rh_ch4 units are reported to by Pg C (seems high)
2. Even when permafrost is set to 0 we are seeing non 0 rh_ch4 which is unexpected 

<img width="685" height="593" alt="Image" src="https://github.com/user-attachments/assets/bc14bbc0-695f-4ebe-a9cf-ba509c28f574" />

To reproduce the plot above I set permafrost_c = 0 in an ini file and then launched command line hector runs and visualized the hector outputstreams results.

kdorheim:
Things to check 

- [ ] Is the rh_ch4 being added to the atmosphere in the[ ch4_component ](https://github.com/JGCRI/hector/blob/662e6ca22fec7925ded008af9851e6cfc16f8907/src/ch4_component.cpp#L171)0 like we are expecting it to be? 
- [ ] Is the correct rh_ch4 variable being returned to the outputsteam in the src/csv_outputstream_visitor.cpp? 
- [x] Does this behavior happen when fetching results via the R interface?

kdorheim:
>  Does this behavior happen when fetching results via the R interface? 


When I use the same INI files as above but instead launch the runs/query results in R we see the behavior we are expecting 

<img width="608" height="460" alt="Image" src="https://github.com/user-attachments/assets/348f23ea-1426-4d27-a942-8fe39cbbcb1d" />

PR #798, opened 2025-12-12
RH CH4 bug fix

kdorheim:
This is a bug fix that addresses/answers the two issues identified by Raines Lucas in #795. Previously the csv output visitor was returning rh instead of rh_ch4.

Now the rh_ch4 results are 0 when permafrost is turned off. 

image

There was also a question of whether the units were correct, and yes, the rh_ch4 variable should be PgC.


bpbond (approved):
Good detective work @kdorheim ! Thanks for figuring this out.

PR #799, opened 2025-12-31
Output stream visitor bug fix

kdorheim:
This PR addresses a rounding issue that is only relevant to the CSV output stream & addresses an inline TODO item related to the csvouput stream visitor.  

**The Problem**

While working on #795, I realized that there were some numerical differences between Hector results written out by the CSV output stream visitor vs. fetched via R. 

When we compare the numerical results from hector output via R vs. csv outputter the overall summary table absolute error (this is across multiple scenarios & variables)

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-9b9aba8f-7fff-7154-39c4-bb1668ae7c2a"><div dir="ltr" style="margin-left:0pt;" align="center">
Min | 1 IQR | Median | Mean | 3 IQR | Max
-- | -- | -- | -- | -- | --
0 | 0 | 0.000018 | 0.081499 | 0.003465 | 4.999791

</div><br /><p dir="ltr" style="line-height:1.38;margin-top:0pt;margin-bottom:0pt;"><span style="font-size:11pt;font-family:Arial,sans-serif;color:#000000;background-color:transparent;font-weight:400;font-style:normal;font-variant:normal;text-decoration:none;vertical-align:baseline;white-space:pre;white-space:pre-wrap;" </span></p></b><br class="Apple-interchange-newline">

Of the 80 variables considered 57 of them had a MAE &gt; 1E-6 

Looking at the variable with the largest error, DO ocean C, while the relative MAE of 4.999791. A visual comparison of  DO ocean C reveals a step like behavior in the csv result which seems odd and is most likely due to rounding. 

Screenshot 2025-12-31 at 8 48 53 AM

Same data but different perspective. 

Screenshot 2025-12-31 at 8 50 07 AM

**the solution** 

It turns out the csvFile has a default precision level of 6 whereas typical C++ double precision is 16. Since R was pulling directly from the C++ precision level we needed to make sure that the csvFile was using the same precision level. 

After changing the precision level of the csvFile the MAE between R and the csv output the MAE is now 1.035752e-13. Which is much better! And now no longer see the step behavior we were seeing before. 

Screenshot 2025-12-31 at 8 52 57 AM

### DIFF
--- a/src/csv_outputstream_visitor.cpp
+++ b/src/csv_outputstream_visitor.cpp
@@ -74,6 +74,7 @@ CSVOutputStreamVisitor::CSVOutputStreamVisitor(ostream &outputStream,
   current_date = 0;
   datestring = "";
   spinupstring = "";
+  csvFile.precision(15);
 }
 
 //------------------------------------------------------------------------------
@@ -134,7 +135,6 @@ void CSVOutputStreamVisitor::visit(Core *c) {
 void CSVOutputStreamVisitor::visit(ForcingComponent *c) {
   if (!core->outputEnabled(c->getComponentName()))
     return;
-   csvFile.precision(15);
 
   if (c->currentYear < c->baseyear)
     return;
--- a/src/csv_outputstream_visitor.cpp
+++ b/src/csv_outputstream_visitor.cpp
@@ -109,18 +109,6 @@ void CSVOutputStreamVisitor::visit(Core *c) {
   core = c;
 }
 
-// TODO: consolidate these macros into the two MESSAGE ones,
-// and shift string literals to D_xxxx definitions
-
-// Macro to send a variable with associated unitval units to some output stream
-// Takes s (stream), c (component), xname (variable name), x (output variable)
-#define STREAM_UNITVAL(s, c, xname, x)                                         \
-  {                                                                            \
-    s << linestamp() << c->getComponentName() << DELIMITER << xname            \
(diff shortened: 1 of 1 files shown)


src/csv_outputstream_visitor.cpp
```diff
@@ -146,7 +134,7 @@ void CSVOutputStreamVisitor::visit(Core *c) {
 void CSVOutputStreamVisitor::visit(ForcingComponent *c) {
   if (!core->outputEnabled(c->getComponentName()))
     return;
-  streamsize oldPrecision = csvFile.precision(4);
+   csvFile.precision(15);
```
bpbond: Why do this at all here? Should we simply set the precision we want when the stream is opened and leave it?

bpbond (commented):
Question -- do we want to be constantly changing the precision of the stream?

src/csv_outputstream_visitor.cpp
kdorheim: oh good point! so would that be in the main.cpp file?  where the stream is opened? Or part of the `CSVOutputStreamVisitor` constructor?

bpbond: In the class constructor.

kdorheim:
@bpbond does this look correct?

bpbond (approved):
Looks good to me @kdorheim !

PR #801, opened 2026-02-03
Consistent Unit Conversion & Parameters

kdorheim:
Right now when we are converting emissions --> ppX for various concentration types, they are inconsistent with one another. If we use a consistent value for the number of moles of CO2 (simpleNbox), CH4, N2O, and halocarbons in the atmosphere, we can be consistent across the hector components. 

Right now I have just implemented this change to the CH4 component because I have some questions for @bpbond

### DIFF
--- a/inst/input/hector_ssp245.ini
+++ b/inst/input/hector_ssp245.ini
@@ -7,6 +7,7 @@ endDate=2300
 trackingDate=9999	; year to start tracking (only carbon currently)
 do_spinup=1			; if 1, spin up model before running (default=1)
 max_spinup=2000		; maximum steps allowed for spinup (default=2000)
+atmosphere_mol=1.727e20	; number of moles in the atmosphere, mols 
 
 ;------------------------------------------------------------------------
 [ocean]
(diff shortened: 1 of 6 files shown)


inst/input/hector_ssp245.ini
```diff
@@ -7,6 +7,7 @@ endDate=2300
 trackingDate=9999	; year to start tracking (only carbon currently)
 do_spinup=1			; if 1, spin up model before running (default=1)
 max_spinup=2000		; maximum steps allowed for spinup (default=2000)
+atmosphere_mol=1.727e20	; number of moles in the atmosphere, mols
```
bpbond, line 10: Hmm, why make this a parameter? Is it uncertain, or something we expect users to change?

src/ch4_component.cpp
```diff
@@ -157,24 +156,38 @@ void CH4Component::run(const double runToDate) {
...
   } else {
 
-    // modified from Wigley et al, 2002
-    // https://doi.org/10.1175/1520-0442(2002)015%3C2690:RFDTRG%3E2.0.CO;2
-    const double current_ch4em = CH4_emissions.get(runToDate).value(U_TG_CH4);
+   // When emission driven hector uses a modified implementation of
+   // Wigley et al, 2002 (https://doi.org/10.1175/1520-0442(2002)015%3C2690:RFDTRG%3E2.0.CO;2)
+      
+    // Unit conversion realted terms
+    #define PG_C_TO_TG_CH4 (1000.0 * 16.04 / 12.01) // should this be #define or would it make sense to do something like constexpr?
```
bpbond, line 163: I think `#define` is fine although tagging @pralitp for his $0.02

src/ch4_component.cpp
```diff
@@ -157,24 +156,38 @@ void CH4Component::run(const double runToDate) {
...
+    
+    // Determine the conversion factor used to convert from Tg CH4 --> ppbv
+    double atm_mols = core->getAtmMols();
+    const double Tg_to_mol = 1 * 1e12 * (1/16.04);         // 1 Tg CH4 --> moles of CH4
```
bpbond, line 167: Whichever you choose, though, be consistent. I would use a mix of `#define` and `const` without a good reason.

bpbond: `#define` with all-caps names seems clearer to me, and makes it super easy to see constants in equations.

bpbond (commented):
Hi @kdorheim apologies for my slow response on this.

Interestingly, this (varying physical constants in the codebase) is something that E3SM is working on right now, too!

Some comments for you to consider; happy to discuss more. Thanks for tackling this.

kdorheim:
Closing because #804 does a better job of this.

PR #802, opened 2026-03-13
Initial Carbon Pools

kdorheim:
This PR has no effect on hector output behavior, but it does the following 
1) it increases the initial size of the earth pool to prevent the earth pool from going negative in certain idealized runs as part of RCMIP 
2) there was a parameter called atmos_co2, which was supposed to initialize the size of the atmosphere carbon pool in Pg C, this parameter is no longer necessary/active since the atmosphere C pool is initialized with the pre-industrial CO2 concentration


bpbond (approved):
Your PR comment talks about `atmos_c`, but it's `atmos_co2` that's removed.

kdorheim:
Ah good, catch, atmos_co2 stands for the atmospheric co2 in terms of Pg C. I am always messing them up.

PR #804, opened 2026-03-17
Universal Physical Constants

kdorheim:
This is take 2 at implementing consistent physical constants across hector components based on Ben's feedback to #801. We have added a physical_constants.hpp file where physical constants used across hector components are defined.

So far I have only implemented the change for the CH4 component & seeking feedback if this is the way we would like to continue.


inst/include/physical_constants.hpp
```diff
@@ -0,0 +1,22 @@
...
+ *  physical_constants.hpp
+ *  hector
+ *
+ *  Created by K. Dorheim on March 2026.
+ *
+ */
+
+// The number of moles in the atmosphere
+constexpr double ATMOSPHERE_MOL = 1.727e20;
+constexpr double CH4_molarMass = 16.043;
```
bpbond: Change to ALL CAPS for consistency?

bpbond (approved):
This looks exactly right to me @kdorheim ! It makes huge sense to centralize things in a physical constants header file.

kdorheim:
Closing this draft PR for now, may return to it at a later date.

